Issue #75, opened 2016-11-08
Integrate valgrind in CircleCI build target


fuchsto:
Valgrind should be added in at least a minimal CI target to check for memory leaks and consistency defects.

devreal:
I would prefer someone already familiar with CircleCI to do the integration as I cannot give this a high priority right now. It will also be much quicker to do for the maintainers of the CI scripts. Maybe @fmoessbauer can take over this task?

fuchsto:
@fmoessbauer is sick for the next days, actually it's the same situation here, it's not like CI is a core task for us either.
Also, I would prefer if the whole team, or at least more than 1-2 members, understand CI, otherwise the team in Munich degenerates to second level support. We have been new to CI two weeks ago as well.

And valgrind integration is not a mission critical feature of course, so I didn't expect a quick implementation of this issue request, we just should have it in our tasks.

fuchsto:
@devreal But, speaking of it: we should provide a documentation of our CI procedure to make contributions easier for other team members.

Documentation of CI is definitely a task for @fmoessbauer and me (@fuchsto).

See support issue #76 

Once documentation is available, we should discuss the current CI procedure and define all required integration test scenarios in a brief conference call.

devreal:
I agree that there  should be more than one person understanding the CI. I had a look at some of the scripts yesterday and it was not clear to me where to start with the intergration. Once we have documentation from #76 I'm happy to look into that.

fuchsto:
Perfect, thanks!

Yes, the CI scripts need some cleanup, shot myself in the foot quite frequently myself.
We will consolidate deployment first to make things less cryptic.

devreal:
I have been playing around with this today, mainly trying to get the CI and docker working locally (I still cannot build a docker container locally due to network issues). My first shot is at branch sup-75-valgrindci, please have a look at it. I am created a new docker configuration that installs valgrind and sets EXEC_WRAP, which is then used as a prefix to the test executable in dash-test.sh. Comments are welcome. 

One thing that is still missing: I think it only makes sense to run valgrind under the debug configuration so there is no need to also run the other configurations. I haven't figured out how to pass this information to the scripts so far.

fmoessbauer:
Nice Job, looks great! I implemented the described behavior by just skipping the main loop if container is `openmpi2_vg` and target is not Debug. I also extended the documentation according to this.

[...]

PR #93, opened 2016-11-12
Valgrind CI Integration

fuchsto:
Addresses #75


fuchsto:
Please use a squash commit to merge.

fmoessbauer:
Actually the valgrind CI target is not executed on circleci as otherwise the build would always fail due to false positives. However each developer can run the container and CI script on his machine and check the logs.

Issue #126, opened 2016-11-19
DASH algorithms are not team aware


fmoessbauer:
Almost all DASH algorithms are not team aware because they use `dash::barrier()` internally. This has to be changed to team local barriers.

If using GlobIters, this can be done the following way:
`auto & team = iter.pattern().team()` and `team.barrier()` instead of `dash::barrier()`.

fuchsto:
Well, this method chain

```
iter.pattern().team().barrier()
```

... is based on existing concepts already, so why not use it to extend the `GlobalIterator` concept by a method `barrier()` to synchronize all units in the team that have access to the iterator? It would always pick the correct team, too.

This would correspond to an atomic barrier (`dash::Atomic<T>.flush()|.barrier()`) in principle.

Issue #125, opened 2016-11-19
dash::Array does not support teams properly


fmoessbauer:
If using the array constructor with a team other than `dash::Team::All()`, the array is only allocated for the first (aka `team.position() == 0`) team. In all other teams, the array is not allocated.

See [this unit test](https://github.com/dash-project/dash/blob/bug-125-array-team/dash/test/ArrayTest.cc#L171) in `bug-125-array-team` as showcase.

fuerlinger:
It looks like the problem is the absolute `dash::myid()` in `GlobMem.h` used to construct the local pointers. When working with global-view indices it actually seems to work alright.

fuchsto:
@dash-project/developers 

Just some thoughts: 
I'm wondering if `dash::myid()` and `dash::barrier()` etc. should really behave the way they do at the moment.

If they have global reference, then they are actually not that useful. No algorithm or data structure could use them because they refer to a global, possibly disjoint, set of units.
When would an algorithm / container / whatever actually be interested in the global team hierarchy?

For example, all containers use `dash::Team::All()` as default initializer for the associated team.
But this means that users have to change their usage of the container interface (specifying a team explicitly or not) depending on the state of the active team (split or not).
Whe should at least provide something like `dash::Team::Local()` so default teams for containers
have stable semantics.

The interface design I know from several places is: make interfaces agnostic from unrelated states
and maintain semantics if possible.

So when a team engages in a collective algorithm, the library interface provided to the algorithm
should always be the same. To the algorithm (or container, or whatever), there should be no difference 
between the `Team::All()` case and the child team case. When a team is split, it creates its own universe.

Requesting the global state (`dash::Team::All()` or `team.parent()` for example) should be an explicit operation.
Currently it is an explicit operation to request the default, the "typically useful and less surprising" behavior.

fuerlinger:
`dash::barrier()` is just shorthand for the more long-winded `dash::Team::All().barrier()`, same for `dash::myid()`. It is the right and natural thing to do for applications that don't construct teams explicitly, which might be the majority.

For applications that use explicit teams, things are more complex of course. A unit can be part of an arbitrary number of teams and therefore there is no alternative sensible default team a collective operation should refer to (there is no notion of a single  _active_ team, there can me multiple active teams in the sense that the Team object has been successfully constructed). There is no way around explicitly specifying the team in those cases. DASH containers are always allocated over a team and therefore container related operations can always choose the right team (as e.g., `array.barrier()` already does). 

[...]

PR #129, opened 2016-11-19
Feat hdf5 team

fmoessbauer:
Extends dash's hdf5 storage driver to support containers that are only allocated by a subset of units.


fmoessbauer:
Currently blocked due to #126 and #125

fuchsto:
@fmoessbauer Fixes for #125 and #126 have pull requests pending.

fmoessbauer:
Still blocked due to PR #135.

ddiefenthaler:
@fuchsto This pull request looks fine to me.
But I think a more experienced eye should also take a look.

PR #132, opened 2016-11-19
Barriers on local team in algorithms

### LINKED ISSUE #126: DASH algorithms are not team aware

fuchsto:
Fixed #126

### DIFF
--- a/dash/include/dash/GlobMem.h
+++ b/dash/include/dash/GlobMem.h
@@ -125,8 +125,8 @@ public:
     _begptr = _allocator.allocate(_nlelem);
     DASH_ASSERT_MSG(!DART_GPTR_ISNULL(_begptr), "allocation failed");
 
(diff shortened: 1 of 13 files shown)


fmoessbauer:
This is still not valid, as `auto myid = dash::myid()` is used, which refers to the `Team::All()`. Hence, global ID 0 has to participate which might not be the case. `team.myid()` has to be used instead.

I will fix that tomorrow.

fuchsto:
@fmoessbauer Yes, sorry, all references to `dash::size()` and `dash::myid()` should be fixed in branch `bug-126-algo-barriers` now.

PR #133, opened 2016-11-19
Resolution of local unit id in GlobMem

### LINKED ISSUE #125: dash::Array does not support teams properly

fuchsto:
Fixes #125

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,6 +94,9 @@ Features:
         - `dart__base__locality__domain`
         - `dart__base__locality__unit`
 
+Fixes:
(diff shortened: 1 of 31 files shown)


fmoessbauer:
There are still some team all related functions like `dash::size()` or `dash::myid()` used. This leads to the problems described in #132.

PR #145, opened 2016-11-22
Fixed compiler error in dash::mismatch

fuchsto:


dash/include/dash/algorithm/Accumulate.h
```diff
@@ -82,21 +83,22 @@ ValueType accumulate(
...
   auto l_first        = index_range.begin;
   auto l_last         = index_range.end;
   auto l_result = std::accumulate(l_first, l_last, init, binary_op);
 
-  dash::Array<index_t> l_results(dash::size());
+  dash::Array<index_t> l_results(team.size());
 
   l_results.local[0] = l_result;
 
   in_first.pattern().team().barrier();
```
devreal: I assume `in_first.pattern().team().barrier()` is the same as `team.barrier()` here? A unified use would improve readability.

devreal (changes requested):
Usage of `team.barrier()` vs `in_first.pattern().team().barrier()` should be unified

fuchsto:
@devreal Done, also fixed missing `#endif` in `algorithm/Mismatch.h` 

@JosefSchaeffer There is no unit test suite for `dash::mismatch`, please add it!

devreal (approved):

fuchsto:
Merging after feedback from @devreal

Issue #63, opened 2016-11-05
Tests hang using OpenMPI 2.0


fmoessbauer:
The following tests hang when using openmpi 2.0. For testing and verification the new docker container `dashproject/ci:openmpi2` can be used. The container provides a openmpi 2.0 environment which is build from [sources](https://www.open-mpi.org/software/ompi/v2.0/).

It is also possible to ssh into the container on circleci and debug there.

# OpenMPI v2.0

## 2 Units
- [ ] SharedTest.AtomicAdd
- [ ] TransformTest.Array*

## 3 Units
- [x] CopyTest.AsyncLocalToGlobPtr

fuchsto:
@fmoessbauer I think some of these failed test are subsequent errors due to a defect in some test case that corrupts the runtime.
Can you please configure the containers such that each test suite is run in isolation using 

```
mpirun [ ... ] dash-test-mpi --gtest_filter="TheTestSuite.*"
```

... instead of a single execution of  `dash-test-mpi`?
I see no other way to locate the underlying bug.

fuchsto:
All clues lead to a defect in progress communication in OpenMPI:
https://github.com/open-mpi/ompi/issues/157

When an `MPI_Accumulate` is followed by an `MPI_Barrier` and OPAL progress threads are disabled, a deadlock occurs:

![openmpi2_opal_lock](https://cloud.githubusercontent.com/assets/37075/20332965/beed504c-abaf-11e6-9c66-b4f587f0f524.png)

In the `SharedTest.AtomicAdd` test case, all units add a local value to a shared global address using `MPI_Accumulate`. To wait for completion of all remote operations, units call `MPI_Barrier`.
If any unit X did not complete its `MPI_Accumulate` operation when any of its target units Y enters the barrier, progress communication stalls and unit X will never reach the barrier, in effect blocking execution forever.

The usage pattern in DART/DASH is fine. To fix these deadlock situations, OpenMPI has to be built with
support for MPI_THREAD_MULTIPLE  and OPAL pogress threads, but I didn't find a known good build configuration, yet.

And apparently, enabling progress threads leads to stability issues:
https://bugzilla.redhat.com/show_bug.cgi?id=1105902

Some additional remarks on progress threads and performance:
https://www.open-mpi.org/community/lists/users/2012/06/19588.php

[...]

PR #152, opened 2016-11-23
Fixed test case CopyTest.AsyncLocalToGlobalGptr for openmpi2

fuchsto:
Addresses #63


fuchsto:
!!! **Tentative merge to force update of docker container** !!!

Issue #140, opened 2016-11-21
Multiple memory leaks in DART


devreal:
On multiple occasions memory does not seem to be free'd properly in DART. Most of the leaks seem to stem from the use of MPI groups and a missing MPI_Group_free, e.g., : 

```
==1725== 80 (72 direct, 8 indirect) bytes in 1 blocks are definitely lost in loss record 39 of 84
==1725==    at 0x4C2DB8F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==1725==    by 0x52BDEAF: ompi_group_allocate_plist_w_procs (in /home/joseph/opt/openmpi-master/lib/libmpi.so.0.0.0)
==1725==    by 0x52BDFC7: ompi_group_allocate (in /home/joseph/opt/openmpi-master/lib/libmpi.so.0.0.0)
==1725==    by 0x52BE7E8: ompi_group_union (in /home/joseph/opt/openmpi-master/lib/libmpi.so.0.0.0)
==1725==    by 0x52EC8EA: PMPI_Group_union (in /home/joseph/opt/openmpi-master/lib/libmpi.so.0.0.0)
==1725==    by 0x42B958: dart_group_union (dart_team_group.c:69)
==1725==    by 0x42BC4A: dart_group_addmember (dart_team_group.c:156)
==1725==    by 0x434471: dart__base__host_topology__update_module_locations (host_topology.c:197)
==1725==    by 0x4377CF: dart__base__host_topology__create (host_topology.c:678)
==1725==    by 0x438C1A: dart__base__locality__create (locality.c:174)
==1725==    by 0x43851B: dart__base__locality__init (locality.c:77)
==1725==    by 0x41F830: dart__mpi__locality_init (dart_locality_priv.c:21)
```

However, there are other leaks that are not related to MPI groups, e.g., : 
```
==1725== 30 bytes in 1 blocks are definitely lost in loss record 20 of 84
==1725==    at 0x4C2DB8F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==1725==    by 0x436CFC: dart__base__host_topology__create (host_topology.c:547)
==1725==    by 0x438C1A: dart__base__locality__create (locality.c:174)
==1725==    by 0x43851B: dart__base__locality__init (locality.c:77)
==1725==    by 0x41F830: dart__mpi__locality_init (dart_locality_priv.c:21)
[...]

PR #155, opened 2016-11-23
Bug 140 dart memleaks

devreal:
Fix multiple memory leaks, including leaking `MPI_Group` handles and problems in `host_topology`.


fuchsto:
@devreal Merging the current state of development into your branch should fix this.

fuchsto:
@devreal 

Lots of much needed fixes, thanks!

What's with the TODO comment / question here:
https://github.com/dash-project/dash/pull/155/files#diff-94c24e191167fd4a413d88568786e406R139

I only refactored the parts in DART that I needed for my stuff in DASH, so I have no idea what's the
correct semantics here.
This is also why I don't like functions named "copy" in C and rather use "clone" or "assign" in interfaces (see `dart_locality.h`).
For "copy", there is no intuitive connotation of ownership semantics, either.

fuchsto:
Fixes #140

devreal:
@fuchsto: My todo was due to the fact that `dart_group_copy` only makes a shallow copy of the dart_group_t object (it does not duplicate the MPI group object). I thought that passing the same MPI group as input and output to `MPI_Group_union` is erroneous. However, I after re-thinking this it's OK. Deleted it in #160.

For your question in the same place: it's a global unit ID ;)

Issue #146, opened 2016-11-22
Execute all test suites in single, isolated runs of dash-test


fuchsto:
Currently, if a test leads to a broken or undefined state of the DASH runtime, subsequent tests may fail and thus report false positives.

Add a script that allows to perform tests as isolated runs of single test suites.

fuchsto:
Created branch `sup-146-test-isolated`

devreal:
Thanks a lot for this, will try it today.

devreal:
Tried it. Unfortunately, the script fails on our cluster due to the `-k` parameter not being supported by `timeout` in versions < 8.5, see http://git.savannah.gnu.org/cgit/coreutils.git/tree/NEWS#n1626. Is the additional kill really necessary?

fuchsto:
@devreal 
The kill-switch (no pun intended) was just for additional stability. Didn't know the option is non-standard.

PR #166, opened 2016-11-25
Check for version of timeout and other fixes in dash-test-all-single.sh

devreal:
Check for version of GNU timeout (has to be at least 8.5 for -k to work)
nocolor cannot be an alias
Fix error handling on first invocation of mpirun 

See #146


fmoessbauer:
Nice Job! We also should document this new CI feature in `dash/doc`

Issue #52, opened 2016-11-02
Add CI configuration using NastyMPI


fuchsto:
Some defects are only visible by chance, like missing barriers after local initialization.
I think we could drastically reduce false positives in CI builds by integrating NastyMPI.

fmoessbauer:
Support for NastyMPI is provided in branch `feat-52-nastympi`. Currently NastyMPI has to be compiled by the user.

For the CI tests it would be nice if cmake pulls and builds it automatically if  specified (like google test). @rkowalewski can you provide a cmake build script for nastympi? That would make the process much easier.

rkowalewski:
@fmoessbauer: Yes, I will do that.

fmoessbauer:
@rkowalewski : Please check if I integrated NastyMPI correctly in DASH. How can I verify that the proxy is actually used?

rkowalewski:
The original solution had two minor issues:
a) While we link statically we always use shared libraries. For this reason we should use the shared library version of Nasty-MPI. --> Fixed
b) The order in which libraries are listed during the link step mattes. Before the small fix, Nasty-MPI was simply added to the end of all the libraries. However, we have to link Nasty-MPI before the native MPI library. --> Fixed

On my local machine Nasty-MPI prints now debug output. I have checked it with a simple dash::array test.
@fmoessbauer All changes are now in the corresponding branch.

fuchsto:
@rkowalewski Thanks!
But currenlty the NastyMPI-tests fail with:

~~~
/opt/dash/build-ci/20161128-223703--uuid-9d6d693f/Nasty/bin/dash/test/mpi/dash-test-mpi:
error while loading shared libraries: libnasty_mpi.so.0: cannot open shared object file:
No such file or directory
~~~

See logs in https://circleci.com/gh/dash-project/dash/551

rkowalewski:
Is there any chance to see the linker commands on circle CI? I have checked my local build scripts and it works. Probably LD_LIBRARY_PATH, or similar, is different in the docker machine.

fmoessbauer:
Hi, I don't think this is a circle CI specific problem, as the tests are run inside our docker containers. Just run the container locally as described in this [wonderful documentation](http://dash.readthedocs.io/en/latest/CI/#online-ci-providers)

fmoessbauer:
The problem is quite simple. NastyMPI is not copied to DASH's install dir. Hence it is not found when running the CI script.

fmoessbauer:
[...]

PR #170, opened 2016-11-29
Feat 52 nastympi

fmoessbauer:
Adresses #52


fmoessbauer:
Finally drilled it down to a bug in NastyMPI when using gcc. See #52 for details. I fixed that in branch `NastyMPI/bug-gcc-macro` and now the CI works (but does not pass due to actual errors). However after that is finally fixed, we should pull from NastyMPI's master branch again.

fuchsto:
@rkowalewski Also see my PR #189 which is more stable at this point.

fmoessbauer:
Do we still need this PR? Aren't the changes already covered by the already merged PR #189 ?

fuchsto:
Yes, we need this PR as it fixes the integration of NastyMPI.

fmoessbauer:
Unfortunately we hit the memory limit on CircleCI (4 GB per container). However the memory dump looks highly suspicious to me, as the memory consumption of one mpi process is way higher than the others. For me, this indicates a memory leak. See this [memorydump](https://668-45930061-gh.circle-artifacts.com/0/tmp/memory-usage.txt).

| PID     | RSS/Memory  | %CPU  | COMMAND|
| ------- | ------- | --------- | ------------ |
| 21921 | 3591264 | 33.6 | dash-test-mpi |
| 21922 | 476232 | 71.2 | dash-test-mpi |
| 21923 | 475704 | 70.4 | dash-test-mpi |

fuchsto:
Heh, looks like you found a memory leak in NastyMPI.
Your debugging-Fu was stronger than mine, I thought it was a corrupt free / bad malloc in finalization of NastyMPI.
I'll see if I can valgrind it out.

rkowalewski:
I will have a look at this.

rkowalewski:
@fuchsto @fmoessbauer After investing much time into these memory leaks and `MPI_Win_get_attr` errors after calling `MPI_Finalize` I have pinpointed and fixed it. Now the build looks good however the status is "TIMED OUT". I cannot really tell why but maybe the timeout in our CI scripts is too low? We may also reduce the NASTY_SLEEP_INTERVAL to 3 and try it again.

PR #176, opened 2016-11-30
Fixed default types for 32-bit targets, added CI checks

fuchsto:


devreal:
Looks good to me, can be merged as soon as all checks passed.

fuchsto:
Thanks for the quick reply!

Issue #143, opened 2016-11-22
Strong typing for local and global IDs


clemensmanert:
Create a new own type for each kind of id, so they can not be mixed up any more.

devreal:
Can you give an example for which IDs this is problematic?

fuchsto:
@fmoessbauer Why do you use `dash::myid()` in GlobMem even if there is a team explicitly specified?

https://github.com/dash-project/dash/blame/development/dash/include/dash/GlobMem.h#L129

fmoessbauer:
This is required, as `lbegin(dart_unit_t global_id)` requires global unit ids. Hence, I renamed the function parameters too. IMO this is not a good solution as I already mentioned in the meeting, however otherwise we have to change the DART interface.

I know you have changed it in PR #133, however that was not a good one, as then the unit test failed. In fact, the bug was in the [Array implementation](https://github.com/dash-project/dash/pull/135/files) and not in GlobMem. I stumbled upon this curiosity too.

fuchsto:
@fmoessbauer Disregard that, I got it myself, can't quite remember why I implemented it this way, though.

EDIT: deleted my comment, you were too quick to answer and I can't seem to undelete it

fuchsto:
@dash-project/developers 

After all, I think @clemensmanert is right.
In C++ noone would give this a second thought and use phantom types or any other strong typing for
exactly this use case.
Documentation is not a solution. Any arcane interface can be documented, good interfaces don't need it.

I would introduce types in DART like:

~~~c
typedef struct {
  dart_unit_t  id;
} dart_global_unit_t;

typedef struct {
  dart_unit_t  id;
} dart_local_unit_t;
~~~

... with some union-magic for convenient strong typing. This would work for DART (C99).

This would solve the situation in `GlobMem` where some methods expect global, some local unit ids.
We could just provide both:

~~~c++
pointer lbegin(dart_global_unit_t gunit) {
  // ...
}

pointer lbegin(dart_local_unit_t lunit) {
  // convert to global unit id and pass to dart_global_unit_t overload
}
~~~

devreal:
To me it looks like this is only some (though maybe more prominent) form of documentation as well as long as we do not make sure that `dart_unit_t` itself is _never_ used by the user in DASH. If implicit conversion has to happen we lose all semantic information at some point or another. Not sure what interfaces have to be changed for this to work.  Or maybe I'm missing the point here?

fuchsto:
[...]

PR #186, opened 2016-12-06
Implement typed unit IDs

devreal:
Implement typed unit IDs to differentiate between global (rare) and team-local unit IDs (common). Spotted some wrong usages in the process of converting most places where `dart_unit_t` was used. 

Addresses #143


devreal:
Please carefully review and consider a squash merging

fuchsto:
I'm on it, thanks @devreal

dash/examples/ex.02.matrix/main.cpp
```diff
@@ -40,7 +40,7 @@ int main(int argc, char* argv[])
   dash::TeamSpec<2> teamspec_2d(team_size, 1);
   teamspec_2d.balance_extents();
 
-  dart_unit_t myid   = dash::myid();
+  dash::global_unit_t myid   = dash::myid();
```
fmoessbauer, line 43: Why is myid of `global_unit_t`? Does that even matter for this example?

dash/include/dash/GlobMem.h
```diff
@@ -253,13 +253,13 @@ class GlobMem
    * \param global_unit_id id of unit in \c dash::Team::All()
    */
   const ElementType * lbegin(
```
fmoessbauer, line 255: A globmem is allocated with a team, hence its (at least) public methods should use local unit ids. @fuchsto , @devreal , @fuerlinger please correct me if I am wrong.

I know, this will create a cascade of changes in DASH.

dash/include/dash/SharedCounter.h
```diff
@@ -22,7 +22,7 @@ class SharedCounter {
    */
```
fmoessbauer, line 22: The `SharedCounter` should either be deleted or reimplemented, as it is not team aware. I will open an issue for that.

fuchsto:
The changes need documentation.

`dash::internal::unit_id` should not be located in internal namespace as it would be hidden from the public interface otherwise.

To users, the interface of unit id types is highly relevant, however. Use cases like:

~~~c++
for (dash::local_unit_t lid { 0 }; lid < nunits; ++lid) { ... }
~~~

... are not self-explanatory if the operator overloads are not documented.

More important, type conversion operators should be defined as `= delete`.
Currently there is a conversion operator to `dart_unit_t` and a conversion constructor from `dart_unit_t`.

In effect, this allows use cases we intended to prohibit in the first place:

~~~c++
dash::local_unit_t  l_uid { 12 };
dash::global_unit_t g_uid { 12 };
l_uid = g_uid; // works, but shouldn't even compile
~~~

Added test suite `UnitIdTest` for further test of unit id semantics.

devreal:
We can move the unit_id struct to the public interface, no problem. I will also add documentation for the type. 

The reason for adding the conversion operator to dart_unit_t is to avoid having to access .id when calling dart_* functions. The problem is not the conversion c'tor but the assignment operator. I will delete it. 

I am a bit unhappy that we now have to write `dash::local_unit_t lid { 0 };` instead of a simple assignment but allowing implicit construction from `dart_unit_t` would prevent us from catching cases of wrong usage if the raw type is passed to a DASH function.

fuchsto:
You can, but

~~~c++
operator dart_gptr_t() noexcept
~~~

... should be explicit then.

Otherwise an implicit conversion `local_unit_t -> dart_gptr_t -> global_unit_t` would be allowed.
Assignments like

~~~c++
dash::local_unit_t lid = 12;
~~~

... should also be okay then, but

~~~c++
dash::local_unit_t lid = dash::global_unit_t(12);
~~~

... should not, while

~~~c++
dash::local_unit_t lid = static_cast<dart_gptr_t>(dash::global_unit_t(12));
~~~

... is okay again.

devreal:
Mhh, making 
```
operator dart_unit_t() noexcept
```
explicit seems to also prevent simple comparison with integer operands, e.g., `_myid == -1`. That would require an even bigger change to the code base.

fuchsto:
Number literals in code have implicit type `int`, we can rely on that.
So specializing operators for `int` should help.

Ah, but then again, `dart_gptr_t` *is* an `int`.
I'll find a solution, I adore stuff like that.

devreal:
Unfortunately, `dart_unit_t` is a `typedef` to `int32_t` (which in fact is `int`). The compiler does not honor different typedef'd types. So that won't help either.

fuchsto:
Yes, noticed that a wee moment later, see my edited comment.

I will add some unit tests for discussion.

devreal:
It seems that we can actually delete 
```
template<unit_scope Scope>
unit_id<IdScope> operator=(const unit_id<Scope> & id)
```

Have to test that though.

dash/include/dash/Team.h
```diff
@@ -542,7 +543,7 @@ class Team
   size_t                  _position     = 0;
   size_t                  _num_siblings = 0;
   mutable size_t          _size         = 0;
-  mutable dart_unit_t     _myid         = -1;
+  mutable local_unit_t    _myid{-1};
```
fmoessbauer: Use better `UNDEFINED_LOCAL_UNIT_ID`

dash/include/dash/algorithm/Find.h
```diff
@@ -47,7 +47,7 @@ GlobIter<ElementType, PatternType> find(
 
     // Pointer to first element in local memory:
     const ElementType * lbegin        = first.globmem().lbegin(
-                                           team.myid());
+                                           dash::myid());
```
fmoessbauer: Use `dash::Team::GlobalUnitID()`

dash/include/dash/algorithm/LocalRange.h
```diff
@@ -269,7 +269,7 @@ local_range(
   // Local start address from global memory:
   auto pattern = first.pattern();
   auto lbegin  = first.globmem().lbegin(
-                   pattern.team().myid());
+                   dash::myid());
```
fmoessbauer: Use `dash::Team::GlobalUnitID()`

dash/include/dash/algorithm/MinMax.h
```diff
@@ -191,7 +191,7 @@ GlobIter<ElementType, PatternType> min_element(
 
     // Pointer to first element in local memory:
     const ElementType * lbegin        = first.globmem().lbegin(
-                                          team.myid());
+                                          dash::myid());
```
fmoessbauer: Use `dash::Team::GlobalUnitID()`

dash/include/dash/iterator/GlobIter.h
```diff
@@ -304,7 +304,7 @@ class GlobIter
     DASH_LOG_TRACE_VAR("GlobIter.[]", local_pos.index);
     // Global reference to element at given position:
     return ReferenceType(
-             _globmem->at(local_pos.unit,
+             _globmem->at(dash::local_unit_t(local_pos.unit),
```
fmoessbauer: `dash` not necessary, so remove it

dash/include/dash/util/UnitLocality.h
```diff
@@ -63,12 +63,12 @@ class UnitLocality
   }
 
   UnitLocality(
-    dart_unit_t   unit)
-  : UnitLocality(dash::Team::All(), unit)
+    global_unit_t   unit)
+  : UnitLocality(dash::Team::All(), local_unit_t(unit))
   { }
```
fmoessbauer, line 68: Seems strange to me. Why `Team::All` and local id's ?

dash/test/GlobMemTest.cc
```diff
@@ -6,7 +6,7 @@ TEST_F(GlobMemTest, ConstructorInitializerList)
   auto target = dash::GlobMem<int>(target_local_elements);
 
   std::vector<int> glob_values;
-  for (int u = 0; u < dash::size(); u++) {
+  for (dash::local_unit_t u{0}; u < dash::size(); u++) {
```
fmoessbauer: Use `dash::global_unit_t` as team splits would break that loop.

fmoessbauer (commented):
Hi, nice job!
Just some minor points mentioned in the comments.

dash/include/dash/GlobMem.h
devreal: I agree, this is a point I noticed but forgot to mention in the discussion. We should fix this to use local_unit_t.

dash/include/dash/iterator/GlobIter.h
devreal: Good catch!

dash/include/dash/util/UnitLocality.h
devreal: The idea here is to enforce the use of a global ID if the team is guaranteed to be `dash::Team::All()`.

devreal:
Thanks a lot for your comments. I will fix the issues tomorrow.

dash/include/dash/GlobMem.h
fuchsto: @devreal @fmoessbauer @fuerlinger See my comment here:

https://github.com/dash-project/dash/issues/143#issuecomment-265541969

dash/test/GlobMemTest.cc
devreal: How would a team-split happen in this loop?

dash/include/dash/util/UnitLocality.h
devreal: Sorry, got your questions wrong: dash::Team::All() is a team like any other, so we have to pass a team-local unit ID to this constructor. We cannot handle the difference between `dash::Team::All()` and arbitrary teams with the types we have (unfortunately).

dash/test/GlobMemTest.cc
fmoessbauer: Sorry, thats my mistake. I mixed up types and semantics. I thought that a local unit type has a limited range (but as it is constructed without a team that is of course not right).
However for me it still looks strange as `dash::size()` refers to a "global" size, but currently we do not have something like local and global sizes. I do not even think that this might be beneficial at any place...

Simply ignore my comment. Your code is correct 👍

fuchsto:
@devreal
We should be ready for merging this behemoth.

Summary of my changes:

- Rebased to recent state of development branch
- Updated implementation of DART/DASH locality modules to typed unit ids
- Resolved your requests for review (marked `TODO[TF]` in comments)
- Renamed `*local_unit_t` to `*team_unit_t` and dependent definitions like `DART_UNDEFINED_LOCAL_UNIT_ID` / `DART_UNDEFINED_TEAM_UNIT_ID` etc.

**Please use a regular merge** (as in: **not** squash & merge) as the commit history is very, very valuable for future bug hunts.

Issue #184, opened 2016-12-04
Also run the CI tests with shared windows disabled


fuerlinger:
Once DART successfully builds again with disabled shared windows, we should run CI tests with both configurations: enabled and disabled. 

Disabled shared windows more closely resembles the distributed memory setting and might uncover issues that would otherwise go unnoticed. Plus it is needed for NastyMPI to work (intercept puts and gets).

fmoessbauer:
I agree, that this should definitely be tested. However this (again) rises the time the CI takes to finish.

Implemented in branch [feat-184-ci-sw](https://github.com/dash-project/dash/compare/feat-184-ci-sw).

fmoessbauer:
On CircleCI the tests fail using the OpenMPI2 container with more than one unit: See [CI Output](https://circleci.com/gh/dash-project/dash/602#artifacts/containers/2).

@fuchsto please have a look at the logs.

fuchsto:
I already changed the build config for "Minimal" to MPI shared windows disabled.
Shouldn't that suffice?

fuchsto:
It's fixed in branch `feat-52-nastympi`, will be fixed with merge of PR #170 

See branch `feat-184-shwin`

fuchsto:
@rkowalewski @fmoessbauer @fuerlinger 

In PR #189, I re-enabled MPI shared windows in CI builds as there apparently is a bug in OpenMPI 2, if I'm not mistaken.

There is a deadlock where one unit is entering a barrier, and another unit calls `MPI_Put` to a window shared with the barrier'ed unit (but in it's own segment):

![dash-log-openmpi2-deadlock_annotated](https://cloud.githubusercontent.com/assets/37075/20964782/3c8d9984-bc73-11e6-97e4-3412119cc0ac.jpg)

This should not block, if I understand the MPI standard correctly.

For the full log, see CircleCI build no. 646:
<https://circleci.com/gh/dash-project/dash/646>

Note that this failed for OpenMPI 2 only.
For MPICH and OpenMPI 1, the Minimal and NastyMPI test configurations passed.

This defect has could not be provoked by NastyMPI, another hint that it's actually a bug in OpenMPI 2.

rkowalewski:
[...]

PR #189, opened 2016-12-07
Applying integration of NastyMPI CI build

fuchsto:


dash/include/dash/Array.h
```diff
@@ -1014,7 +1066,10 @@ class Array
   void barrier() const
   {
     DASH_LOG_TRACE_VAR("Array.barrier()", m_team);
-    if (nullptr != m_team && dash::Team::Null() != *m_team) {
+    if (nullptr != m_globmem) {
+      m_globmem->flush_all();
+    }
+    if (nullptr != m_team && *m_team != dash::Team::Null()) {
```
devreal, line 1072: Can m_team ever be a nullptr? It seems to be initialized from a dash::Team reference in all c'tors.

fuchsto: Currently yes, but it's a pointer, so I check it. Members can be specified by *lots* of operations and checking for `nullptr` is more robust than validating every possible member assignment.
Also, it's in `Array::barrier()` so it's not like the test will have a say in performance.

fuchsto: ... also, most lint tools would otherwise nag about this.

fuchsto:
@devreal @fmoessbauer @rkowalewski 

See my comments in #184 

Please review and merge.
Please do not use squash merge as every commit refers to a separate logical block of changes.

devreal:
No objections from my side.

dash/test/CopyTest.cc
```diff
@@ -358,19 +358,40 @@ TEST_F(CopyTest, BlockingLocalToGlobalBlock)
...
   // Assign initial values: [ 1000, 1001, 1002, ... 2000, 2001, ... ]
   for (auto l = 0; l < num_elem_per_unit; ++l) {
-    array.local[l] = 0;
+    array.local[l] = ((dash::myid() + 1) * 10000) + (l * 10);
     local_range[l] = ((dash::myid() + 1) * 1000) + l;
   }
   array.barrier();
 
+  // Block- and global offset of target range:
+  auto block_offset  = _dash_size - 1 - dash::myid();
```
fmoessbauer, line 371: This holds until #186 is merged

fmoessbauer (changes requested):
Looks Ok to me. However calculations involving `dash::myid()` have to be explicit casted to e.g. `int`. And `GlobMemTest.ConstructorInitializerList` fails in CI

dash/test/CopyTest.cc
fuchsto: Requested change is in scope of #186, not this PR

rkowalewski:
Looks fine to me!

PR #190, opened 2016-12-09
Make handling of handles more robust

devreal:
Make creation and handling of handles more robust by defaulting to NULL and only returning a valid handle on success (in particular, do not return free'd handles)

### DIFF
--- a/dart-impl/mpi/src/dart_communication.c
+++ b/dart-impl/mpi/src/dart_communication.c
@@ -421,6 +421,8 @@ dart_ret_t dart_get_handle(
   uint64_t     offset = gptr.addr_or_offs.offset;
   int16_t      seg_id = gptr.segid;
 
+  *handle = NULL;
+
   /*
    * MPI uses offset type int, do not copy more than INT_MAX elements:
    */
@@ -434,8 +436,6 @@ dart_ret_t dart_get_handle(
     DART_LOG_ERROR("dart_get_handle ! failed: Unknown segment %i!", seg_id);
     return DART_ERR_INVAL;
   }
-
-  *handle = (dart_handle_t) malloc(sizeof(struct dart_handle_struct));
 
   if (seg_id > 0) {
     unit_g2l(index, target_unitid_abs, &target_unitid_rel);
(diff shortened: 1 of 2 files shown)


dart-impl/mpi/src/dart_communication.c
```diff
@@ -667,7 +658,13 @@ dart_ret_t dart_put_handle(
                    "(from local allocation) "
                    "target_unit:%d offset:%"PRIu64"",
                    nelem, dtype, target_unitid_abs, offset);
```
fmoessbauer, line 660: Are this faulthy encoded characters?

dart-impl/mpi/src/dart_communication.c
```diff
@@ -1156,13 +1150,15 @@ dart_ret_t dart_waitall_local(
...
     MPI_Status  *mpi_sta;
     MPI_Request *mpi_req;
     mpi_req = (MPI_Request *) malloc(num_handles * sizeof(MPI_Request));
     mpi_sta = (MPI_Status  *) malloc(num_handles * sizeof(MPI_Status));
     for (i = 0; i < num_handles; i++)  {
-      if (handle[i] != NULL) {
+      if (handle[i] != NULL && handle[i]->request != MPI_REQUEST_NULL) {
         DART_LOG_TRACE("dart_waitall_local: -- handle[%"PRIu64"]: %p)",
                        i, (void*)handle[i]);
         DART_LOG_TRACE("dart_waitall_local:    handle[%"PRIu64"]->dest: %d",
```
fmoessbauer, line 1164: Are this faulthy encoded characters?

fmoessbauer (commented):
Hi, maybe have a look at the suspicious characters.
Forget it, this are format specifiers. It's correct.

dart-impl/mpi/src/dart_communication.c
devreal: Do you mean the `%"PRIu64"` strings? They are macros used to correctly print the specified types (PRInt unsigned64).

fmoessbauer: Yes, the syntax was new to me. It's correct 👍

rkowalewski:
It is just a minor comment. But it is kind of redundant if we name a struct type like this: `struct mytype_struct {}`. It is simply not needed because the syntax already contains this information. Using mytype_struct without the keyword struct does not compile anyway. 
And if we apply some typedefs we say `typedef struct mytype mytype_t`.

dart-impl/mpi/src/dart_communication.c
fuchsto: This is the "most portability-compliant" variant I know to print an `u64`.
Yes, it does look strange and ugly, I should have formatted this more nicely.
A space-separated series of strings is treated as a single string in C/C++:

~~~c
printf("Foo" "Bar" "Batz");
// same as
printf("FooBarBatz");
// so:
printf("offset: %"PRIu64"");
// is same as (assuming PRIu64 = "ul")
printf("offset: %ul");
~~~

PR #196, opened 2016-12-11
Feat ex stencil

fmoessbauer:
Please review this example with regard to:
- simplicity
- correctness (simple blur filter)
- performance
- bugs in DASH

Please do not merge until all bug sections are removed.


dash/examples/ex.11.simple-stencil/main.cpp
```diff
@@ -0,0 +1,221 @@
...
+  draw_circle(data_old, 100, 100, 50);
+
+  dash::barrier();
+  write_pgm("testimg_input.pgm", data_old);
+  dash::barrier();
+
+  for(int i=0; i<niter; ++i){
+    smooth(data_old, data_new);
+    dash::barrier();
+    std::copy(data_new.lbegin(), data_new.lend(), data_old.lbegin());
```
ddiefenthaler: Is there a reason to not use dash::copy? Same bug as in line 36?

dash/examples/ex.11.simple-stencil/main.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+void write_pgm(const std::string & filename, const Array_t & data){
+  if(dash::myid() == 0){
+
+    auto ext_x = data.extent(0);
+    auto ext_y = data.extent(1);
+    std::ofstream file;
+    file.open(filename);
+
+    file << "P2\n" << ext_x << " " << ext_y << std::endl;
```
ddiefenthaler: Is there a reason to use std::endl over '\n'?

dash/examples/ex.11.simple-stencil/main.cpp
```diff
@@ -0,0 +1,221 @@
...
+    for(long y=0; y<ext_y; ++y){
+      auto first = data.begin()+ext_x*y; 
+      auto last  = data.begin()+(ext_x*(y+1));
+
+//      BUG!!!!
+//      dash::copy(first, last, buffer.data());
+
+      for(long x=0; x<ext_x; ++x){
+//        file << buffer[x] << " ";
+        file << static_cast<int>(data[x][y]) << " ";
```
ddiefenthaler: I think it is easier to read if the NArray is accessed like this,
dash::copy would be preferred for perfomance reasons, wouldn't it?

fuerlinger:
Looks good!

A few notes: 

Instead of a deep copy of the two arrays, just switch pointers or references, for example similar to this: 
````
  for( int i=0; i<30; i++ ) {
    dash::Array<int>& prev = i%2? arr1 : arr2;
    dash::Array<int>& curr = i%2? arr2 : arr1;

    foo(prev, curr);
  }
````

Data access is currently done element-wise in global-view using `.at()`, right? That will be slow. Performance should be much better when using `.local`. For the inner portion of each domain the code should be almost identical, the border area will be trickier and there are several options how to do it. Asynchronously fetching a whole border column row/column will probably perform best.

File: dash/examples/ex.11.simple-stencil/main.cpp [hunk 4]
```diff
@@ -0,0 +1,237 @@
...
+  }
+  dash::barrier();
+}
+
+void set_pixel(Array_t & data, index_t x, index_t y){
+  const element_t color = 1;
+  auto ext_x = data.extent(0);
+  auto ext_y = data.extent(1);
+
+  x = (x+ext_x)%ext_x;
```
ddiefenthaler, line 72: What are the reasons for this computation,
if x is expected to be larger than ext_x I would just use `x%ext_x`

dash/examples/ex.11.simple-stencil/main.cpp
```diff
@@ -0,0 +1,237 @@
...
+  // Check who owns center, owner draws
+  if(!data.at(x0, y0).is_local()){
+    return;
+  }
+
+  int       f     = 1-r;
+  int       ddF_x = 1;
+  int       ddF_y = -2*r;
+  index_t   x     = 0;
+  index_t   y     = r;
```
ddiefenthaler, line 88: Explanations what these variables mean would be nice.

dash/examples/ex.11.simple-stencil/main.cpp
```diff
@@ -0,0 +1,237 @@
...
+    set_pixel(data, x0+x, y0-y);
+    set_pixel(data, x0-x, y0-y);
+    set_pixel(data, x0+y, y0+x);
+    set_pixel(data, x0-y, y0+x);
+    set_pixel(data, x0+y, y0-x);
+    set_pixel(data, x0-y, y0-x);
+  }
+}
+
+template<typename Array_t>
```
ddiefenthaler: Why is this function templated and the other ones not?

ddiefenthaler (commented):
I left some comments in the code,
overall the program looks structured enough to be simple to understand,
nevertheless doxygen commentaries for the functions would be nice.

dash/examples/ex.11.simple-stencil/main.cpp
fmoessbauer: Not really. Its just to flush the stream after the header.

dash/examples/ex.11.simple-stencil/main.cpp
fmoessbauer: Yes, of course, but that is extremly slow. I intended to fetch a whole row asynchronously, but there is a bug in dash::copy which asserts that 0 elements will be copied. However I manually checked the GlobIter and the distance is definitly not 0.

Maybe @fuchsto can have a look at. I was not able to reproduce this in a similar unit test.

dash/examples/ex.11.simple-stencil/main.cpp
fmoessbauer: I stumbled upon this too. X might be negative and hence I have to use modulo twice to simulate the torus surface.

dash/examples/ex.11.simple-stencil/main.cpp
fmoessbauer: Agree on that. I just copied it from the parallel computing assignment.

dash/examples/ex.11.simple-stencil/main.cpp
fmoessbauer: Good Catch. Change it if you like.

dash/examples/ex.11.simple-stencil/main.cpp
fmoessbauer: The interface of dash::copy does not support copying a GlobIter to another GlobIter.

fmoessbauer: And this should be changed to @fuerlinger's suggestion as well.

dash/examples/ex.11.simple-stencil/main.cpp
ddiefenthaler: I don't think that this is really useful in this case,
after all only one process is writing to the file.

dash/examples/ex.11.simple-stencil/main.cpp
ddiefenthaler: Thanks for the explanation, that makes sense.

fuchsto:
Yes, `dash::copy` does not support GlobalRange to GlobalRange yet. This should just need an adapter, will open an issue for that and also look at your stencil use cases.

fuchsto:
@fmoessbauer @ddiefenthaler 
Is this PR ready to merge? If so, do so.

fmoessbauer:
I'm going to merge it this evening. There are still some Bugs in it.

@fuerlinger : The computation intense part (the inner cell) is calculated using pointers to local memory. Only for initially setting the circles and for calculating the boundary area, global access is used.
However this part is still unoptimized in hope, I can use the stencil and halo features from @dhinf 

@fuchsto The bug in `dash::copy` (line 37) is a different one. This should definitely work.

fuchsto:
@fmoessbauer The halo- / stencil sutff in development is my original implementation, the modifications presented by @dhinf are in branch `feat-halo`, don't know about the difference, though.

Issue #195, opened 2016-12-11
Usage of Halo and Stencil in dash


fmoessbauer:
Hi,

for a student project, I am currently writing a stencil application.
Can anybody tell me how to use `dash::HaloBlock` and stencils? There is some code and documentation in dash, however for me it is not clear how to create a HaloBlock from a `dash::NArray`.

From the meeting in Munich I thought this is supported by dash.

Halo Blocks would also be nice for the [astro tutorial](https://github.com/dash-project/dash-apps/tree/master/dash_tutorial_2017-01).

fuchsto:
Let's have a Skype session on that. The actual missing part is automatic packing of boundary regions. We don't have it yet, but HaloBlock will simplify it a lot.

fuchsto:
... oh, yes: the changes by @dhinf are not merged in development

PR #197, opened 2016-12-12
Feat halo

fmoessbauer:
Merge halo implementation of @dhinf in development (#195). This feature is especially needed in #196 and the [astro tutorial](https://github.com/dash-project/dash-apps/issues/1).

Please review carefully and also check if the tests run on a "real" cluster. Not just on online CIs, as there at max 3 units are spin up.


fmoessbauer:
Unit tests are currently missing as well. There is a TODO in `GlobStencilIterTest.cc` to adopt it to the HaloMatrix.

According to my stencil example, I assume that something with the boundary region is not correct. Unfortunately I do not know, whether the bug is in my code or in DASH.
The pictures are generated using example 11 in branch [feat-halo-ex-stencil](https://github.com/dash-project/dash/tree/feat-halo-ex-stencil) with two units. The suspicious parts are marked red.

### ex.11.halo-stencil
testimg_output_halo

### ex.11.simple.stencil
testimg_output_simple

fuchsto:
Great! Will review all changes back at the hotel and put my comments here.

dash/include/dash/HaloMatrix.h
```diff
@@ -0,0 +1,259 @@
...
+//#include <dash/Allocator.h>
+#include <dash/iterator/HaloMatrixIterator.h>
+#include <dash/Halo.h>
+#include <dash/Matrix.h>
+#include <type_traits>
+
+namespace dash {
+
+template<typename MatrixT, typename HaloSpecT>
+class HaloMatrix
```
fuchsto: The design of `dash::HaloMatrix` violates a lot of concepts and the container definition.
For example, `ibegin`, `iend` and `bbegin`, `bend` should be views, like `dash::HaloMatrix.boundary.begin|.end`.
Methods `updateHalo`, `waitHaloAsync` etc. are obviously not container concerns, but algorithms.
We would have to re-implement those for every container that supports boundary regions.
I'm okay with putting this implementation in examples, but we can't publish it as a container type.

fuchsto (changes requested):

dash/include/dash/HaloMatrix.h
fmoessbauer: Furthermore it's cumbersome to use it, as no access to the underlying container is provided. Hence the halo has to be created on each iteration (see [here](https://github.com/dash-project/dash/blob/feat-halo-ex-stencil/dash/examples/ex.11.halo-stencil/main.cpp#L120)).

For me, a halo is a prime example for the decorator pattern, because the halo adds features to a container. It's not really an algorithm. However halos should support all containers with block views.

fuchsto: @fmoessbauer Yes, that was my intention with `dash::HaloBlock` exactly. It's just a view on a regular NArray block. `dash::HaloBlock` then again represents a sequence container.

See:
- HaloBlock: \
  https://github.com/dash-project/dash/blob/development/dash/include/dash/Halo.h#L1274
- BlockBoundaryView: \
  https://github.com/dash-project/dash/blob/development/dash/include/dash/Halo.h#L1064
- Unit tests of `GlobStencilIter`, `HaloBlock`: \
  https://github.com/dash-project/dash/blob/development/dash/test/GlobStencilIterTest.cc#L308

The unit tests demonstrate the usage of halo views like:

~~~c++
auto boundary_begin = halo_block.boundary().begin();
auto boundary_end   = halo_block.boundary().end();
auto halo_begin = halo_block.halo().begin();
auto halo_end   = halo_block.halo().end();
// north halo region:
auto h_region_n = halo_block.halo_region({ -1,  0 });
auto h_region_n_begin = h_region_n.begin();
auto h_region_n_end   = h_region_n.end();
~~~

fuchsto: Hrrrmmm, I had a variant of `dash::Matrix` where the block view type could be specified as a template parameter. Didn't migrate the branch with the implementation to github, apparently.

fuchsto: Okay, how about moving `dash::HaloMatrix` by @dhinf to the stencil example directory and merging? I want the commits by @dhinf in development so his contribution is visible. We will then open a separate issue on refactoring it.

fmoessbauer: I prefer moving it to `dash/experimental`, as already two examples are using the halo stuff. `dash/experimental` won't be included in `libdash.h` so whenever a user dares to use it, he has to include ` dash/experimental/...` explicitly.
Furthermore moving it to examples will give name clashes in the future.

fuchsto: Eh, "experimental" is for concept proposals actually, but it's okay, will refactor it some time soon.

fmoessbauer:
IMO it can be merged.

fuchsto:
The original implementations of `dash::Halo*` and `dash::GlobStencilIter` are considered stable and should **by no means** be overwritten by the altered versions (now placed in experimental namespace).

PR #199, opened 2016-12-14
added doxygen comments style - review please

BenjaProg:
added doxygen comments for ex.02.array-copy
changed start_index default value in ex.02.array-copy for better clearity in
output

Please give me some feedback to array-copy:
- are additional comments in the code okay and done correctly (in regard to doxygen)
- changed start_index default value from 0 to 10 because the default output were only zeros. 
  Which makes sense of course but I think it is clearer if at least two IDs are printed


fuchsto (commented):
Yup!

Issue #203, opened 2016-12-15
dash::transform segmentation fault with unsigned char


BenjaProg:
Is dash::transform<type> intended to be used with unsigned char?
It works for me with char, int and unsigned int (testet so far) but not with unsigned char, there I get a segmentation fault. 

That's the function call:
~~~c++
  if (0 != myid) {
    dash::transform<T>( histo.lbegin  () ,
                        histo.lend    () ,
                        histo.begin   () ,
                        histo.begin   () ,
                        dash::plus<T> ());
  }
~~~

fmoessbauer:
I will have a look at it tomorrow. In the meantime you can debug it with @ddiefenthaler if you like.

ddiefenthaler:
When calling dash::transform it passes the given arguments down to dart.
This also includes the type which happens in [Line 41](https://github.com/dash-project/dash/blob/development/dash/include/dash/algorithm/Transform.h#L41) of Transform.h
As of `dash::dart_datatype` is currently not specialized for `unsigned char`,
`DART_TYPE_UNDEFINED` is passed to `dart_accumulate` which then segfaults.

* One possible solution would be to simply specialize `dash::dart_datatype<unsigned char>`,
but this may fail as there is no appropriate value in `dart_datatype_t`.
* On the other hand it would be nice to move this error to compile time, if specializing is not possible.

BenjaProg:
[fast fix worked out with ddiefenthaler](https://github.com/dash-project/dash/commit/6033d10f9345f7a02eb37162fb20f7bb39ec0b09)

fmoessbauer:
Opened PR #204 to review and fixed order of types.

PR #204, opened 2016-12-16
Feat 203 dart type u char

fmoessbauer:
Add `unsigned char` to dart types.


BenjaProg:
Thanks for the reordering.
In my mind it was ordered just as you did now, but I was indeed to tired yesterday, if that went wrong.

ddiefenthaler:
@fuchsto When reviewing this pull request please consider Issue #203 as well.
I'm not sure if this solution is really the preferred one,
as I can't judge if anything will break in DART when passing a unsigned char as `DART_TYPE_BYTE`

PR #210, opened 2016-12-20
Use a copy of MPI_COMM_WORLD to avoid conflicts on global collective operations

devreal:
The copy is named DART_COMM_WORLD and is meant for internal use only.

### DIFF
--- a/dart-impl/base/src/locality.c
+++ b/dart-impl/base/src/locality.c
@@ -33,13 +33,13 @@
 
 #define DART__BASE__LOCALITY__MAX_TEAM_DOMAINS 32
 
-dart_host_topology_t *
+static dart_host_topology_t *
 dart__base__locality__host_topology_[DART__BASE__LOCALITY__MAX_TEAM_DOMAINS];
 
-dart_unit_mapping_t *
+static dart_unit_mapping_t *
 dart__base__locality__unit_mapping_[DART__BASE__LOCALITY__MAX_TEAM_DOMAINS];
 
-dart_domain_locality_t *
+static dart_domain_locality_t *
 dart__base__locality__global_domain_[DART__BASE__LOCALITY__MAX_TEAM_DOMAINS];
 
 /* ====================================================================== *
--- a/dart-impl/mpi/include/dash/dart/mpi/dart_team_private.h
+++ b/dart-impl/mpi/include/dash/dart/mpi/dart_team_private.h
@@ -144,6 +144,9 @@
 
 extern dart_team_t dart_next_availteamid;
 
+extern MPI_Comm dart_comm_world;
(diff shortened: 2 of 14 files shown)


fmoessbauer:
@devreal: CircleCI reports a segfault in the openmpi container.

devreal:
Looking at it right now.

devreal:
This is a nasty one, Valgrind reports:

```
==6151== Invalid free() / delete / delete[] / realloc()
==6151==    at 0x4C2EDEB: free (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==6151==    by 0x70B1FC: dart__base__locality__delete (locality.c:246)
==6151==    by 0x6F6F19: dash::Team::unregister_team(dash::Team*) (Team.h:530)
==6151==    by 0x6F71F8: dash::Team::~Team() (Team.h:171)
==6151==    by 0x6B3FFF7: __run_exit_handlers (exit.c:82)
==6151==    by 0x6B40044: exit (exit.c:104)
==6151==    by 0x6B26836: (below main) (libc-start.c:325)
==6151==  Address 0x99cbe0 is 0 bytes inside data symbol "ompi_mpi_comm_null"
```

Not sure where `ompi_mpi_comm_null` comes from but I will make my way through it...

devreal:
I think it is related to a incomplete move assignment/ctor of dash::Team, currently trying a fix.

devreal:
OK, so the reason is that `DART_TEAM_NULL` (aka `-1`) is passed to `Team::unregister_team`, which then unconditionally calls `dart__base__locality__delete` on the DART team ID. I have not yet figured out where this null team comes from, the design is way too complex imo and there are `DART_TEAM_NULL` flying around everywhere. I already started getting rid of them where possible but I am not there yet. 

I think it has been pure luck this bug didn't show up earlier...

devreal:
After half day of debugging this, I found the root cause of the problem: the fields `dash::Team::_team_null` and `dash::Team::_team_all` are _actually destructed on exit_, leading to a call to `Team::unregister_team` without checking for `DART_TEAM_NULL`, leading to `-1` being used as an index... This should be fixed now, along with some other things I came across on my journey. 

Nevertheless, I noticed another point: The way we currently use `dash::Team::_team_[null|all]` is dangerous since in several places they are passed as reference to the caller, i.e., `TeamLocality::team()`, who is then able to alter these teams. Of course, the user should not be able to alter these invariants! Fixing this is not easy, since pointers to Teams are used in several places. Ideally, these invariants would be `constexpr`, which is hard to do with the current implementation of the ctor...

fmoessbauer:
Such a tiny change but this amount of debugging... Thank you for taking your time into this. :+1: I will make a separate ticket for refactoring and cleanup of `dash::Team`

PR #215, opened 2016-12-21
Replace SIZE_MAX with C++-equivalent 

mwernerds:
std::numeric_limits<size_t>::max() for better compiler support (didn't on gcc4.7, now it works)
Thanks for the great project,
   Martin


devreal (changes requested):
Please add brackets for improved readability, i.e., 

((std::numeric_limits<size_t>::max()) >> sh)

PR #218, opened 2016-12-21
Feat hdf5 generic

fmoessbauer:
- Add support for custom datatypes in hdf5 IO
- Cleanup and preparation for views
- many small fixes

### DIFF
--- a/dash/include/dash/io/hdf5/IOManip.h
+++ b/dash/include/dash/io/hdf5/IOManip.h
@@ -80,6 +80,18 @@ public:
   { }
 };
 
+class type_converter {
+private:
+    type_converter_fun_type _converter;
+public:
+    type_converter(type_converter_fun_type fun = get_h5_datatype<UNSPECIFIED_TYPE>)
+    : _converter(fun) { }
+
+    operator type_converter_fun_type(){
+        return _converter;
+    }
+};
+
 } // namespace hdf5
 } // namespace io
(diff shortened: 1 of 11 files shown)


fuchsto:
Yesyes, that should work!

Ha, well, *sigh*, there indeed *does* exist a set of formally very attractive abstractions that would not only work but even turn our distributed + async + formatting concepts in a piece of mathematical art, but the "boilerplate" is rather hefty.

Well, feast your eyes upon the proposed view-, range- and stream iterator (!) concepts:
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4560.pdf

I'm gradually transforming our iterator, view and domain types to these, but it will take some time.

fuchsto (commented):
@fmoessbauer 
Yesyes, that should work!

Ha, well, \*sigh\*, there indeed *does* exist a set of formally very attractive abstractions that would not only work but even turn our distributed + async + formatting concepts in a piece of mathematical art, but the "boilerplate" is rather hefty.

Well, feast your eyes on the proposed view-, range- and stream iterator (!) concepts:
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4560.pdf

I'm gradually transforming our iterator, view and domain types to these, but it will take some time.

Oh, here's a reference implementation. Enjoy!
https://github.com/ericniebler/range-v3

PR #224, opened 2016-12-27
fixed return type of MatrixRef.team()

fmoessbauer:


fuchsto (commented):
Curious that this didn't come up earlier.

Issue #228, opened 2017-01-04
TeamSpec::balance_extents() only operates on 2 dimensions


ddiefenthaler:
When looking at balance_extents in [TeamSpec.h](https://github.com/dash-project/dash/blob/dd42d162a1ed0e80c934efe94c8c1977d1317a6b/dash/include/dash/TeamSpec.h#L233),
I noticed that it only balanced the extents in the first two dimensions.

As I'm probably going to need this for more dimensions in the future,
I'll try to rewrite it so that it gets balanced in all dimensions. See pull request #229.

Apart from that the current algorithm doesn't seem to be ideal,
for example 144 units get balanced to 16x9 instead of 12x12.

fuchsto:
@ddiefenthaler Yes, that is the expected behavior. Note that the balancing is constrained with hardware topology domain sizes.
For example, with 4 sockets and 2 NUMA domains in the hardware topology, the algorithm optimizes surface-to-volume ratio by balancing extents constrained to even multiples of 8.

ddiefenthaler:
Those constraints are interesting.
I'm quite sure that balance_extents in the current development branch does not consider any topology information. (Current implementation: [TeamSpec.h](https://github.com/dash-project/dash/blob/dd42d162a1ed0e80c934efe94c8c1977d1317a6b/dash/include/dash/TeamSpec.h#L233))
Another example, the first one might be poorly chosen: 225 units get balanced to 9x25 instead of 15x15.

fuchsto:
Addressed in #229

PR #229, opened 2017-01-04
Rewritten TeamSpec::balance_extents() to balance more than two dimensions

ddiefenthaler:
I've rewritten balance_extents to use a different algorithm.


ddiefenthaler:
The idea behind the algorithm is to equally distribute the prime factors across the dimensions as good as possible. Starting with the largest factors they get multiplied with `this->extents[i]` with `i` growing.
As long as there are enough remaining prime factors this condition should be true:
`this->_extents[0] >= this->_extents[1] >= ... >= this->_extents[MaxDimensions-1]`
If the last iteration over `this->_extents` is incomplete the direction gets reversed to even the effect out. (Starting with `this->_extents[MaxDimensions-1]` downwards)

Thinking about it, it may make sense to change the direction each iteration over `this->_extents`,
but I'm not sure about this... I can't see what would happen after the second iteration.

fuchsto (changes requested):
Thank you, but please implement additional unit tests to cover the new functionality.

fuchsto:
Addresses #228

ddiefenthaler:
I added a unit test,
but I think that the algorithm should still be changed once more.
I assume that my approach only works well if the number of units factorizes in not to many different factors.
Apart from that the algorithm does not consider any constraints.
@fuchsto It would be nice if you could clarify how balance_extents should behave.

ddiefenthaler:
I added another commit, I think this is in a mergable state now.

I am sure there are more optimizations possible, but for now this works too.

I removed the adjustment of `_is_linear`, that led to `MatrixTest.CopyRow` failing,
but I think this is another error so I will open a seperate issue for that.

ddiefenthaler:
I think this Pull request is fine now,
@fuchsto can you review this again?

dash/include/dash/TeamSpec.h
```diff
@@ -218,39 +219,56 @@ class TeamSpec :
...
-          this->_extents[0] = extent_x;
-          this->_extents[1] = extent_y;
-        }
+    // Equally distribute factors to extents.
+    // Start with the largest factors and multiply them onto the lowest value
+    for (auto it = teamsize_prime_factors.rbegin(); it != teamsize_prime_factors.rend(); ++it) {
+      DASH_LOG_TRACE("TeamSpec.balance_extents()",
+                     "factor:", it->first, "x", it->second);
+      for (auto i = 1; i < it->second + 1; ++i) {
+        new_extents.insert(it->first * *new_extents.begin());
```
fmoessbauer, line 259: You are not a fan of brackets, are you? ;)

fuchsto (approved):

fuchsto:
@ddiefenthaler Looking good!

fuchsto:
@ddiefenthaler I don't care too much about efficiency of this implementation for now. Integer factorization is NP-hard, anyways.
This function is supposed to be called once for a reasonably large unit of work, as in domain decomposition for a 2048x2048 matrix-matrix multiplication. Even if it was inefficient, it would not notably affect the overall time to completion.

PR #232, opened 2017-01-06
Feat 231 ccwarn

fuchsto:


dash/include/dash/internal/Logging.h
```diff
@@ -20,10 +20,6 @@
 #include <sys/types.h>
 #include <unistd.h>
 
-namespace dash {
-  // forward-declaration
-  global_unit_t myid();
-}
```
fmoessbauer, line 23: That cannot be removed, as logging is used in Team.cc and vise versa.
Error (gcc5):
```bash
In file included from /opt/dash/dash/include/dash/util/Timer.h:20:0,                                  │ 1 file changed, 4 insertions(+)
                 from /opt/dash/dash/src/util/TimestampClockPosix.cc:1:                               │ felix@HIGHPOWER  ~/repos/github-dash/dash   feat-hdf5-views ●  git commit -a
/opt/dash/dash/include/dash/internal/Logging.h: In function 'void dash::internal::logging::Log_Line(co│[feat-hdf5-views aaa90a2] skip ASIO test if using mpich, as ASIO unrelated concurrency problems occure
nst char*, const char*, int, const char*, const string&)':                                            │ 1 file changed, 10 insertions(+), 2 deletions(-)
/opt/dash/dash/include/dash/internal/Logging.h:194:29: error: 'myid' is not a member of 'dash'        │ felix@HIGHPOWER  ~/repos/github-dash/dash   feat-hdf5-views  git push
   dash::global_unit_t uid = dash::myid();
```

fmoessbauer (changes requested):
The compiler flag changes look good to me, but then we also should fix all the warnings. Currently my compiler (gcc5) generates a massive amount of warnings.

Request changes: see my comment on `dash::myid()`

dash/include/dash/internal/Logging.h
fuchsto: Right, re-added the forward-declaration.

fuchsto:
@fmoessbauer The "excessive" compiler warnings are only enabled in developer (not debug) build config, so we don't have to fix all of them right now. There just should be a way to make them visible.

fmoessbauer (approved):

Issue #231, opened 2017-01-06
Enable all available compiler warnings


fuchsto:
We currently compile with warning flags `-Wall -Wpedantic -Wextra` which is basically reasonable, however g++ offers several additional warning options that are not enabled with this configuration.

rkowalewski:
Which one for example do you mean?

EDIT: Got it. I assume you refer to Pull request #232

devreal:
Thanks for the additional flags. While this is only a developer option, it greatly helps to have a clean build again soon. Otherwise, more critical warnings might go unnoticed. I will start fixing the DART part soon.

fuchsto:
@devreal Thank you, but don't give it high priority.

devreal:
[UPDATE] Submitted comment before finished writing, classical slipped-off-my-mouse mistake ^^ [/UPDATE]

Just out of curiosity, I ran a make and looked at the new warnings. Some seem to be pure style-related, e.g.,
```
dart_team_private.c:79:5: warning: function declaration isn’t a prototype [-Wstrict-prototypes]
 int dart_adapt_teamlist_destroy ()
```
Here, a `void` is missing in the parentheses. 

Others are really strange, e.g., 

```
/home/joseph/src/dash/dash/dart-impl/base/src/locality.c: In function ‘dart__base__locality__group_subdomains’:
/home/joseph/src/dash/dash/dart-impl/base/src/locality.c:954:3: warning: assuming signed overflow does not occur when simplifying conditional [-Wstrict-overflow]
   for (int g = 0; g < num_existing_domain_groups; g++) {
   ^
```

This seems to be some strange behavior of GCC, see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55217

Only the warnings introduced with `-Wdiscarded-qualifiers` seem to be useful at this point:
```
/home/joseph/src/dash/dash/dart-impl/base/src/internal/domain_locality.c: In function ‘dart__base__locality__domain__parent’:
/home/joseph/src/dash/dash/dart-impl/base/src/internal/domain_locality.c:328:17: warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
     *domain_out = (const dart_domain_locality_t *)(domain_in);
```

fuchsto:
That was unintended, the style-related warnings should be disabled.

devreal:
Addendum: `-Wmissing-prototypes` seems to be useful as well (spots functions that could be marked `static`), e.g., 

```
[...]

PR #234, opened 2017-01-13
Exclude warnings for Cray compiler and remove asm dump

devreal:
Fixes build with Clang compiler (does not support the asm dump flags)

See #231


CMakeExt/CompilerFlags.cmake
```diff
@@ -83,7 +103,7 @@ endif()
 set (CC_WARN_FLAG  "${DASH_DEVELOPER_CC_FLAGS}")
 set (CXX_WARN_FLAG "${DASH_DEVELOPER_CXX_FLAGS}")
 
-if (ENABLE_COMPILER_WARNINGS)
+if (ENABLE_DEVELOPER_COMPILER_WARNINGS)
   if (NOT "${CMAKE_CXX_COMPILER_ID}" MATCHES "Cray")
     # Flags for C and C++:
     set (CXX_WARN_FLAG "${CXX_WARN_FLAG} -Wall -Wextra -Wpedantic")
```
fuchsto: The warning flags `-Wall -Wpedantic -Wextra` should be enabled in every configuration, I consider those the "bare minimum".

fuchsto (approved):
Fixed in my recent merge from development

Issue #226, opened 2017-01-03
Extent of underfilled block in BlockPattern is wrong


fmoessbauer:
If a `dash::Pattern` (Blockpattern) has underfilled blocks, currently the following holds:
```
pattern.blocksize(i) == pattern.local_block(last_idx).extent(i);
```
This is not correct if the last block is underfilled. See unit test `BockPatternTest.UnderfilledPatternExtent` for details.

Furthermore, if a unit only has underfilled blocks, `pattern.blocksize(0)` returns the size of the underfilled block, which is not correct because the following should hold:
```
pattern(
    size_spec,
    dash::DistributionSpec<2>(
      dash::TILE(block_size_x),
      dash::TILE(block_size_y)),
      ...);
pattern.blocksize(0) == block_size_x;
```

Branch: `bug-226-pattern`

knuedd:
I think the same bug haunts the tutorial example with a 2D matrix. The sum of the local extents in a dimension is sometimes not the global extent. Also the local iterators iterate over too many elements if the last block in a dimension is under-filled.

I tried to show that behavior in branch bug-226-pattern in a new test case MatrixTest.UnderfilledBlockedPatternExtents but there some asserts tell me, that the global extents should be a multiple of the number of units.

Below there is an example code that demonstrates this behavior:

    #include <unistd.h>
    #include <iostream>
    #include <cstddef>
    #include <iomanip>
    #include <cstdlib>

    using std::cout;
    using std::endl;
    using std::setw;
    using std::flush;

    #include <libdash.h>

    int main( int argc, char* argv[] ) {

        dash::init( &argc, &argv );

        dart_unit_t myid= dash::myid();
        size_t numunits= dash::Team::All().size();
        dash::TeamSpec<2> teamspec( numunits, 1 );
        teamspec.balance_extents();

        uint32_t w= 13;
        uint32_t h= 7;
        auto distspec= dash::DistributionSpec<2>( dash::BLOCKED, dash::BLOCKED );
        dash::Matrix<uint32_t, 2> matrix( dash::SizeSpec<2>( h, w ),
            distspec,
            dash::Team::All(), teamspec );

        matrix.barrier();

        if ( 0 == myid ) {
            cout << "global extent is " << matrix.extent(1) << " x " << matrix.extent(0) << 
                " == " << matrix.extent(1) * matrix.extent(0) << endl << endl << flush;
        }

        matrix.barrier();
        
[...]

PR #235, opened 2017-01-17
Bug 226 pattern

fmoessbauer:
Regards #226


fuchsto (commented):
Fixed unit test for default_index_type int

PR #236, opened 2017-01-17
Major cleanup of includes in test sources

fuchsto:
- To reduce compile time of tests after changes in DASH, unit tests now only include dependent headers directly instead of `libdash.h`.
- Some header files in DASH have been adjusted likewise.
- All test fixtures now subclass `dash::test::TestBase` to reduce duplicate code and facilitate extensibility.
- Also, `dash::internal::accumulate_*impl` has been moved to `dash/Transform.h` and renamed to `dash::internal::transform_*impl`.

No implementation should be affected by this PR.

### DIFF
--- a/dash/test/HDF5ArrayTest.h
+++ b/dash/test/HDF5ArrayTest.h
@@ -3,25 +3,18 @@
 
 #ifdef DASH_ENABLE_HDF5
 
-#include <gtest/gtest.h>
-#include <libdash.h>
-
 #include "TestBase.h"
 
 // for CustomType test
 #include "hdf5.h"
 
 
-class HDF5ArrayTest : public ::testing::Test {
+class HDF5ArrayTest : public dash::test::TestBase {
 protected:
-  dash::global_unit_t _dash_id;
-  size_t      _dash_size;
-  std::string _filename = "test_array.hdf5";
-  std::string _dataset  = "data";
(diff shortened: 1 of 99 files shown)


dash/test/HDF5ArrayTest.h
```diff
@@ -30,25 +23,20 @@ class HDF5ArrayTest : public ::testing::Test {
   }
 
   virtual void SetUp() {
-    dash::init(&TESTENV.argc, &TESTENV.argv);
-    _dash_id   = dash::myid();
-    _dash_size = dash::size();
-    if(_dash_id == 0) {
+    dash::test::TestBase::SetUp();
+
+    if(dash::myid().id == 0) {
```
devreal, line 28: (dash::myid() == 0) should work as well, not?

devreal (changes requested):
Fine with me, thanks for your efforts! Just one comment: there are several places in which `(dash::myid().id == 0)` can be simplified to `(dash::myid() == 0)`.

dash/test/HDF5ArrayTest.h
fuchsto: Yes it does, doesn't make a difference, though.

fmoessbauer (commented):
Wasn't that the job of clemens ? Anyway: Great work!!

Issue #239, opened 2017-01-19
Fix empty allocations


fmoessbauer:
Currently GlobMem allocations hang if at least one unit does not contribute any local elements. This is due to collective MPI operations in DART which are skipped for units without local elements. Furthermore the implementation of memory allocation does not comply to the specifications made in the header.

For details also have a look at my comment [here](https://github.com/dash-project/dash/issues/226#issuecomment-273561852) in issue #226 .

Therefor I recommend to fix the implementation and clarify the specification as follows:

- all allocators have to support empty allocations
- all units of a team take part in a team-wide allocation
- all allocation calls have to be passed to the down-most (DART) allocator
- a dash allocator must not lessen the synchronisation requirements of the down-most allocator

devreal:
I just checked the DART code for skipped collective operations. I cannot see that DART itself skips collectives on empty allocations. So it's just that DASH is skipping collective DART calls on empty allocations?

fmoessbauer:
There are two sides of the problem:
1. DASH skips the collective operation `dart_team_memalloc_aligned`.
2. DART does not allow empty allocations, as `MPI_win_attach` does not work for empty windows (not sure if this is the bug, but it's the error MPI reports). However a simple `if(nelem == 0){++nelem;}` in DART fixes the problem. But this is definitly not the right way to solve the problem.

devreal:
Can you please paste the MPI error message? At least the standard does not seem to enforce a size greater zero for `MPI_Win_attach`.

fmoessbauer:
Executed using the openmpi (v1.10.5) docker container:

```
$>mpirun -n 4 --allow-run-as-root bin/dash-test-mpi
...
[  RUN     ] FindTest.EmptyContainer
...
[cffe0d848854:11978] *** An error occurred in MPI_Win_attach
[cffe0d848854:11978] *** reported by process [1494548481,2]
[cffe0d848854:11978] *** on win pt2pt window 7
[cffe0d848854:11978] *** MPI_ERR_ARG: invalid argument of some other kind
[cffe0d848854:11978] *** MPI_ERRORS_ARE_FATAL (processes in this win will now abort,
[cffe0d848854:11978] ***    and potentially your MPI job)
```

See branch [bug-239-allocate](https://github.com/dash-project/dash/compare/bug-239-allocate) for my implementation.

devreal:
[...]

PR #240, opened 2017-01-19
Bug 239 emptyalloc

devreal:
Fix handling of empty allocations. 

Addresses #239


fmoessbauer:
Looks good and fixes the problems at least on my local machine. I will merge the PR after CI has passed.

PR #243, opened 2017-01-19
added unit test for local viewspec for matrices with underfilled blocks

fmoessbauer:


rkowalewski:
Hmmm your merge was actually too fast. I wanted to add a small minor change to your test. But anyway you can do it on your own:
Please substitute your `elementsvisited` loop count with `std::distance` to eliminate this loop.

fmoessbauer:
But `std::distance` uses a different semantic on random access iterators and even does not throw an assertion if the iterator leaves the iteration space. So maybe it's better to just add `std::distance`, but keep the loop as it is?

rkowalewski:
hm why that? in this case `lbegin()`and `lend()` are pointers and pointers are random access iterators. So they satisfy the requirements of `std::distance` and the semantic translates simply to (size_type *) lend - (size_type *) lbegin. And this is the difference of the
subscripts of the two pointers within the contiguous memory as defined by [ANSI C](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf), section 6.5.6.

It is a problem if the iteration space is non contiguous but in that case our "local" iterators cannot be native pointers.  Or did I miss something?

fmoessbauer:
My point was, that the pointers do not have to point to the allocated memory space. But of course, you are right that this makes absolutely no difference until we do not dereference them and access the memory. I just did not think of them as pointers, but iterators.

fuchsto:
@fmoessbauer @rkowalewski Good job, guys!

PR #247, opened 2017-01-23
Enable dyanmic log filtering 

devreal:
Add support for log levels specified via environment variable DART_LOG_LEVEL=[ERROR|WARN|INFO|DEBUG|TRACE]


rkowalewski:
I made some small minor improvements to avoid duplication in DART_LOG_* macros.

fmoessbauer:
We should add documentation for the logging features.

devreal:
Agreed, will add documentation soon.

devreal:
I think the macros themselves are self-explaining. I have the documented the environment variable and log levels in `dart.h`. It now reads:

> Logging
> 
> DART can be configured to produce log output with different log levels, a feature that is mainly meant for debugging purposes. To enable general logging output, the parameter -DENABLE_DART_LOGGING=ON should be passed to CMake when building DART/DASH. Alternatively, the pre-compiler macro DART_ENABLE_LOGGING can be defined manually. Please note that the additional log output may cause notable performance overhead and should not be enabled for production runs.
> 
> The verbosity of the log output can be controlled at runtime through the environment variable DART_LOG_LEVEL, whose value (if set) controls the maximum log level. Possible values are:
> 
>     DART_LOGLEVEL_ERROR: Emit only messages on errors that are fatal (similar to having logging disabled).
>     DART_LOGLEVEL_WARN: Emit error messages and non-fatal warnings.
>     DART_LOGLEVEL_INFO: In addition to errors and warnings, emit additional information on the execution of the DART library.
>     DART_LOGLEVEL_DEBUG: Issue detailed debugging output on (mostly) all DART methods executed.
>     DART_LOGLEVEL_TRACE: In addition to the above, also output information on the internal state of DART.
> 

This is just a guideline and I'm sure reality is different at this point in time (`DART_LOG_INFO` is not used at all). However, adapting to this can be a gradual transition. 

I'm not sure whether and how we want to translate this to DASH. My feeling is that the log levels of the two should be controlled independently. Introducing an environment variable `DASH_LOG_LEVEL` would be straight-forward if we agree on it and there could be a fall-back to `DART_LOG_LEVEL` if `DASH_LOG_LEVEL` is not set. Let me know what you prefer.

fmoessbauer (approved):

dart-impl/base/src/logging.c
```diff
@@ -22,3 +22,29 @@ const int dart__base__unit_term_colors[DART_LOG_TCOL_NUM_CODES-1] = {
...
+
+
+enum dart__base__logging_loglevel
+dart__base__logging_env_loglevel()
+{
+  static enum dart__base__logging_loglevel level = DART_LOGLEVEL_TRACE;
+  static int log_level_parsed = 0;
+  if (!log_level_parsed) {
+    const char *envstr = getenv(DART_LOGLEVEL_ENVSTR);
+    if (envstr) {
```
fuchsto, line 34: DART should not use `getenv` itself.
The run-time configuration of DART is maintained in `dart_config`, there already is a field `log_enabled` which is used by DASH.
The utiliy class `dash::util::Config` encapsulates the parsing of environment variables and already supports `DASH_ENABLE_LOGGING=OFF` (or =0, =1, =yes, ...) among many other environment settings (enabling tracing, chunk size settings, and the like).

devreal: I've seen the `dart_config` 'interface', which is horrible as it exposes an internal data structure to the world. This might be OK for setting a single value but it breaks pretty much every interface design rule. 

Moreover, considering that DART is meant to be used as a stand-alone runtime system, there should be means of controlling the log level without changing client code, i.e., without having to call DART functions at runtime. I do not see what is wrong with calling `getenv` in DART.  We can implement a way to override the environment setting (if any) through a re-designed `dart_config` interface, though.

fuchsto: I agree that `dart_config` does not present a well-defined interface, but parsing environment settings in several unrelated places is a bad thing to do.
The implementation of `dash::util::Config` is way more comprehensive than the current snippet in DART.
DART should be considered a stand-alone deployable, I agree, so it should provide its own run-time config interface. Parsing `ENV` is not required, however.
If at all, `getopt` should be used for this.

So I propose to extend dart-if by a configuration interface and use it in `dash::util::Config` to apply environment settings.

devreal: > DART should be considered a stand-alone deployable, I agree, so it should provide its own run-time config interface. 

Exactly that is my point: a user who wants to control the log verbosity of the underlying DART runtime would have to do so through API calls, either fiddling in an `ENV` or command line parser for flexibility himself or recompile his application. AFAICS, it is common practice to control the runtime through environment variables, e.g., OpenMP variables and MPI implementation-specific configuration flags.

> If at all, `getopt` should be used for this.

Not at all. `getopt` is used for _application_ specific command line parameters, not to control the underlying runtime. 

> The implementation of `dash::util::Config` is way more comprehensive than the current snippet in DART.

I agree that the parser is simple but it does what it needs to do. There is no fanciness required. There might be more DART-specific environment variables in the future, e.g., the number of threads, so we can build a central abstraction for `ENV` in DART. 

I strongly recommend keeping the DART runtime configuration flexible and independent from DASH. Logging is a good example: If  `DASH_ENABLE_LOGGING` was the only way of controlling log verbosity I can never suppress DASH log info and see full DART trace logs without manually issuing a DART API call.

fuchsto:
Ok, I'm with you.
So we declare the configuration of DART (with environment flag prefix `DART_`) independent from DASH (with environment flag prefix `DASH_`).
That's reasonable. I will review the DASH config to respect this.

Addendum: But the syntax for options should be the same for DASH and DART. So DART should support expressions like

`DART_FOO_SIZE=23k`
`DART_ENABLE_BAR=on`

and the like.

And you're right with `getopt`, of course.

PR #248, opened 2017-01-23
Make dash::Arrays in algorithms team-aware

devreal:
Make `dash::Array`s in algorithms team-aware
Fix includes of `dash/Array.h`
Also make `SharedCounter` and `GlobDynamicMem` team-aware


fmoessbauer (approved):
Good work!

PR #250, opened 2017-01-24
Fixed extended compiler warnings for clang-3.8

fuchsto:


dash/include/dash/TeamSpec.h
```diff
@@ -218,6 +218,7 @@ class TeamSpec :
     DASH_LOG_TRACE_VAR("TeamSpec({extents})", this->_extents);
   }
 
+#if 0
```
devreal: Copy constructor not needed anymore? Why not remove the dead code?

dash/include/dash/matrix/internal/MatrixRefView-inl.h
```diff
@@ -23,28 +23,12 @@ ::MatrixRefView(
...
   _l_viewspec()
 {
-  // TODO: Check if initializing local viewspec with default viewspec is okay.
+  // TODO: Check if initializing local viewspec with default viewspec
+  //       is okay.
   DASH_LOG_TRACE_VAR("MatrixRefView(matrix)", matrix);
 }
 
 template<typename T, dim_t NumDim, class PatternT>
-MatrixRefView<T, NumDim, PatternT>
```
devreal, line 31: I assume the user should not be able to copy a MatrixRefView anymore?

devreal (commented):
Two comments inline. We don't need the copy constructors anymore?

dash/include/dash/matrix/internal/MatrixRefView-inl.h
fuchsto: Why? The copy constructor still exists.

fuchsto:
@devreal @fmoessbauer Bouncing for review

dash/include/dash/matrix/internal/MatrixRefView-inl.h
devreal: I only see two constructors:
```
MatrixRefView<T, NumDimensions, PatternT>();
  MatrixRefView<T, NumDimensions, PatternT>(
Matrix<T, NumDimensions, index_type, PatternT> * matrix);
```

Am I missing something?

fuchsto: There always is a copy-constructor if it is not hidden or deleted.
The copy-constructor is implicitly defined by the compiler otherwise.
The previous definition specified it explicitly, but this was only required for logging and prevented default rule of 5 semantics.

devreal: Dang, of course...

devreal (approved):

dash/include/dash/TeamSpec.h
fuchsto: Yes, not needed. I removed the disabled code section.

fmoessbauer (approved):

Issue #118, opened 2016-11-18
Improve thread-support of DART/DASH


devreal:
The following features in DART and DASH should be improved to support thread-parallel access:

1. Interface to control thread-aware initialization of the underlying communication library (mainly MPI)
2. Communication operations (mainly put/get and atomic operations, collective operations will not be thread-safe due to restrictions in MPI)
3. Global memory allocation (needs protection of internal shared data structures in DART) 
4. Team management may not be required to be thread-safe but we need to ensure not to use in potentially thread-parallel code-paths
5. Improve documentation of which operations or operation combinations are thread-safe and which are not

See also #109.

fuchsto:
I will take care of the global memory allocation and team management aspects.
We also should think about meaningful tests / verification measures.

There is a light-weight, convenient unit test framework for thread-parallel test cases called *partest* from the [siemens/embb](https://github.com/siemens/embb) ecosphere that I really like.
I don't know if it is readily available to include it in distribution, but it would be great if we could integrate it in our "private" CI at least.

Oh, and there is *Divine* for explicit state model checking, but I found it only practicable for very minimal, stripped-down use cases. Even then, verification usually takes the better part of a day.

(Then again, we have lots of CPU hours to spare on SuperMUC)

devreal:
[UPDATE: Added third point]

I created a branch (`bug-118-threadsafety`) to deal with this and started investigating the road-blocks in the global memory allocation on the DART side. There are two issues right now:

1. The segment management is using a global shared hash table that is used to map segment IDs from DART `gptr`s to the segment information. There are three possible solutions:
    - Use lock-free lists inside the hash table buckets. This can be done using `_Atomic` in C11 but will cause some pain. 
    - Use (spin-)locks to protect the global data-structure.
    - Since we only have 16-bit segment IDs and they are not reused we could also pre-allocate information for all segments, which would be around 3MByte with the current data-structure. That is the easiest, fastest, but also least efficient method ^^

2. The buddy allocator used for non-aligned allocations is not thread-safe. See #220 

3. There are some integral shared state variables as well. In the case of global memory, it's `dart_memid`. They can be handled through atomics. I will start implementing an abstraction for integral atomic operations after the holidays.

fuchsto:
Fortunately, these issues are anything but exotic. These are classic parallel memory management problems.
Not too long ago I was lucky enough to be involved in a project at Siemens CT which addresses these challenges for embedded systems, called EMBB.
[...]

PR #252, opened 2017-01-25
Improve thread-safety in DART

devreal:
Improve thread-safety of certain DART operations and provide an interface for initialization of a thread-safe runtime. 

Collective allocation and communication are now marked as thread-safe given that they are not called on the same team concurrently.

Addresses #118


dart-impl/base/include/dash/dart/base/atomic.h
```diff
@@ -0,0 +1,42 @@
...
+ * See also the file kmp_os.h of the Clang OpenMP library, which served as inspiration.
+ *
+ * TODO: Check that all compilers support these intrinsics, esp. Cray?
+ */
+
+#ifndef DASH_DART_BASE_ATOMIC_H_
+#define DASH_DART_BASE_ATOMIC_H_
+
+
+#define DART_FETCH_AND_ADD64(ptr, val)   __sync_fetch_and_add((int64_t *)(ptr), (val))
```
fuchsto: These definitions need a guard checking for availability of the atomic compiler extensions, something in the line of `HAVE_SYNC_FETCH_AND_ADD`.

For a from-scratch implementation, see:
- https://github.com/siemens/embb/blob/master/base_c/src/atomicfunc_32.asm
- https://github.com/siemens/embb/blob/master/base_c/src/atomicfunc_64.asm

Also, please reformat the code for 80 columns.

build.sh
```diff
@@ -59,6 +59,7 @@ rm -Rf $BUILD_DIR/*
                         -DDART_IF_VERSION=3.2 \
                         -DINSTALL_PREFIX=$HOME/opt/dash-0.3.0/ \
                         -DDART_IMPLEMENTATIONS=mpi \
+                        -DENABLE_THREADING=ON \
```
fuchsto: Please decide for either `ENABLE_THREADING` or `ENABLE_MULTITHREADING`.

dart-if/v3.2/include/dash/dart/if/dart_types.h
```diff
@@ -215,6 +215,18 @@ typedef int32_t dart_team_t;
...
+typedef enum
+{
+  /** No support for thread-based concurrency in DART is provided. */
+  DART_THREAD_SINGLE = 0,
+  /**
+   * Support for thread-based concurrency is provided by DART and
+   * the underlying runtime.
+   */
+  DART_THREAD_MULTIPLE = 10
+} dart_thread_level_t;
```
fuchsto: Puns like "thread level" have my heartfelt support, but perhaps we should name it `dart_thread_support_level_t`?

dash/include/dash/algorithm/SUMMA.h
```diff
@@ -121,6 +121,27 @@ typedef dash::pattern_layout_properties<
...
+        typename MatrixType::pattern_type>;
+
+template<
+  typename MatrixTypeA,
+  typename MatrixTypeB,
+  typename MatrixTypeC
+>
+using check_summa_pattern_constraints =
+    typename std::enable_if<
+      summa_pattern_constraints<MatrixTypeA>::satisfied::value &&
```
fuchsto: Should be `MatrixTypeA`, `MatrixTypeB`, `MatrixTypeC` (see lines below)

dash/include/dash/algorithm/SUMMA.h
```diff
@@ -121,6 +121,27 @@ typedef dash::pattern_layout_properties<
             dash::pattern_layout_tag::linear
         > summa_pattern_layout_constraints;
 
+template<typename MatrixType>
+using summa_pattern_constraints =
```
fuchsto: This only works for the current `N x N` implementation of SUMMA. The apparently redundant original definition served as a reminder that pattern constraints may depend on each other.
This simplification is okay for now, but it should not be in a public namespace. Please move it to `dash::internal` or `dash::detail`.

dart-impl/base/include/dash/dart/base/atomic.h
```diff
@@ -0,0 +1,42 @@
...
+#define DART_FETCH_AND_ADD32(ptr, val)   __sync_fetch_and_add((int32_t *)(ptr), (val))
+#define DART_FETCH_AND_SUB64(ptr, val)   __sync_fetch_and_sub((int64_t *)(ptr), (val))
+#define DART_FETCH_AND_SUB32(ptr, val)   __sync_fetch_and_sub((int32_t *)(ptr), (val))
+
+#define DART_FETCH_AND_INC64(ptr)        __sync_fetch_and_add((int64_t *)(ptr), 1LL)
+#define DART_FETCH_AND_INC32(ptr)        __sync_fetch_and_add((int32_t *)(ptr), 1)
+#define DART_FETCH_AND_INC16(ptr)        __sync_fetch_and_add((int16_t *)(ptr), 1)
+#define DART_FETCH_AND_INC8(ptr)         __sync_fetch_and_add((int8_t  *)(ptr), 1)
+#define DART_FETCH_AND_INCPTR(ptr)       __sync_fetch_and_add((void   **)(ptr), sizeof(**ptr))
```
fuchsto: `sizeof(**ptr)` should be `sizeof(**(ptr))`, same in line 34.

fmoessbauer (changes requested):
I cannot see any unit tests of the thread safety features. Our CI containers do have MPI libs with thread support enabled. For the sake of completeness I added the docker scripts to this branch.

devreal:
I added three test-cases, most importantly testing concurrent memory allocation on different teams and simple concurrent put: https://github.com/dash-project/dash/blob/bug-118-threadsafety/dash/test/ThreadsafetyTest.cc

Feel free to shoot me more ideas you might have :)

fmoessbauer (approved):

fuchsto (changes requested):

build.sh
devreal: I ended up with ENABLE_THREADSUPPORT, which imo describes the semantics best.

dart-impl/base/include/dash/dart/base/atomic.h
devreal: I added a guard flag and an unsafe alternative. CMake errors out if thread-support is requested but the builtins are not available. 

I am aware of the EMBB code but I would like to avoid pulling in assembler code by all means because it is just not portable. Just think of moving to ARM... The `__sync` builtins are supported by all compilers we currently support. Ideally we would use the `__atomic` builtins or C11 `atomic`s (at some point) but at least the latter is still not supported by the Intel compiler at this point.

devreal:
@fuchsto I think I addressed all of your points. Please review and let me know if you have any other points :)

fuchsto:
@devreal On it, right after lunch!

dart-impl/base/include/dash/dart/base/atomic.h
fuchsto: I agree! We def' do not want to ship assembly (If we do some day: I'm actually way better at ARM asm than x86)

build.sh
fuchsto: Splendid!

dart-impl/base/include/dash/dart/base/atomic.h
```diff
@@ -0,0 +1,372 @@
...
+#define DART_DEC_AND_FETCH64(ptr)    \
+          __sync_sub_and_fetch((int64_t *)(ptr), 1LL)
+#define DART_DEC_AND_FETCH32(ptr)    \
+          __sync_sub_and_fetch((int32_t *)(ptr), 1)
+#define DART_DEC_AND_FETCH16(ptr)    \
+          __sync_sub_and_fetch((int16_t *)(ptr), 1)
+#define DART_DEC_AND_FETCH8(ptr)     \
+          __sync_sub_and_fetch((int8_t  *)(ptr), 1)
+#define DART_DEC_AND_FETCHPTR(ptr)   \
+          __sync_sub_and_fetch((void   **)(ptr), sizeof(**ptr))
```
fuchsto: Should be `sizeof(**(ptr))`
But don't bother, can fix this myself before merging.

dash/test/ThreadsafetyTest.cc
```diff
@@ -0,0 +1,131 @@
...
+  {
+    int thread_id = omp_get_thread_num();
+    for (int i = 0; i < 100; ++i) {
+      // thread 0: contribute to allocation on team_all
+      if (thread_id == 0) {
+        if (i) {
+          arr_all.deallocate();
+          ASSERT_EQ_U(arr_all.size(), 0);
+        }
+        arr_all.allocate(team_all.size(), dash::DistributionSpec<1>(), team_all);
```
fuchsto: Increasing the size of the `dash::Array` instance might improve the probability to catch synchronization bugs.

dash/test/ThreadsafetyTest.cc
```diff
@@ -0,0 +1,131 @@
+
```
fuchsto, line 1: There are no tests of the introduced atomic operations (`DART_COMPARE_AND_SWAP*` etc.), right?

fuchsto (changes requested):
Looks fine for me, but unit tests should be way more comprehensive.

devreal:
I will try to think of and create some more unit tests that could trigger errors.

dash/test/ThreadsafetyTest.cc
devreal: Correct, mainly because they are only used internally and not exposed to the outside world. It seems I cannot include `dash/dart/base/atomic.h` in a test file.

devreal:
I created some additional unit tests to cover dash::memalloc and DASH algorithms. Please review and merge if appropriate.

fuchsto (approved):

fuchsto:
@devreal Splendid! Will merge as soon as CI is done.

fuchsto:
Looks like the clang build in CI does not support OpenMP:

<https://1203-45930061-gh.circle-artifacts.com/0/tmp/circle-junit.9IGEHUV/mpich/clang/Release/logs/build.log>

Build system correctly reports the missing thread support:

```
-- Enable multithreading:    (ENABLE_THREADSUPPORT)          OFF
```

(While I'm at it: this line should be placed above, at the other `ENABLE_` entries, and alignment of `OFF` is wrong)

This should be okay. The test case should be disabled if DASH / DART are built without OpenMP support, however.

Issue #249, opened 2017-01-24
User Request: Atomics in dash::UnorderedMap


fuchsto:
User request I received by mail:

> I have come across Dash and found it very interesting. This is something that could allow me, for once, to avoid having to redo everything from scratch with MPI.
> 
> I am trying to use an UnorderedMap for my purposes but I am stumbling into difficulties because one works on a local copy that has to be synchronised later. I am not in the usual situation of a halo, for example, where each node works on specific cells that will be updated at a specific time. In my case, any node could be able to create or increment any element of the map. Even worse, several nodes could try to create an element with the same key at the same time. In that case, the logic would be to simply add their values.
> 
> This is for a genomics application where every node is processing part of a genome and we count specific strips of DNA. Each of these strip will correspond to the key of the map and we simply have to count the number of times they appear. There is a huge number of possibilities (up to 10^12 or even more) of those strips but not all of them will be found. Any node could find any strips. So two nodes could find the same strip and try to insert that element in the map, if it was not already there, or to increment it and that operation should be atomic. 
> 
> It is for that reason that the local copy is causing me difficulties. Of course, I do not know well your library. I have just been experimenting with it so I might have missed an obvious way to do it. I would be grateful if you could give me a suggestion or simply tell me that this is a bad idea. I am not asking for technical help but a mere suggestion.
> 
> Thank you!

fuchsto:
My reply:

First, thank you for evaluating DASH for your application.
We highly appreciate feedback from users and will gladly provide technical support.

The "vanilla" configuration of dash::UnorderedMap is designed to provide a distributed key/value store.
If I understand your use case correctly, you use mapped values for shared counters.
So, for a simplified example:

```c++
dash::Array<GATCValue> dna;
dash::UnorderedMap<GATCSequence, int> map;
// ...
// find match in local data segment:
auto matched_strip_begin = std::find(dna.local.begin(),
                                     dna.local.end(),
                                     gatc_match_fun);
// ...
auto matched_dna_strip = GATCSequence(*matched_strip_begin, ...);
// increment global hit count of matched strip:
map[matched_dna_strip]++; // <-- not atomic
```

(Let me know if I got your use case wrong.)

In DASH, we try to mimic thread-safety and consistency semantics of the STL as closely as possible.
The consistency violation you observed is caused by non-atomic changes to the mapped values. Much like in STL containers, changes to *values* of container elements do not affect container semantics
(opposed to adding/removing container elements).
[...]

PR #255, opened 2017-01-26
Feat 249 atomic

fmoessbauer:
This PR introduces a clean atomic concept in DASH. It provides an easy to use interface for atomic operations similar to `std::atomic<T>` for elements stored in DASH containers. The STL atomic concept cannot be ported 1:1 as remote accesses require copying (amongst others).

**The interface is as follows:**

- Phantom type `dash::Atomic<T>` which can (only) be placed in DASH containers.
- Specialisation of `dash::GlobRef<Atomic<T>>` which provides interface of `std::Atomic` and proxies it to atomic DART calls. Methods which cannot be performed atomically are disabled / deleted.
- Function-based interface `dash::atomic::` (works only for atomic types)
- DART interface for atomic operations which internally uses the corresponding MPI operations.
- Static assertions to check weather atomic operations on the underlying type are supported

**Currently neither implemented nor specified:**

- memory model information
- memory model traits
- handling of local accesses e.g. `array.lbegin().load()`

**Mixing atomic and non-atomic accesses**

As the constructor of `dash::GlobRef<T>` takes a DART gptr which is not strongly-typed, atomic accesses to non-atomic elements (and vice versa) are possible. In some cases this might be beneficial, as atomic operations are expensive compared to non-atomic ones.
However between sections with non-atomic and atomic operations there has to be a **barrier and a memory flush**. Otherwise the behavior is undefined. Furthermore we do not recommend using this "hack" at all.

An example of this is `dash::Shared` which provides both non-atomic (`shared.get()`) and atomic (`shared.atomic.get()`) access.

**Container Example:**
```c++
dash::Array<dash::Atomic<int>> array(100);
// supported as Atomic<value_t>(value_t T) is available
dash::fill(array.begin(), array.end(), 0);

if(dash::myid() == 0){
  array[10].store(5);
  }
dash::barrier();
array[10].add(1);
// postcondition:
// array[10] == dash::size() + 5
```

The following does not work:
```c++
dash::Array<dash::Atomic<int>> array(100);
array.local[10].load();               // not allowed
dash::atomic::load(array.local[10]);  // not allowed
dash::atomic::load(array.lbegin());   // not allowed
```


fuchsto (changes requested):
@dash-project/developers Do not use wrapper functions for atomic operations that hide the atomic type!

See my comment here:

https://github.com/dash-project/dash/issues/249#issuecomment-275514703

fmoessbauer:
As discussed yesterday, I finally managed to change the interface in a way that follows as close as possible the C++11 standard as well as `std::atomic`.

This introduces the following changes:
- old `dash::Atomic` becomes `dash::AtomicAddress`
- new version of `dash::Atomic` is phantom type
- provide algorithms to change `GlobRef<Atomic<>>`
- specialize `GlobRef<Atomic>` to provide atomic interface
- name atomic interface according to std::atomic

Please review it very carefully.

dash/include/dash/Atomic.h
```diff
@@ -34,58 +39,56 @@ class Atomic
...
+  void set(ValueType value)
   {
-    DASH_LOG_DEBUG_VAR("Atomic.set()", val);
-    DASH_LOG_TRACE_VAR("Atomic.set",   _gptr);
-    *(reference(_gptr)) = val;
-    DASH_LOG_DEBUG("Atomic.set >");
+    DASH_LOG_DEBUG_VAR("AtomicAddress.set()", value);
+    DASH_LOG_TRACE_VAR("AtomicAddress.set",   _gptr);
+    *(pointer(_gptr)) = value;
+    dart_flush_all(_gptr);
```
devreal: Two things: 

1. Is the flush really required? The assignment should use `dart_put_blocking`, which issues a flush at the end. 
2. More importantly: This operation is not atomic and might thus conflict with other atomic operations executed on this memory. I suggest using `dart_accumulate`, which is guaranteed to be atomic (implemented in terms of `MPI_Accumulate`.

dash/include/dash/Atomic.h
```diff
@@ -34,58 +39,56 @@ class Atomic
...
    * Get a reference on the shared atomic value.
    */
   reference get()
   {
-    DASH_LOG_DEBUG("Atomic.get()");
-    DASH_LOG_TRACE_VAR("Atomic.get", _gptr);
+    DASH_LOG_DEBUG("AtomicAddress.get()");
+    DASH_LOG_TRACE_VAR("AtomicAddress.get", _gptr);
     DASH_ASSERT(!DART_GPTR_ISNULL(_gptr));
     reference ref(_gptr);
```
devreal: See my comment on `set()`. Better use `dart_fetch_and_op` with `DART_OP_NONE`. Alternatively, we might consider implementing a counterpart of `MPI_Get_accumulate`, i.e., `dart_get_accumulate`.

dash/include/dash/Atomic.h
```diff
@@ -202,19 +204,288 @@ class Atomic
...
+   * 
+   * \return  True if value is exchanged
+   */
+  bool compare_exchange(
+    /// value expected to be found in the atomic object
+    ValueType expected,
+    /// value to store in the atomic object if it is as expected
+    ValueType desired)
+  {
+    // TODO;
```
devreal: I guess this requires something like `dart_compare_and_swap()`, will add an implementation of it today.

dash/include/dash/GlobRef.h
```diff
@@ -335,24 +338,260 @@ class GlobRef
     size_t offs = (size_t) &( reinterpret_cast<P*>(0)->*mem);
     return member<MEMTYPE>(offs);
   }
-
+
```
devreal: grrr, why all those trailing whitespaces?

dash/include/dash/GlobRef.h
```diff
@@ -335,24 +338,260 @@ class GlobRef
...
+    return fetch_sub(value) - value;
+  }
+  
+private:
+  
+  dart_gptr_t _gptr;
+  
+};
+
+
```
devreal: I suggest moving this specialization to `Atomic.h` as non-atomic use does not require it to compile successfully.

devreal (changes requested):
Great work! Some suggestions for improvements inline. I will take care of implementing `dart_compare_and_swap` today.

fmoessbauer:
@devreal Thanks for your comments. I did some major restructuring of the atomic types / classes and algorithms. Now the interface is clean and can be loaded by including `dash/Atomic.h`.

And now to the technical part: I changed the implementation according to your comments, but unfortunately now the atomics do not work anymore. For me it seems that the dart versions of MPI accumulate functions lose some of the atomicity guarantees. So please have a look at the changes. In the meantime we could use commit 347cbdb which works perfectly.

fuchsto:
@fmoessbauer: Good, please wait for my review before further implementation.

devreal:
@fmoessbauer Thanks for restructuring the code, much better now. You mean that it does not work because of the Segfault? I'm not sure whether this is related to `dart_fetch_and_op`, as the segfault happens in in a flush:

```
#0  0x00007ffff5d64428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff5d6602a in __GI_abort () at abort.c:89
#2  0x00007ffff5da67ea in __libc_message (do_abort=do_abort@entry=1, fmt=fmt@entry=0x7ffff5ebd8a2 "*** %s ***: %s terminated\n") at ../sysdeps/posix/libc_fatal.c:175
#3  0x00007ffff5e4756c in __GI___fortify_fail (msg=<optimized out>, msg@entry=0x7ffff5ebd884 "stack smashing detected") at fortify_fail.c:37
#4  0x00007ffff5e47510 in __stack_chk_fail () at stack_chk_fail.c:28
#5  0x00007ffff08721ac in ?? () from /usr/lib/openmpi/lib/openmpi/mca_btl_vader.so
#6  0x00007ffff50091ea in opal_progress () from /usr/lib/libopen-pal.so.13
#7  0x00007fffeaf70895 in ?? () from /usr/lib/openmpi/lib/openmpi/mca_osc_pt2pt.so
#8  0x00007fffeaf71667 in ompi_osc_pt2pt_flush_all () from /usr/lib/openmpi/lib/openmpi/mca_osc_pt2pt.so
#9  0x00007ffff6d7cf04 in PMPI_Win_flush_all () from /usr/lib/libmpi.so.12
#10 0x000000000071fca8 in dart_flush_all (gptr=...) at /home/joseph/src/dash/dash/dart-impl/mpi/src/dart_communication.c:1108
#11 0x0000000000606bad in dash::GlobMem<dash::Atomic<int>, dash::allocator::CollectiveAllocator<dash::Atomic<int> > >::flush_all (this=<optimized out>)
    at /home/joseph/src/dash/dash/dash/include/dash/GlobMem.h:420
#12 dash::Array<dash::Atomic<int>, long, dash::BlockPattern<1, (dash::MemArrange)1, long> >::barrier (this=0x7fffffffd160, this=0x7fffffffd160)
    at /home/joseph/src/dash/dash/dash/include/dash/Array.h:1082
#13 dash::Array<dash::Atomic<int>, long, dash::BlockPattern<1, (dash::MemArrange)1, long> >::deallocate (this=this@entry=0x7fffffffd160)
    at /home/joseph/src/dash/dash/dash/include/dash/Array.h:1174
#14 0x0000000000600dae in dash::Array<dash::Atomic<int>, long, dash::BlockPattern<1, (dash::MemArrange)1, long> >::~Array (this=0x7fffffffd160, 
    __in_chrg=<optimized out>) at /home/joseph/src/dash/dash/dash/include/dash/Array.h:819
#15 AtomicTest_AtomicInterface_Test::TestBody (this=<optimized out>) at /home/joseph/src/dash/dash/dash/test/AtomicTest.cc:205
#16 0x00000000006e80eb in void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*)
    ()
#17 0x00000000006e16ef in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) ()
#18 0x00000000006c4e2c in testing::Test::Run() ()
#19 0x00000000006c57c4 in testing::TestInfo::Run() ()
#20 0x00000000006c5eb7 in testing::TestCase::Run() ()
#21 0x00000000006ccffe in testing::internal::UnitTestImpl::RunAllTests() ()
#22 0x00000000006e96c3 in bool testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) ()
#23 0x00000000006e2531 in bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) ()
#24 0x00000000006cba9a in testing::UnitTest::Run() ()
#25 0x00000000004d8a92 in RUN_ALL_TESTS () at /home/joseph/src/dash/dash/build/gtest/src/GTestExternal/googletest/include/gtest/gtest.h:2233
#26 main (argc=1, argv=0x7fffffffd938) at /home/joseph/src/dash/dash/dash/test/main.cc:70
```

Is that the Segfault you see as well? This error appeared using OpenMPI 1.10.2. I will test other versions and MPICH as well.

devreal:
It seems that the `fetch_and_op` operation corrupts the stack in some way in the test suite. However, I am not able to reproduce this with a small example so it's painful to debug. I'll keep digging through it...

fmoessbauer:
@devreal Thanks for putting time into this. Yesterday I talked to @rkowalewski and @fuchsto about the flushes. We need them, as we have to ensure remote completion. However I guess they are not necessary for pure read operations?

devreal:
That's what I realized today, I was assuming that atomic operations are performed on the spot. I guess I was too puzzled by the random errors the missing flushes caused, all MPI implementations are terrible in these cases. Inside `get()`, a call to `dart_flush_local` should be sufficient. Not sure whether that makes any difference though.

fuchsto (approved):

dash/include/dash/atomic/AtomicAddress.h
```diff
@@ -0,0 +1,270 @@
...
+#include <dash/internal/Logging.h>
+
+namespace dash {
+
+// forward decls
+template<typename T>
+class GlobRef;
+
+template<typename ValueType>
+class AtomicAddress
```
fuchsto: For consistency, this class template should be named `GlobAtomicRef`. It acts as extension of `GlobRef` by atomic operations, like `GlobAsyncRef` does for async operations.
Leading to my second point: This type should satisfy (and extend) the `GlobRef` concept. I think it already does, but it should be checked and documented.

The name `AtomicAddress` is misleading / incorrect in any case:
An address is not atomic, it's just a (compound) number.
That also is to say: the template class does not provide operations on an **address**, and this would be nonsensical. So let's not use this term here.

Let me illustrate in simplified pseudo code:

```c++
Address a(something);
Address b = a.sub(3); // <-- looks like pointer arithmetic?
```

fuchsto (changes requested):
The memory model traits are still missing, but I wouldn't care for now. The changes definitely are an improvement compared to the current implementation in `development`.

The memory model concept will be addressed in a separate issue.

dash/include/dash/atomic/GlobAtomicRef.h
```diff
@@ -0,0 +1,367 @@
...
+
+/**
+ * Specialization for atomic values. All atomic operations are 
+ * \c const as the \c GlobRef does not own the atomic values.
+ */
+template<typename T>
+class GlobRef<dash::Atomic<T>>
+{
+  static_assert(dash::dart_datatype<T>::value != DART_TYPE_UNDEFINED,
+         "dash::AtomicAddress only valid on integral and floating point types");
```
fuchsto: Should be `dash::GlobRef<dash::Atomic<T>>` in static assertion message

dash/include/dash/atomic/GlobAtomicRef.h
```diff
@@ -0,0 +1,367 @@
...
+    store(value);
+    return *this;
+  }
+  
+  /**
+   * Set the value of the shared atomic variable.
+   */
+  void store(const T & value) const
+  {
+    DASH_LOG_DEBUG_VAR("AtomicAddress.set()", value);
```
fuchsto: Log messages have scope "AtomicAddress"

fuchsto (changes requested):

fuchsto (approved):

fuchsto:
@fmoessbauer Are all comments by @devreal addressed in your recent changes?

devreal (approved):

PR #261, opened 2017-02-03
Fixes the move semantic of LocalAllocator and CollectiveAllocator

### LINKED ISSUE #258: const moves assignment operators
In the code are at least two move assignment operators with a constant variable.
[self_t & operator=(const self_t && other) noexcept](https://github.com/dash-project/dash/blob/development/dash/include/dash/allocator/LocalAllocator.h#L124) and 
[self_t & operator=(const self_t && other) noexcept](https://github.com/dash-project/dash/blob/development/dash/include/dash/allocator/CollectiveAllocator.h#L126)

I think the const before the member name is wrong.

fmoessbauer:
Fixes #258

### DIFF
--- a/dash/include/dash/allocator/CollectiveAllocator.h
+++ b/dash/include/dash/allocator/CollectiveAllocator.h
@@ -210,6 +210,30 @@ public:
    */
   void deallocate(pointer gptr)
   {
+    _deallocate(gptr, false);
+  }
+
+private:
+  /**
+   * Frees all global memory regions allocated by this allocator instance.
+   */
+  void clear() noexcept
+  {
+    for (auto gptr : _allocated) {
+      _deallocate(gptr, true);
+    }
+    _allocated.clear();
+  }
+  /**
+   * Deallocates memory in global memory space previously allocated in the
+   * active unit's local memory.
+   */
+  void _deallocate(
+    /// gptr to be deallocated
+    pointer gptr,
+    /// if true, only free memory but keep gptr in vector
+    bool    keep_reference = false)
+  {
     if (!dash::is_initialized()) {
       // If a DASH container is deleted after dash::finalize(), global
       // memory has already been freed by dart_exit() and must not be
@@ -228,24 +252,12 @@ public:
       dart_team_memfree(_team_id, gptr),
       DART_OK);
     DASH_LOG_DEBUG("CollectiveAllocator.deallocate", "_allocated.erase");
-    _allocated.erase(
+    if(!keep_reference){
+      _allocated.erase(
         std::remove(_allocated.begin(), _allocated.end(), gptr),
         _allocated.end());
-    DASH_LOG_DEBUG("CollectiveAllocator.deallocate >");
-  }
-
-private:
-  /**
-   * Frees all global memory regions allocated by this allocator instance.
-   */
-  void clear() noexcept
-  {
-    for (auto gptr : _allocated) {
-      // TODO:
-      // Inefficient as deallocate() applies vector.erase(std::remove)
-      // for every element.
-      deallocate(gptr);
     }
+    DASH_LOG_DEBUG("CollectiveAllocator.deallocate >");
   }
 
 private:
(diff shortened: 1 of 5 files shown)


dash/include/dash/allocator/CollectiveAllocator.h
```diff
@@ -123,12 +126,15 @@ class CollectiveAllocator
   /**
    * Move-assignment operator.
    */
-  self_t & operator=(const self_t && other) noexcept
+  self_t & operator=(self_t && other) noexcept
   {
     // Take ownership of other instance's allocation vector:
     if (this != &other) {
-      std::swap(_allocated, other._allocated);
```
devreal, line 130: The use of `std::swap` should be OK here because `other` is slated for deallocation after a move, thus deallocating the current `_allocated`. Also, the call to `other._allocated.clear()` should not be required afaics.

devreal (changes requested):
I think the calls to `other._allocated.clear()` are not required since the vector is empty after `std::move(other._allocated)`. Also see my comment on the use of `std::swap` inline.

dash/include/dash/allocator/CollectiveAllocator.h
fmoessbauer: Are you really sure about that: If `_allocated` is not empty, the elements are moved to `other._allocated`. This is especially problematic, as the deallocator of `other` will free them later. The call to `other._allocated.clear()` is necessary, as the move assignment operator of `std::vector` does not ensure that the elements are freed. We have to ensure that however to avoid deallocation in the destructor. My statement is based on the documentation on [cppreference / `std::vector`](http://en.cppreference.com/w/cpp/container/vector/operator%3D). Maybe I missed something.

devreal: > This is especially problematic, as the deallocator of other will free them later.

That was my point. They will be free'd once `other` is destructed. Is that problematic? Since `other` is an `rvalue`, it's lifetime is limited and the destruction will happen soon anyway (most likely after the call). If it is an `xvalue`, `other` is in a _valid but unspecified_ state afterwards, which is OK with the semantics of move. 

After all, this is just a shortcut and I am fine with keeping it the way it's done in you changes.

fmoessbauer: Ok, got your point: We have to differentiate between destruction of the allocator, deallocation of the vector and deallocation of the `dart_gptr`. I implemented it in this way to be absolutely sure that the vector is empty as otherwise in [`clear()`](https://github.com/dash-project/dash/blob/20fdf5d3d5799681c145504f589f611673fa6d82/dash/include/dash/allocator/LocalAllocator.h#L226) each `dart_gptr` is freed. This function is called in the deallocator of `LocalAllocator` and hence would invalidate the `dart_gptr` of the new `LocalAllocator`.

devreal: > This function is called in the deallocator of LocalAllocator and hence would invalidate the dart_gptr of the new LocalAllocator.

But after the assignment, the `rvalue` owns the `gptr` that were previously owned by _our_ instance of the allocator, which has no reference to them anymore after the swap. So it is OK to free them in the destructor of the `rvalue`. Anyway, your version makes the semantics more obvious :)

fmoessbauer:
I did a little performance enhancement in the deallocation phase and removed the range based for loop which modified the container. Please review again ;)

dash/test/CollectiveAllocatorTest.cc
```diff
@@ -12,3 +14,35 @@ TEST_F(CollectiveAllocatorTest, Constructor)
...
+
+    if(dash::myid().id == 0){
+      // assign value
+      int value = 10;
+      (*gptr) = value;
+    }
+    dash::barrier();
+
+    target_new       = Alloc_t(); 
+    target_new       = std::move(target_old);
```
fuchsto, line 37: This tests move-assignment, good. But move-construction is not covered in tests, is it?

fmoessbauer: Now it is ;)

devreal (approved):

Issue #259, opened 2017-01-31
Call of a constant function within a non-constant function


Fettpet:
In the functions [const_mapped_type_reference at(const key_type & key) const](https://github.com/dash-project/dash/blob/development/dash/include/dash/map/UnorderedMap.h#L513)
 and
 [const_mapped_type_reference at(const key_type & key) const](https://github.com/dash-project/dash/blob/development/dash/include/dash/map/UnorderedMapLocalRef.h#L232)
the operator[] is called. The operator[] has no const overload.

fuchsto:
The actual defect is that `at(const key_type & key) const` is implemented using `operator[](key)`.

Just as for `std::unordered_map`, there is no const overload for `operator[](key)` as it would create an element with the specified key if it doesn't exist yet:

<http://en.cppreference.com/w/cpp/container/unordered_map/operator_at>

fuchsto:
Fixed in #262

PR #262, opened 2017-02-04
Fixed operator[] const in UnorderedMap* templates

fuchsto:
Addresses #259

### DIFF
--- a/dash/include/dash/map/UnorderedMap.h
+++ b/dash/include/dash/map/UnorderedMap.h
@@ -513,17 +513,42 @@ public:
   const_mapped_type_reference at(const key_type & key) const
   {
     DASH_LOG_TRACE("UnorderedMap.at() const", "key:", key);
-    // TODO: Unoptimized, currently calls find(key) twice as operator[](key)
-    //       calls insert(key).
-    const_iterator git_value = find(key);
-    if (git_value == _end) {
+    auto found = find(key);
+    if (found == _end) {
       // No equivalent key in map, throw:
       DASH_THROW(
         dash::exception::InvalidArgument,
         "No element in map for key " << key);
     }
-    auto mapped = this->operator[](key);
-    DASH_LOG_TRACE("UnorderedMap.at > const", mapped);
+    dart_gptr_t gptr_mapped   = iterator(this, found.pos()).dart_gptr();
+
+    value_type  * lptr_value  = static_cast<value_type *>(
+                                  found.local());
+    mapped_type * lptr_mapped = nullptr;
+    DASH_LOG_TRACE("UnorderedMap.at", "gptr to element:", gptr_mapped);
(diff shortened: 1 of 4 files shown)


dash/include/dash/map/UnorderedMap.h
```diff
@@ -513,17 +513,42 @@ class UnorderedMap
...
+    DASH_LOG_TRACE("UnorderedMap.at", "byte offset of mapped member:",
+                   mapped_offs);
+    // Increment pointers to element by byte offset of mapped value member:
+    if (lptr_value != nullptr) {
+      // Convert to char pointer for byte-wise increment:
+      char * b_lptr_mapped  = reinterpret_cast<char *>(lptr_value);
+      b_lptr_mapped        += mapped_offs;
+      // Convert to mapped type pointer:
+      lptr_mapped           = reinterpret_cast<mapped_type *>(b_lptr_mapped);
+    }
```
fmoessbauer: This looks way to complex to me. Why can't you just use `lptr_mapped = &(lptr_value->second)`?

fuchsto: If I remember correctly, this became necessary for some MPI runtimes.

The actual reasons behind this are that 
- `std::pair` cannot be used as MPI data type directly.
- Offset-to-member only works reliably with `offsetof` in the general case

I have to use `offsetof` in line 535 above as there is no instance of `value_type` available that could be used to calculate the member offset as `l_ptr_value` is possibly undefined.
Using `std::declval()` instead (to generate a compile-time pseudo-instance for member resolution) only works if `Key` and `Mapped` are default-constructible.

Finally, the distance obtained from `&(lptr_value->second) - lptr_value` had different alignment than the address obtained via `offsetof` in some cases, depending on the combination of MPI runtime and compiler. Apparently some compilers / standard libs have special treatment (padding?) for members of `std::pair` such that `__builtin_offsetof(type, member)` differs from the member-offset provided by the type system.

The alternative, using `offsetof` (which resolves to `__builtin_offsetof` automatically if needed) and manual pointer increments works, however.

devreal: Why is `&(lptr_value->second)` valid in cases where `value_type` is not of `standard_layout`?

fuchsto:
@fmoessbauer Bounce for review of my recent changes

fmoessbauer (approved):
Looks good! However the pointer arithmetics on `lptr_value` are still not clear to me. Let's discuss this tomorrow.

devreal:
Please thoroughly document the `lptr_value` quirk before merging.

fuchsto:
@devreal 
I moved the quirky bits to a separate private method and added documentation there.

fuchsto:
Aaah ... this is embarassing ... I forgot to `git add` my recent changes. Will create a new pull request, run CI, and merge it.

PR #264, opened 2017-02-04
many fixes for index calculations in BlockPattern with underfilled blocks

fmoessbauer:
This commit fixes the offsets and sizes of blocks in the `dash::BlockPattern` and it's specialization `dash::BlockPattern<1,...>` for underfilled blocks. It took me some time to haunt down the bugs because there were no (really, NO) unit tests on this functionality for underfilled cases.
The wrong pattern information were the cause of sporadic crashes in the HDF5 adapter for some pattern configurations.

So please @dash-project/developers : Write **unit tests** for every new function / feature / procedure / whatever. This will make life so much easier.


fuchsto (approved):
Thank you again for your effort.
Note to all @dash-project/developers :
Patterns will be gradually redesigned as expression templates in the near future (see feat-191-lranges for first advancements in this), so don't invest too much time in the current pattern types.

PR #270, opened 2017-02-09
Documented quirks for mapped type address resolution in UnorderedMap

fuchsto:
PR for missing changes from #262


fuchsto:
PR for missing changes from #262

PR #271, opened 2017-02-09
Improvements to TestBase

devreal:
1. Fix output of DASH size during test initialization/tear-down
2. Introduce generic EXPECT_EQ_U macro correctly comparing floating point values. This allows for correct checks even in places where EXPECT_EQ_U is called on templated types, e.g., in the HDF5 tests.

The latter causes GCC to issue a warning if `NULL` is passed as `expected`. I played around with the templates but couldn't find a simple way for fixing it. The warning is issued once during test-suite compilation:

```
In file included from /home/joseph/src/dash/dash/build/gtest/src/GTestExternal/googletest/include/gtest/gtest.h:1874:0,
                 from /home/joseph/src/dash/dash/dash/test/TestBase.h:6,
                 from /home/joseph/src/dash/dash/dash/test/DARTLocalityTest.h:4,
                 from /home/joseph/src/dash/dash/dash/test/DARTLocalityTest.cc:2:
/home/joseph/src/dash/dash/dash/test/DARTLocalityTest.cc: In member function ‘virtual void DARTLocalityTest_ExcludeLocalityDomain_Test::TestBody()’:
/home/joseph/src/dash/dash/build/gtest/src/GTestExternal/googletest/include/gtest/gtest_pred_impl.h:147:45: warning: passing NULL to non-pointer argument 3 of ‘typename std::enable_if<(! std::is_floating_point<_Tp>::value), testing::AssertionResult>::type testing::internal::assert_float_eq(const char*, const char*, const T&, const S&) [with T = long int; S = dart_domain_locality_s*; typename std::enable_if<(! std::is_floating_point<_Tp>::value), testing::AssertionResult>::type = testing::AssertionResult]’ [-Wconversion-null]
   GTEST_ASSERT_(pred_format(#v1, #v2, v1, v2), \
```

Note that the code that causes the warning is never executed.

### DIFF
--- a/dash/test/MatrixTest.cc
+++ b/dash/test/MatrixTest.cc
@@ -893,7 +893,7 @@ TEST_F(MatrixTest, DelayedAlloc)
                          "phase:",       phase_coords, "=", phase,
                          "expected:",    expected,
                          "actual:",      actual);
-          EXPECT_DOUBLE_EQ_U(expected, actual);
+          EXPECT_EQ_U(expected, actual);
         }
       }
     }
--- a/dash/test/TestBase.h
+++ b/dash/test/TestBase.h
@@ -1,6 +1,8 @@
 #ifndef DASH__TEST__TEST_BASE_H_
 #define DASH__TEST__TEST_BASE_H_
 
+#include <type_traits>
+
 #include <gtest/gtest.h>
 
 #include <dash/Init.h>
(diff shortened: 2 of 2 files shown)


fuchsto (approved):
Nice!
There are some additional improvements in `feat-191-lranges` that fit nicely to this and further simplify `TestBase`.

Issue #274, opened 2017-02-13
Passing lambdas to algorithm interfaces (for_each, generate, min_element, ...)


devreal:
I noticed that the signature of `dash::for_each` requires a `std::function<void(ElementType &)>` to be passed in as functor. In particular, it seems that this excludes passing lambdas instead of pre-defined function arguments. The following fails to compile with the error that the lambda is not derived from `std::function`:

```
dash::Array<int> arr(dash::size());
dash::for_each(arr.begin(), arr.end(), [](int& val){ val = dash::size(); })
```

Is there any reason for explicitly excluding lambdas here? On my system, the function type is part of the template specification:

```
  template<typename _InputIterator, typename _Function>
    _Function
    for_each(_InputIterator __first, _InputIterator __last, _Function __f) { ... }
```

rkowalewski:
I agree with this limitation. Moreover, lambdas are more efficient in many cases compared to `std::function` as they can be better optimized by the compiler.

fuchsto:
I agree, there was no specific reason against using a template type like `UnaryFunction`, I just didn't think of it back then.

fuchsto:
Fixed in #275

PR #275, opened 2017-02-13
Generalized interfaces of algorithm interfaces for lambdas

fuchsto:
Addresses #274


devreal (changes requested):
Please document the expected signature of the function. Also, members like `dash::generate` should be adopted as well (not sure where else support for lambdas is missing)

devreal:
I noticed that `dash::generate` and `dash::[min|max]_element` are affected as well.

fuchsto:
@devreal Good catch, will extend the PR to the affected algorithm interfaces.

fuchsto:
Added documentation of template parameters (mostly function signatures), generalized function parameters to deduced template types.
I ran a `grep -R 'std::function' dash/include/dash/algorithm/*` which now comes back clean.

@devreal Should cover your comments.

dash/include/dash/algorithm/ForEach.h
```diff
@@ -96,25 +58,27 @@ void for_each(
  * function on its local elements only. The index passed to the function is
  * a global index.
  *
- * \tparam      ElementType  Type of the elements in the sequence
- * \tparam      IndexType    Parameter type expected by function to
- *                           invoke, deduced from parameter \c func
+ * \tparam      ElementType            Type of the elements in the sequence
+ * \tparam      UnaryFunctionWithIndex Function to invoke for each element
+ *                                     in the specified range with signature
+ *                                     \c void (const ElementType &, index_t)
```
devreal, line 68: It does not have to be const, right?

dash/include/dash/algorithm/MinMax.h
```diff
@@ -137,24 +140,26 @@ const ElementType * min_element(
...
  * \complexity  O(d) + O(nl), with \c d dimensions in the global iterators'
  *              pattern and \c nl local elements within the global range
  *
  * \ingroup     DashAlgorithms
  */
 template<
-  typename ElementType,
-  class    PatternType>
+  class ElementType,
+  class Compare = std::less<const ElementType &>,
```
devreal: Put `compare` (with default value) to the end of the the template parameter list?

dash/include/dash/algorithm/ForEach.h
```diff
@@ -14,65 +14,27 @@ namespace dash {
...
+  class    PatternType,
+  class    UnaryFunction >
 void for_each(
   /// Iterator to the initial position in the sequence
   const GlobIter<ElementType, PatternType> & first,
   /// Iterator to the final position in the sequence
   const GlobIter<ElementType, PatternType> & last,
   /// Function to invoke on every index in the range
-  ::std::function<void(ElementType &)> & func)
+  UnaryFunction                              func)
```
devreal, line 41: `UnaryFunction` can be passed as `const &` (saves same cycles with `std::function`)

devreal (dismissed):
Looks good, a few minor comments inline.

dash/include/dash/algorithm/ForEach.h
fuchsto: Does not "have to", but it should:

<http://en.cppreference.com/w/cpp/algorithm/for_each>

The qualifier merely states that accepting const references to values is supported.

dash/include/dash/algorithm/ForEach.h
fuchsto: Oh, no:
If `const Func &` is passed, the type `UnaryFunction` would resolve to `const Func &`.
When an lvalue function object is passed, the compiler will deduce `const T &`.
With `const UnaryFunction &`, non-const parameters would be explicitly forbidden.

Also see the signature of `std::for_each`:

<http://en.cppreference.com/w/cpp/algorithm/for_each>

dash/include/dash/algorithm/MinMax.h
fuchsto: Definitely, fixed in my upcoming commit.

fuchsto:
@devreal Updated documentation and fixed tparam order.

dash/include/dash/algorithm/ForEach.h
fuchsto: Documentation has been extended and now also explicitly refers to `std::for_each`.

Issue #283, opened 2017-02-15
NArray/Matrix as const throws compiler error


BenjaProg:
Hey,
in my simplified example I want to call a function with a const refernce to a NArray/Matrix.
But then I get an compiler error.
~~~shell
main.cpp:27:30: error: passing ‘const local_type {aka const dash::LocalMatrixRef<unsigned int, 2, 2, dash::BlockPattern<2, (dash::MemArrange)1u, long int> >}’ as ‘this’ argument discards qualifiers [-fpermissive]
~~~
I made a very reduced example in [feat-283-constMatrix](https://github.com/dash-project/dash/blob/feat-283-constMatrix/dash/examples/ex.12.feat-283/main.cpp).
If the parameter in [line12](https://github.com/dash-project/dash/blob/feat-283-constMatrix/dash/examples/ex.12.feat-283/main.cpp#L12) is const then the compiler error arises.

This example is without really meaning, but to show the problem. 
The real example would be [cowichan_product](https://github.com/dash-project/dash-apps/blob/master/cowichan/product/main.cpp).
Everything is fine as long as NArray/Matrix is not a const parameter.

BenjaProg:
I think that's just a missing feature in form of missing const versions of some function(s)?

fmoessbauer:
I will have a look at this tonight. If you want to assist, please create an unit test which is not compilable (but should be). Then it's much easier to track down the problem and this also enhances the code coverage.

BenjaProg:
@fmoessbauer you solved my example already, i just tested compiling the cowichan example and it worked fine 👍 in my branch feat-283-constMatrix is the example which should not compile without your changes.
I can write a test for matrix access but I can't do it in the next days, because I have to put more time into my bachelor thesis ;-)

fmoessbauer:
I tested it with your example, so it should work. You can pull from the bug branch if necessary. Unfortunately there are still some things to discuss before we can merge the PR. And there are some problems regarding the CI API of github at the moment.

fuchsto:
Fixed in PR #291

BenjaProg:
With the latest changes my examples from above don't comile with const parameter anymore.
Trying to compile the example from the feat branch leads to following error.
[constMatrix compiler Error.txt](https://github.com/dash-project/dash/files/787597/constMatrix.compiler.Error.txt)
An Extract:

~~~c++
[...]

PR #284, opened 2017-02-16
fixed const access at LocalMatrixRef by making sub, row, col, rows, cols const

fmoessbauer:
Fixes use cases like the following (issue #283):

```c++
void foo(const dash::NArray<uint, 2> & matIn){
  // ...
  auto mPtr = matIn.local.row(0).lbegin();
  // ...
}
```

### DIFF
--- a/dash/include/dash/matrix/LocalMatrixRef.h
+++ b/dash/include/dash/matrix/LocalMatrixRef.h
@@ -246,17 +246,17 @@ public:
     operator[](index_type n) const;
 
   template<dim_t NumSubDimensions>
-  LocalMatrixRef<T, NumDimensions, NumDimensions-1, PatternT>
-    sub(size_type n);
-  inline LocalMatrixRef<T, NumDimensions, NumDimensions-1, PatternT>
-    col(size_type n);
-  inline LocalMatrixRef<T, NumDimensions, NumDimensions-1, PatternT>
-    row(size_type n);
+  const LocalMatrixRef<T, NumDimensions, NumDimensions-1, PatternT>
+    sub(size_type n) const;
+  inline const LocalMatrixRef<T, NumDimensions, NumDimensions-1, PatternT>
+    col(size_type n) const;
+  inline const LocalMatrixRef<T, NumDimensions, NumDimensions-1, PatternT>
+    row(size_type n) const;
 
   template<dim_t SubDimension>
-  LocalMatrixRef<T, NumDimensions, NumDimensions, PatternT> sub(
+  const LocalMatrixRef<T, NumDimensions, NumDimensions, PatternT> sub(
     size_type n,
-    size_type range);
+    size_type range) const;
 
   /**
    * Create a view representing the matrix slice within a row
@@ -267,11 +267,11 @@ public:
    *
    * \see  sub
    */
-  inline LocalMatrixRef<T, NumDimensions, NumDimensions, PatternT> rows(
+  inline const LocalMatrixRef<T, NumDimensions, NumDimensions, PatternT> rows(
     /// Offset of first row in range
     size_type offset,
     /// Number of rows in the range
-    size_type range);
+    size_type range) const;
 
   /**
    * Create a view representing the matrix slice within a column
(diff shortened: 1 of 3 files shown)


dash/include/dash/matrix/internal/LocalMatrixRef-inl.h
```diff
@@ -445,28 +445,28 @@ ::sub(
 
 template<typename T, dim_t NumDim, dim_t CUR, class PatternT>
 inline LocalMatrixRef<T, NumDim, NumDim-1, PatternT>
-LocalMatrixRef<T, NumDim, CUR, PatternT>
+const LocalMatrixRef<T, NumDim, CUR, PatternT>
```
devreal: Mhh, the position of the first `const` seems strange to me. Shouldn't a `const` function return a `const` reference? (aka, we need both `const` and non-`const` versions of this function?) 

Also, instead of `inline` maybe these functions can be marked `constexpr`?

fmoessbauer: We cannot return references, as the functions return temporaries. I made them constexpr whenever possible in the next commit.

fmoessbauer:
I discovered a strange behavior while checking if we need non-const and const versions of the methods:

```c++
  dash::Matrix<int, 2, index_t, pattern_t> matrix(dash::SizeSpec<2>(8, 15));
  
  auto const & matrix_by_ref = matrix;
  auto & matrix_local  = matrix.local;
  
  // works
  el = *(matrix.local.lbegin());
  el = ++(*(matrix.local.lbegin()));
  el = ++(*(matrix.local.row(0).lbegin()));
  
  // test access using const & matrix
  el = matrix_by_ref[0][0];
  el = matrix_by_ref.local[0][0];
  //el = ++(*(matrix_by_ref.local.lbegin()));   // does not compile
  el = ++(*(matrix_by_ref.local.row(0).lbegin()));
  matrix_by_ref.local.row(0)[0] = 5;

  // test access using non-const & matrix.local
  *(matrix_local.lbegin()) = 5;
```

fmoessbauer:
Now we have both non-const and const versions. We definitely need them. The strange behavior is also fixed. Some wild `constexpr` fooled me.

fuchsto:
I agree with this request.
In my n-dim views branch, I addressed const-correctness of GlobIter, GlobPtr, GlobRef ... in a more fundamental context:

https://github.com/dash-project/dash/pull/288

Will review for overlap and report back.

fuchsto:
Please review against PR #291

dash/include/dash/matrix/LocalMatrixRef.h
```diff
@@ -242,22 +242,37 @@ class LocalMatrixRef
...
+  constexpr const LocalMatrixRef<T, NumDimensions, CUR-1, PatternT>
     operator[](index_type n) const;
 
-  template<dim_t NumSubDimensions>
   LocalMatrixRef<T, NumDimensions, NumDimensions-1, PatternT>
-    sub(size_type n);
-  inline LocalMatrixRef<T, NumDimensions, NumDimensions-1, PatternT>
     col(size_type n);
-  inline LocalMatrixRef<T, NumDimensions, NumDimensions-1, PatternT>
+  constexpr const LocalMatrixRef<T, NumDimensions, NumDimensions-1, PatternT>
```
fuchsto, line 250: The use of const types is incorrect as described in #290 
Should be
```c++
constexpr LocalMatrixRef<const T, ...>
```

Consider that `LocalMatrixRef` is a reference proxy. The definition of `value_type` of a const-reference should be `const T`. With the definition here, it's `T`.

fuchsto (changes requested):

fuchsto:
Changes in this pull request have been merged into #291 and fixed with correct definitions of const reference/proxy types.

The positive effects are visible e.g. in examples/ex.11.halo-stencil/main.cpp:

Previously (also with changes in this PR), this compiled but should lead to compile error (const-ness of parameter `data`):

```c++
void write_pgm(const std::string & filename, const Array_t & data) {
  // ...
  auto & first = data.begin();
  dash::copy(first+ext_x*y, first+ext_x*(y+1), buffer.data());
  // ...
}
```
With changes in PR #291, This now leads to a compile error.
The following corrected implementation compiles without error:
```c++
void write_pgm(const std::string & filename, const Array_t & data) {
  // ...
  const auto & first = data.begin();
  dash::copy(first+ext_x*y, first+ext_x*(y+1), buffer.data());
  // ...
}
```

fmoessbauer:
Ok, so I close this PR and we implement the remaining stuff in #291

Issue #290, opened 2017-02-17
GlobPtr, GlobRef, GlobIter const-Correctness


fuchsto:
According to the STL, iterators must be implicitly convertible to const-iterators.
In the current implementation of `dash::Array` and other containers, dependent types like `iterator` and `reference` are defined like

```c++
typedef GlobIter<T>         iterator;
typedef const GlobIter<T>   const_iterator;
typedef GlobRef<T>          reference;
typedef const GlobRef<T>    const_reference;
```

This, of course, is nonsense. It should rather be something like:

```c++
typedef GlobIter<T> iterator;
typedef GlobIter<const T> const_iterator;
// ... etc.
```

The correction of these type definitions has far-reaching consequences.
I fixed the essential semantics and traits and will create a PR for discussion.

To illustrate my point:
Consider:

```c++
typedef    T *                 pointer;
typedef    const pointer const_pointer;
```

In this example, `const_pointer`  defines `T * const`, a const-pointer to non-const value,
not `const T *`, a pointer to const value. The latter, obviously, is the correct definition.

Likewise, the `const` variant of `GlobIter<T>` is not `const GlobIter<T>` (changing the position
of the iterator is not what a `const` iterator is about) but `GlobIter<const T>` (changing the referenced
value is forbidden - which makes sense).

The broken const-correctness is visible in  visible e.g. in examples/ex.11.halo-stencil/main.cpp:
The following code compiles but should lead to compile error (const-ness of parameter `data`):

```c++
void write_pgm(const std::string & filename, const Array_t & data) {
  // ...
  auto & first = data.begin(); // <-- should be:  const auto & first
  dash::copy(first+ext_x*y, first+ext_x*(y+1), buffer.data());
  // ...
}
```

fuchsto:
Pull request #291 contains the related changes and also covers changes from PR #284

PR #291, opened 2017-02-17
Fixed definitions of const iterator and reference types

fuchsto:
Addresses #290 and #283

### DIFF
--- a/dash/include/dash/Array.h
+++ b/dash/include/dash/Array.h
@@ -126,13 +126,13 @@ public:
 
   typedef IndexType                                     difference_type;
 
-  typedef T &                                                 reference;
+  typedef       T &                                           reference;
   typedef const T &                                     const_reference;
 
-  typedef T *                                                   pointer;
+  typedef       T *                                             pointer;
   typedef const T *                                       const_pointer;
 
-  typedef T *                                                  iterator;
+  typedef       T *                                            iterator;
   typedef const T *                                      const_iterator;
 
 public:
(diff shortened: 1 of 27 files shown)


fuchsto:
The positive effects of the changes in this PR are visible e.g. in examples/ex.11.halo-stencil/main.cpp:

Previously, this compiled but should lead to compile error (const-ness of parameter `data`):

```c++
void write_pgm(const std::string & filename, const Array_t & data) {
  // ...
  auto & first = data.begin();
  dash::copy(first+ext_x*y, first+ext_x*(y+1), buffer.data());
  // ...
}
```
This now leads to a compile error.
The following corrected implementation compiles without error:
```c++
void write_pgm(const std::string & filename, const Array_t & data) {
  // ...
  const auto & first = data.begin();
  dash::copy(first+ext_x*y, first+ext_x*(y+1), buffer.data());
  // ...
}
```

fuchsto:
Includes and fixes changes from PR #284 

@fmoessbauer :

- I appreciate the ECpp item 3 quirk, but conversion between `LocalMatrixRef<T>` and `LocalMatrixRef<const T>` does not work in this case. We should find a way to make it work, though.
- Please review copy / assignment of `dash::Atomic`

dash/include/dash/Array.h
```diff
@@ -313,7 +316,7 @@ class AsyncArrayRef
    *
    * TODO: Should return GlobAsyncPtr<...>(_array->begin())
    */
-  inline pointer begin() noexcept {
+  inline iterator begin() noexcept {
```
fmoessbauer: `constexpr`

fmoessbauer (approved):
Looks good so far. These changes are really important. I created also a [code-coverage report](https://felixmoessbauer.de/temp/coverage/) for this branch.

Some tests do not succeed in coverage mode, so these are disabled: `$GTEST_FILTER="-TeamLocalityTest*:ThreadsafetyTest*"`

PR #293, opened 2017-02-20
Enforce element types of DASH containers to be trivially copyable

### LINKED ISSUE #241: dash::Matrix doesn't accept struct as template datatype
Why does `dash::Matrix` and thus `dash::NArray` not accept struct as template datatype while `dash::Array` does?
For example:

~~~c++
using Point = struct{ MTRX_TYPE value;
                      uint row, col;};
dash::Array  < Point   > points( nelts );  // ok
dash::NArray < Point,1 > points( nelts );  // bad
dash::Matrix < Point,1 > points( nelts );  // bad
~~~

devreal:
Consistently enforce `std::is_trivially_copyable` on element types of containers (except when using the Cray compiler).
Introduce minimum required compiler version.
Makes `GlobPtr` trivially copyable after #221 has been resolved.
Introduce `dash::Pair` which closely resembles `std::pair` except for being trivially copyable. 
Fixes #241

### DIFF
--- a/dash/include/dash/Matrix.h
+++ b/dash/include/dash/Matrix.h
@@ -135,8 +135,16 @@ template<
   class    PatternT       = TilePattern<NumDimensions, ROW_MAJOR, IndexT> >
 class Matrix
 {
-  static_assert(std::is_trivial<ElementT>::value,
-    "Element type must be trivial copyable");
+  /**
+   * The Cray compiler (as of CCE8.5.6) does not support
+   * std::is_trivially_copyable.
+   *
+   * TODO: Remove the guard once this has been fixed by Cray.
+   */
+#ifndef __CRAYC
+  static_assert(std::is_trivially_copyable<ElementT>::value,
+    "Element type must be trivially copyable");
+#endif
   static_assert(std::is_same<IndexT, typename PatternT::index_type>::value,
     "Index type IndexT must be the same for Matrix and specified pattern");
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Features:
 - Better textual output of unit tests
 - Added support for HDF5 groups
 - Support patterns with underfilled blocks in dash::io::hdf5
+- Allow trivially copyable data types in containers
 
 Bugfixes:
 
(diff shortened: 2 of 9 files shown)


fuchsto:
My Clangs also have issues with `is_trivially_copyable`. Perhaps we should introduce our own type traits for portability.

dash/include/dash/Matrix.h
```diff
@@ -135,13 +135,16 @@ template<
   class    PatternT       = TilePattern<NumDimensions, ROW_MAJOR, IndexT> >
 class Matrix
 {
-#if !defined(_CRAYC) && !defined(__INTEL_COMPILER)
```
fuchsto, line 138: Intel compilers (tested 16.0 and 17.0) don't allow `std::is_trivially_copyable` either. Why did you remove this check?

fuchsto (changes requested):

dash/include/dash/Matrix.h
devreal: Because they do. I just tested with Intel 17.0 and I am sure I also verified 16 back when we discussed #241. Keep in mind that the Intel compiler relies on the GNU-provided STL so you might have to load a newer GCC module (at least I have to do that on our Linux cluster).

devreal:
@fuchsto Just checked with Clang 3.8.0, works fine for me. What version did you test with? 

How would you check trivial-copyable in a custom type trait? `std::is_trivially_copyable` relies on the compiler builtin function `__is_trivially_copyable`...

fuchsto:
I don't see what you mean.
Type traits are not checked, they are defined. So, we would define additional traits in DASH, like we did with `is_atomic`.

fuchsto:
@devreal We could define traits depending on the compiler and its version. If this is just an issue with Cray compilers, it's not worth the effort, of course.

dash/include/dash/Matrix.h
fuchsto: Yes, that seems alright.
I actually did test exactly that, but it's apparently specific to our dev server. When reloading gcc after loading the icc module, there is some hickup and the newer version of gcc is not used.

fuchsto (approved):

PR #294, opened 2017-02-21
Make lbegin(uid) and lend(uid) private

### LINKED ISSUE #269: Passing a unit to GlobMem::lbegin?
I've been scrolling through the `GlobMem` code and found two things noteworthy:

1. There are basically two overloads for `lbegin` and `lend`, each. One takes no arguments and returns the begin of the local memory on the current unit. But what is the other supposed to return? The local pointer on an arbitrary (remote) unit? Any reason to expose this to the user? While it seems like a source of error it also has another flaw: the underlying `dart_gptr_getaddr` returns `NULL` for any unit that is not the current unit. 

2. Since `lbegin` and `lend` do not modify the current `GlobMem` instance it should be safe to keep just the `const` version of these functions and throw away the non-`const` versions to reduce code duplication. Any reason for keeping the non-`const` versions?

devreal:
Make `lbegin|lend(uid)` private in `Glob[Dynamic]Mem` and adapt callers. 

Fixes #269

### DIFF
--- a/dash/include/dash/GlobDynamicMem.h
+++ b/dash/include/dash/GlobDynamicMem.h
@@ -450,8 +450,8 @@ public:
                    "size:", bucket.size,
                    "lptr:", bucket.lptr);
     // Update local iteration space:
-    _lbegin = lbegin(_myid);
-    _lend   = lend(_myid);
+    _lbegin = update_lbegin(_myid);
+    _lend   = update_lend(_myid);
     DASH_ASSERT_EQ(_local_sizes.local[0], _lend - _lbegin,
                    "local size differs from local iteration space size");
     DASH_LOG_TRACE("GlobDynamicMem.grow",
@@ -623,8 +623,8 @@ public:
       _buckets.pop_back();
     }
     // Update local iterators as bucket iterators might have changed:
-    _lbegin = lbegin(_myid);
-    _lend   = lend(_myid);
+    _lbegin = update_lbegin(_myid);
+    _lend   = update_lend(_myid);
 
     DASH_LOG_TRACE("GlobDynamicMem.shrink",
                    "cumulative bucket sizes:",  _bucket_cumul_sizes[_myid]);
@@ -675,9 +675,9 @@ public:
     }
     // Update local iterators as bucket iterators might have changed:
     DASH_LOG_TRACE("GlobDynamicMem.commit", "updating _lbegin");
-    _lbegin = lbegin(_myid);
+    _lbegin = update_lbegin(_myid);
     DASH_LOG_TRACE("GlobDynamicMem.commit", "updating _lend");
-    _lend   = lend(_myid);
+    _lend   = update_lend(_myid);
     DASH_LOG_DEBUG("GlobDynamicMem.commit >", "finished");
   }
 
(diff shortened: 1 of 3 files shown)


dash/include/dash/GlobDynamicMem.h
```diff
@@ -1015,6 +906,60 @@ class GlobDynamicMem
   }
 
 private:
+
+  /**
+   * Native pointer of the final address of the local memory of
+   * a unit.
+   */
+  local_iterator update_lend(
```
fmoessbauer: isn't this a setter, so maybe `set_lend` would be better?

dash/include/dash/GlobMem.h
```diff
@@ -242,53 +242,9 @@ class GlobMem
...
-    DASH_LOG_TRACE_VAR("GlobMem.lbegin >", addr);
-    return static_cast<ElementType *>(addr);
-  }
-
-  /**
-   * Native pointer of the initial address of the local memory of
    * the unit that initialized this GlobMem instance.
    */
-  inline const ElementType * lbegin() const
+  constexpr const ElementType * lbegin() const
```
fmoessbauer: Better use `local_pointer` instead of directly specified type.

dash/include/dash/GlobMem.h
```diff
@@ -302,53 +258,12 @@ class GlobMem
...
-      DART_OK);
-    return static_cast<ElementType *>(addr);
-  }
 
   /**
    * Native pointer of the initial address of the local memory of
    * the unit that initialized this GlobMem instance.
    */
-  inline const ElementType * lend() const
+  constexpr const ElementType * lend() const
```
fmoessbauer: Better use `local_pointer` instead of directly specified type.

dash/include/dash/GlobMem.h
```diff
@@ -462,6 +377,52 @@ class GlobMem
   }
 
 private:
+
+  /**
+   * Native pointer of the initial address of the local memory of
+   * a unit.
+   * \param team_unit_id id of unit in \c dash::Team::All()
+   */
+  ElementType * update_lbegin(
```
fmoessbauer: Better use `local_pointer` instead of directly specified type.

fmoessbauer (approved):
Hi, looks good so far! I just noticed some minor points which could be enhanced. As I learned from the last PRs, we should use typedef types as return types instead of directly naming the type. IMO this makes our code much easier to maintain.

dash/include/dash/GlobMem.h
```diff
@@ -105,7 +105,7 @@ class GlobMem
    *       For example, \c dash::LocalAllocator is used in \c dash::Shared
    *       and only called at owner unit.
    */
-  inline GlobMem(
+  explicit GlobMem(
```
fuchsto, line 109: There is no use declaring this c'tor explicit as `updade_lbegin`, `update_lend` are not constexpr.
In non-strict modes compilers won't complain but the declaration would be misleading in documentation. Callers would expect `constexpr` behavior.

dash/include/dash/GlobMem.h
devreal: Done.

dash/include/dash/GlobDynamicMem.h
devreal: Technically it's not a setter. I refined the implementation and hopefully now it's clear that it's updating the internal state of the object.

fuchsto (changes requested):

dash/include/dash/GlobMem.h
devreal: Nope, `explicit` has nothing to do with `constexpr`. It just prevents implicit conversion from `size_type` to `GlobMem`.

fuchsto: D'aah, sorry, for whatever reason I mixed things up. I saw the `inline` and was on the constexpr train. Carry on.

fuchsto (approved):

Issue #289, opened 2017-02-17
Copy / Assignment of dash::Atomic


fuchsto:
Question on `dash::Atomic`:

Why is the copy constructor deleted, but assignment is not?

- `operator=(T)` is deleted
- `operator=(const self_t &)` is implicitly defaulted

That can't be right
When I explicitly delete assignment, tests don't compile

Also, the constructor `Atomic(T)` is not marked explicit, so it's an implicit conversion constructor
What's the intended semantics here?

The current implementation contradicts the documentation in dash/include/dash/Atomic.h

fuchsto:
For now, I added

```c++
self_t & operator=(const self_t & other) = default;
```

But assignment is still explicitly deleted, so this is not well-defined.

fmoessbauer:
Yes, that should be fixed. From the type-safety side it would be best to disable all forms of assignment, because the member is always set by DART using memory copy. But if all assignments are disabled, the dash algorithms like `dash::fill` will not work anymore, as they use local assignment internally.

AFAIK we have the following options:
- make assignment private and use friend
- allow assignment and live with `*(array.lbegin()) = 5` being possible

fuchsto:
It's actually to be expected that `dash::fill` would not work for `dash::Atomic<T>` containers value types.

Similarily, you can;t use `std::fill` for an STL `std::vector<std::atomic<T>>`.
So a container cannot be "filled" with atomics, they need to exist already. You would have to use `emplace_back` (not implying you should, though).

In general terms: semantics have priority over algorithm convenience. Especially for non-trivial types like atomics. Assignment and conversion (oh my!) should not be allowed.

We might be able to make `dash::for_each` work to initialize a `dash::Array<dash::Atomic<T>>` with a function that uses `elem.store` instead of assignment.

fmoessbauer:
With some knowledge gained the last couple of month, I want to re-think the topic:

> It's actually to be expected that `dash::fill` would not work for `dash::Atomic<T>` containers value types.

No, not really as the assignment operator of `std::Atomic` (`T operator=( T desired )`) is defined. This atomically replaces the value. Of course, that is not possible in DASH, as we have to use DART for atomically replacing the value at an DART address. However our implementation should provide a similar interface if possible. 

> Similarily, you can;t use `std::fill` for an STL `std::vector<std::atomic<T>>`.
So a container cannot be "filled" with atomics, they need to exist already.

[...]

PR #295, opened 2017-02-21
UnorderedMap of Atomic

fuchsto:
Addressing the implications of `dash::Atomic` value semantics, see #289 and #249

### DIFF
--- a/dash/include/dash/Atomic.h
+++ b/dash/include/dash/Atomic.h
@@ -84,10 +84,10 @@ public:
   }
 
 public:
-  template<typename T>
+  template<typename T_>
   friend std::ostream & operator<<(
-    std::ostream    & os,
-    const Atomic<T> & at);
+    std::ostream     & os,
+    const Atomic<T_> & at);
 
 }; // class Atomic
 
--- a/dash/include/dash/Atomic.h
+++ b/dash/include/dash/Atomic.h
@@ -53,21 +53,23 @@ private:
 public:
   typedef T value_type;
 
-  Atomic() = default;
-
-  Atomic(const Atomic<T> & other) = default;
-
-  self_t & operator=(const self_t & other) = default;
+  constexpr Atomic()                                 = default;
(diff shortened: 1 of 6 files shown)


fuchsto:
@fmoessbauer When testing changes to `dash::Atomic`, please use this PR for testing.

With the changes in this PR, using `dash::Atomic` as mapped type in `dash::UnorderedMap` works and modifying access to mapped values lead to atomic operations as expected.

dash/include/dash/map/UnorderedMap.h
```diff
@@ -613,9 +613,10 @@ class UnorderedMap
     /// The element to insert.
     const value_type & value)
   {
-    auto key    = value.first;
-    auto mapped = value.second;
-    DASH_LOG_DEBUG("UnorderedMap.insert()", "key:", key, "mapped:", mapped);
+    auto key = value.first;
```
devreal: Use reference here, i.e., `auto&`.

dash/include/dash/Atomic.h
```diff
@@ -48,11 +53,10 @@ class Atomic {
 public:
   typedef T value_type;
 
-  Atomic(const Atomic<T> & other) = delete;
+  Atomic() = default;
```
devreal: Can be `constexpr`

dash/include/dash/Atomic.h
```diff
@@ -73,6 +77,20 @@ class Atomic {
    */
   operator T() = delete;
 
+  bool operator==(const self_t & other) const {
```
devreal: Can be `constexpr` as well.

devreal (dismissed):

dash/include/dash/map/UnorderedMap.h
fuchsto: auto && even

fmoessbauer (approved):

PR #296, opened 2017-02-21
Fixed Matrix/NArray const view proxies

### LINKED ISSUE #283: NArray/Matrix as const throws compiler error

fuchsto:
Fixes #283 

Actual fix is a missing type cast in `LocalMatrixRef`:

https://github.com/dash-project/dash/pull/296/files#diff-3727b32cf9caed9a261d978fb1a5c2ceR442


BenjaProg (approved):
Looks good to me 👍

Issue #168, opened 2016-11-28
Coordinate-based access to matrix.local throws std::out_of_range on unit!=0


fuerlinger:
Using coordinate-based access to the local storage of a matrix throws when not on unit 0. Access works fine using chained subscripts:

````
int main(int argc, char* argv[])
{
  dash::init(&argc, &argv);

  auto myid = dash::myid();
  auto size = dash::size();

  dash::NArray<int, 2> mat(4,10);

  if(myid==0 ) { std::iota(mat.begin(), mat.end(), myid); }
  dash::barrier();

  if( myid==1 ) {
    // cout << mat.local(0,0) << endl; // throws std::out_of_range
    cout << mat.local[0][0] << endl; // works fine
  }

  dash::finalize();
}
````

fuchsto:
Created branch `bug-168-mat-coords`.

fuchsto:
Fixed in #298

PR #298, opened 2017-02-22
Coord-based access to local NArray elements

fuchsto:
Addresses #168

### DIFF
--- a/dash/include/dash/algorithm/Generate.h
+++ b/dash/include/dash/algorithm/Generate.h
@@ -46,8 +46,6 @@ void generate (
   auto llast  = lrange.end;
 
   std::generate(lfirst, llast, gen);
-
-  first.team().barrier();
 }
 
 } // namespace dash
--- a/dash/include/dash/algorithm/Generate.h
+++ b/dash/include/dash/algorithm/Generate.h
@@ -46,6 +46,8 @@ void generate (
   auto llast  = lrange.end;
 
   std::generate(lfirst, llast, gen);
+
+  first.team().barrier();
 }
 
 } // namespace dash
--- a/dash/include/dash/matrix/internal/LocalMatrixRef-inl.h
+++ b/dash/include/dash/matrix/internal/LocalMatrixRef-inl.h
@@ -368,10 +368,6 @@ LocalMatrixRef<T, NumDim, CUR, PatternT>
 ::operator[](
   index_type pos) const
 {
-  DASH_LOG_TRACE("LocalMatrixRef.[]()",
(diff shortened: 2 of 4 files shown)


dash/include/dash/algorithm/Generate.h
```diff
@@ -46,6 +46,8 @@ void generate (
   auto llast  = lrange.end;
 
   std::generate(lfirst, llast, gen);
+
+  first.team().barrier();
```
fmoessbauer: Are these barriers really intended? This accounts for all dash algorithms.

dash/include/dash/matrix/internal/LocalMatrixRef-inl.h
```diff
@@ -321,14 +323,21 @@ ::at(
...
+  DASH_LOG_TRACE("LocalMatrixRef.at()",
+                 "ndim:",     NumDim,
+                 "curdim:",   CUR,
+                 "l_coords:",   _refview._coord,
+                 "l_viewspec:", _refview._l_viewspec);
   return local_at(
            _refview._mat->_pattern.local_at(
              _refview._coord,
-             _refview._viewspec));
+             _refview._l_viewspec));
```
fmoessbauer, line 340: This is the actual fix, isn't it?

fmoessbauer (approved):
Good Work!

dash/include/dash/matrix/internal/LocalMatrixRef-inl.h
fuchsto: Supposedly this change is all that is required to fix the defect, yes.
`_coord` refers to local index range, `_viewspec` to global index range (global view to local index range, that is). Should be viewspec in local index range, of course. Fixed tests on my machines.

Also added unit test to cover this use case.

dash/include/dash/algorithm/Generate.h
fuchsto: There was a discussion on this some time ago. Didn't we decide for barriers?

fmoessbauer: Ok, I have no idea ... We should document that in the algorithm section. A quick survey showed that some other algorithms (e.g. [fill](https://github.com/dash-project/dash/blob/f716024fe3d8a5408dbaa2551ba1c8e0153fab42/dash/include/dash/algorithm/Fill.h#L35)) do not use barriers.

fuchsto: Ok, will remove the barrier before merging.

PR #299, opened 2017-02-23
Add support for static code analysis using scan-build from llvm

fmoessbauer:
This PR adds a build script which generates a static code analysis report on the fly while building. Just invoke `./build.analyze.sh` and enjoy.

A sample report of development can be seen [here](https://felixmoessbauer.de/temp/build-report/).

### DIFF
--- a/build.analyze.sh
+++ b/build.analyze.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 # This script runs scan-build to perform a static code analysis
 # https://clang-analyzer.llvm.org/scan-build.html
 #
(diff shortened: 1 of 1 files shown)


build.analyze.sh
```diff
@@ -0,0 +1,113 @@
...
+# This script runs scan-build to perform a static code analysis
+# https://clang-analyzer.llvm.org/scan-build.html
+#
+# The actual compilers have to be set in the CCC prefix env. variables. e.g:
+# export CCC_CC=clang-3.8
+# export CCC_CXX=clang++3.8
+
+BUILD_DIR=./build.analyze
+REPORT_DIR=report            # relative to BUILD_DIR
+BUILD_WRAPPER="scan-build-3.8"
```
fuchsto: Why restricting to 3.8?

fuchsto (changes requested):

build.analyze.sh
fmoessbauer: Because unfortunately there is no `scan-build`, so otherwise I would have to parse the path and look for the newest version. IMO this is a good trade off as the variable can easily be set.

fuchsto: My clang installations provide soft-links or have etc/alternatives configured. Same for custom builds. Did you try `/bin/env scan-build` as fallback?

fmoessbauer: Yes, but at least in our docker containers (which are based on ubuntu) this does not work for clang installations from rpm.

fuchsto: Okay, then how about defining `BUILD_WRAPPER` for docker runs and defaulting to `scan-build`?

fuchsto: ... so: same as for `PAPI_HOME`, in principle.

fmoessbauer: Would be possible. We could also provide a fallback to `scan-build-3.8` if `scan-build` was not found

fmoessbauer: That's propably best. But then don't call it `BUILD_WRAPPER` but `SCANBUILD_BIN`.

fuchsto: Sounds fair enough an pragmatic. Make it so.

fuchsto (approved):

fuchsto:
Nice, pushes compliance to CII best practices to 97%.
Integrating static code analysis in CI would give us full compliance.

Issue #216, opened 2016-12-21
Improvements to dash::accumulate


devreal:
The current implementation of `dash::accumulate` has a flaw: only unit 0 (randomly chosen?) actually receives a valid result. This is neither documented nor does the user have a chance to influence this. I seem to remember that we had a similar discussion at our last F2F. IMO, either the user can choose which unit receives the result or it should be broadcast to all units. At the very least, it has to be documented...

Also, `dash::accumulate` should use `dart_accumulate` whenever possible. 

Discovered while working on #212.

fuchsto:
The current realization of `dash::accumulate` is somewhat irritating, yes. Oh yes.

First things first:

- `dart_accumulate` semantically corresponds to `MPI_Accumulate`: `C[i] = reduce(A[i], B[i])`
- `dash::accumulate` semantically corresponds to `std::accumulate`: `Acc = reduce([AccInit] v A[s...e])`

Consequently, `dart_accumulate` is ***not*** a semantically suitable backend for `dash::accumulate`.
C++ and MPI differ in their definitions of "accumulate" (C++ uses the term correctly, however `std::transform` should be named `std::map`, but can't, because of the container. It's a bloody mess).

Semantics of `dart_accumulate` correspond to `dash::transform` (~ `std::transform`), and it is in fact used as its communication backend:

https://github.com/dash-project/dash/blob/development/dash/include/dash/algorithm/Transform.h#L252

The best I could do for this clash of specs was to document it:

- https://github.com/dash-project/dash/blob/development/dash/include/dash/algorithm/Accumulate.h#L70
- https://github.com/dash-project/dash/blob/development/dash/include/dash/algorithm/Accumulate.h#L118

Standardese aside, the current implementation is broken.
I found another defect: the initial value is accumulated in every local partial result, but should only be accumulated once, at unit 0. Currently implemented behavior is:

~~~
// (a, b] = { 1,2,3, 4,5,6 }
// --> accumulate(a,b,+,10) = 31

unit 0:                          | unit 1:
=================================+==================================
dash::accumulate(a, b, +, 10);   | dash::accumulate(a, b, +, 10);
--> local result: 10 + 1 + 2 + 3 | --> local result: 10 + 4 + 5 + 6
    = 16                         |     = 25
----------------------------- barrier ------------------------------
result = 16 + 25 = 41 != 31
~~~

devreal:
[...]

Issue #272, opened 2017-02-10
Launch policies for algorithms


devreal:
At the moment, accesses to global memory of DASH containers through subscript operators are serialized using blocking put operations (`dart_put_blocking`) in `GlobRef`, potentially resulting in poor performance for naive users. I understand that this is useful to avoid synchronization problems in the general case, but allowing asynchronous access through subscript operators would provide an easy way of improving performance of remote access to multiple elements while using the same (simple to use) interface. 

Hence, we could introduce asynchronous epochs, which have an explicit begin and end, e.g., 

```
dash::Array<int> arr(100);
arr.start_epoch();
if (dash::myid() == 0) {
  for (int i = 0; i < arr.size(); ++i) {
    arr[i] = i; // not immediately visible
  }
}
arr.end_epoch(); 
// all changes visible now
``` 
Within an epoch, `GlobRef` will issue non-blocking `put` (+`get`?) operations and data transfers are guaranteed to complete only after the end of the epoch.  

Note: I am aware of `dash::put_value_async` in `OneSided` but I think it's not intuitively usable and quite wordy, e.g., 

```
dash::Array<int> arr(100);
if (dash::myid() == 0) {
  for (int i = 0; i < arr.size(); ++i) {
    dash::put_value_async(i, arr[i]); // is this correct?
  }
}
dash::fence(arr[0]); // sufficient to fence on the first element?
// all changes visible now
``` 

The documentation is not clear on the use of these methods and there are no tests/examples either. I also cannot efficiently mix `dash::put_value_async` and `operator[]` because the latter causes a flush. The idea of asynchronous epochs closely resembles the model employed by MPI-RMA (`MPI_Put` and `MPI_Flush`), just with a nicer interface :)

Please let me know what you think. Maybe there have been discussions about it earlier?

fuerlinger:
Something like that is definitely needed and there was previous discussion about it. Some parts of the functionality described below are in DASH already but I think its not 100% working at the moment.

The idea to provide similar functionality as you describe with an `.async` modifier, like so:

~~~
for (int i = 0; i < arr.size(); ++i) {
  arr.async[i] = i; // not immediately visible
}
arr.flush(); // or arr.fence();
~~~

While `arr[i]` returns a `dash::GlobRef` `arr.async[i]` returns a `dash::AsyncGlobRef` and that would cause `put` and `get` operations without `flush`.

rkowalewski:
We have `GlobAsyncRef` which should satisfy all the requirements

[...]

Issue #104, opened 2016-11-15
Execution Policies for Algorithms


fuchsto:
Execution policies should be introduced that allow to specify the behavior of DASH algorithms.

Usage should correspond to execution policies as proposed for the STL:
http://en.cppreference.com/w/cpp/experimental/transform_reduce

For example:

~~~cpp
// The specialization of dash::transform for one-sided execution results in
// MPI_Accumulate + MPI_Win_flush:

// defaults to one-sided execution policy:
dash::transform(
  in_a_first, in_a_last,
  in_b_first,
  out_first);
// specify one-sided execution explicitly:
dash::transform(
  dash::execution::onesided,
  in_a_first, in_a_last,
  in_b_first,
  out_first);

// For collective execution, dash::transform resolves to MPI_Allreduce which
// is more efficient but requires a coordinated schedule: 
dash::transform(
  dash::execution::collective,
  in_a_first, in_a_last,
  in_b_first,
  out_first);
~~~

fuchsto:
Related to launch policies, see #272

Issue #193, opened 2016-12-09
DASH Task Concept Specification


fuchsto:
Note for later: ensure the DASH Task concept is compatible to the Kokkos task API.

fuchsto:
@devreal Obviously, there is a strong dependency on the DART task concept specifications. Let's have a telco some time soon!

devreal:
I've been meaning to look at the Kokkos task API and whether this is something we should consider. Will do so during the next days and report back. Pretty much any time next week works for me, just PM me.

devreal:
I just read through the document at https://github.com/kokkos/kokkos/blob/master/doc/Kokkos_PG.pdf. To sum up what we need:

1. Parallel dispatch (chapter 7): `parallel_for`, `parallel_reduce`, and `parallel_scan`. Implementation should be straight-forward and possible with the current state of the DART tasking implementation. 
2. Hierarchical parallelism (chapter 8) : so far I have not implemented thread-teams and the like so there is only one set of threads executing tasks. I do not immediately see the benefit of having thread teams (except if we talk about accelerators like GPUs) and it will certainly add more complexity to DART to support it. 

Supporting the parallel dispatch is certainly helpful for easy parallelization. However, my vision of tasking (explicit task definition with dynamic dependencies) diverts quite a bit from what Kokkos seems to offer. But maybe the guide is not up-to-date. Let's discuss this on the phone call.

fuchsto:
Perfect, I

fuchsto:
(oops)

PR #300, opened 2017-02-24
[WIP] Introducing execution- and launch policies

fuchsto:
Addresses issues #272 #104 #216


fuchsto:
Related to launch policies: Task concept specification (see #193)

dash/include/dash/Execution.h
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Parallel execution policy.
+   */
+  class parallel_policy { };
+
+  /**
+   * Parallel non-sequential execution policy.
+   */
+  class parallel_unsequenced_policy { };
```
devreal, line 32: What is the difference in semantics between `parallel_policy` and `parallel_unsequenced_policy`?

dash/include/dash/algorithm/Copy.h
```diff
@@ -88,12 +93,14 @@ OutputIt copy(
...
-  typename ValueType,
+  class ExecutionPolicy,
+  class ValueType,
   class GlobInputIt >
-dash::Future<ValueType *> copy_async(
-  InputIt  in_first,
-  InputIt  in_last,
-  OutputIt out_first);
+dash::Future<ValueType *> copy(
+  ExecutionPolicy && policy,
```
devreal, line 100: The execution policy is ignored atm, right?

devreal (commented):
Thanks for getting the ball rolling. Questions inline.

dash/include/dash/Execution.h
rkowalewski: This is well described on [cppreference](http://en.cppreference.com/w/cpp/algorithm/execution_policy_tag_t).

rkowalewski (approved):
Thanks. It perfectly matches the C++17 Specs.

dash/include/dash/algorithm/Copy.h
fuchsto: Yes, I just pinned the branch as PR to simplify discussion. It's work in progress (= `[WIP]` in title).

fuchsto:
I think I got the required concepts figured out.
This summary is mostly relevant to @devreal @fmoessbauer @fuerlinger 

**Execution Policies**

In C++17, execution policies are introduced in algorithm function interfaces and passed as first argument:

```c++
template< class ExecutionPolicy, class InputIt, class UnaryFunction2 >
void for_each( ExecutionPolicy&& policy, InputIt first, InputIt last, UnaryFunction2 f );
```

It is therefore trivial to specialize / overload algorithm interfaces for execution policies.
From the code-mechanics perspective, we can comfortably extend the STL execution policy concept.
Introducing additional traits and some CRTP adapters should do.

The actual effort is in rather formal aspects. Which is nice, because science.

In principle, this boils down to additional PGAS-specific dimensions for execution policies.
Some examples:

- **collaboration**, such as `collective`, `asymmetric`, `leader`
- **synchronization** such as `nowait` (as known from OpenMP, would skip implicit barriers)
- **unit-level parallelization**, the vanilla STL execution policies, e.g. `parallel` would allow units to spawn OpenMP threads

... you get the idea. Policies can be arbitrarily combined with `foo_policy | bar_policy` as usual.

**Launch Policies**

STL concepts do not cover our requirements regarding launch policies.
In vanilla C++, asynchronous and deferred execution is provided by the `std::launch` and `std::async` etc. function interfaces and algorithms are agnostic of launch policies:

```c++
template< class Function, class... Args >
std::future<typename std::result_of<Function(Args...)>::type>
    async( std::launch policy, Function&& fun, Args&&... args );
```

Parameter `fun` is the actual callable object.

Our algorithm variants depend on launch policies, however. For example, asynchronous copying `dash::copy_async` and blocking copying `dash::copy` are independent interfaces, and I don't think there is practicable alternative that is correct and efficient in the general case.
Implementing blocking and deferred variants as wrappers of a single asynchronous implementation seems sensible, but in fact it hurts performance pretty badly.

I sketched out some alternatives, like: 

- Grouping algorithms by launch policy in namespaces, like `dash::async::copy`. Does not allow combination of policies, so it's not an option.
- Meta-magic: `dash::async` could just replace callables depending on the launch policy. For example, passing `dash::copy` with `dash::launch::async` would invoke `dash::copy_async` instead. Nearly impossible to maintain and extremely irritating as the executed task is different from the specified callable.

So far, my initial approach based on phantom- and container proxy types looks like the best option to me. Instead of namespaces or function template specializations, algorithm variants are realized as overloads for iterator traits.

An iterator for async operations would have `dash::iterator_traits<It>::async_policy_tag` specified.
Containers like `dash::Array` use adapters or proxies to convert iterators between policies.
This is already implemented in `dash::Array<T>::async` and allows usage like this:

```c++
dash::Array<X> array(size);

// Default:
auto copy_end = dash::copy(array.begin(), array.end(), dest);
// -> dash::Array<X>::iterator

// Async:
auto copy_end = dash::copy(array.async.begin(), array.async.end(), dest);
// current implementation:
// -> dash::GlobAsyncIter<T>
// will be:
// -> dash::Future<dash::Array<X>::iterator>

// Lazy:
auto copy_end = dash::copy(array.deferred.begin(), array.deferred.end(), dest);
// -> dash::Array<X>::iterator
```

This is very close to the STL specs and there is no way to get it wrong for developers.
No maintenance effort in algorithms at all, no coupling of concerns, much rejoice.

But also, it is important to note that we now can actually define `dash::async` and `dash::launch` like in STL, and replace iterator parameters automatically (but, different from replacing the callable parameter as described above, with robust semantics):

```c++
// Developer writes:
auto fut_copy_end = dash::async(
             dash::launch::async,
             dash::copy<Foo>,
             array.begin(),
             array.end());
// --> resolves to:
auto fut_copy_end = dash::async(
             dash::launch::async,
             dash::copy<Foo>,
             array.begin().async,  // same as array.async.begin()
             array.end().async);   // same as array.async.end()
```

This is semantically and conceptually robust:

- `GlobIter<T, ..., dash::launch::async>` satisfies `GlobIter<T, ..., dash::launch::any>` and
- The developer specified the launch policy explicitly, so the principle of least astonishment is respected, the behavior and signature corresponds to `std::async`.

So long story short: There is a chance we can have the cake and eat it, too.

In the current implementation in branch `development`, `array.async` is a fully specialized proxy accessor with `dash::ArrayAsyncRef<T>::iterator = GlobAsyncIter<T>`.
I added those as proof of concept., actually these specializations are not necessary for every policy, we can use traits as well.
This is preferrable as we otherwise had to define a fully specialized type for any combination of launch policies, and adding a policy would be the opposite of fun.

Looking forward to your thoughts!

devreal:
Thanks for this brain dump! I think I get the idea. 

__On Execution Policies__
I think this is clear and it helps to get rid of some of the confusion we currently have, e.g., which unit receives the results of a `dash::accumulate` call. One thing to note, though:

> unit-level parallelization, the vanilla STL execution policies, e.g. parallel would allow units to spawn OpenMP threads

Keep in mind that DART will eventually have it's own thread-pool independent of OpenMP. OpenMP is just too restrictive to use it as a general-purpose task scheduler (unfortunately). Executing tasks in DART and OpenMP in parallel likely leads to contention and performance degradation. 

__On Launch Policies__
Just to put this into perspective of general tasking in DASH: calling `dash::async` on `dash::copy` with `dash::launch::async` will convert `array.begin()` to `array.begin().async`, which issues an asynchronous copy and the returned `dash::Future` can be used to wait for completion (through a `flush`), right? 

Now, how is this implemented for other DASH algorithms? Do we provide overrides for them as well? For example, combining `dash::launch::async` with the `collective` execution policy sounds like a bad idea...

My idea of tasking in DASH currently is that calling `dash::async` with `dash::launch::async` specified on any other function will result in an asynchronously executed action, i.e., a task that is scheduled through DART to be executed at some point in time. Again, the `dash::future` is used to wait for completion. The DART side of this is pretty much there (although maybe not extensively tested). 

There are two concepts that I would like to raise awareness of:

1. __Task dependencies__ 
I've been working on a way to define an ordering of tasks through data dependencies in DART by specifying arbitrary memory locations in global memory (similar to OpenMP task data dependencies). I haven't thought much about the DASH abstraction although I think `dash::async` would suite well. This would require `dash::async` to optionally accept dependency definitions. 

2. __Fire-and-forget or Bulk Tasks__
Similar to OpenMP tasks, users may want to define tasks without explicitly waiting for them individually. Instead, tasks are defined in bulk and waited for by a single statement. This is almost necessary if you want to taskify one or multiple iterations of your algorithm. We could either introduce a launch policy for `dash::async` for this type of tasks (returning an _empty_ `dash::Future`) or have a separate abstraction for it (such as `dash::Task::create` and `dash::Task::complete`). I'm OK with both, I just want to make sure we are covering this use-case. 

We should definitely discuss this in Garching. I think we are arriving at a state where the tasking in DART is ready to be made available (though maybe not in its full feature-set). I've been holding it back until after the (ominous) next release so far.

fuchsto:
@devreal 

> Keep in mind that DART will eventually have it's own thread-pool independent of OpenMP

Of course! Just using OpenMP as a reference, but we want light-weight tasks, of course.

> My idea of tasking in DASH currently is that calling dash::async with dash::launch::async specified
> on any other function will result in an asynchronously executed action

That's not a contradiction: the conversion to `array.begin().async` is internal, the visible part of the interface is just the return type `dash::Future<T>` which is also what tasks would return.
It's definitely important to support "real" task models like yours, I kept that in mind in my proposal.

But on the other hand, how would you replace the current `dash::copy_async` by tasks?
Asynchronous one-sided MPI will be impossible to beat in efficiency, wrapping a blocking `dash::copy` in a task would not help.

I assume that eventually, the current async interface of DART will be wrapped in a proper task interface so the C++ layer could prefer tasks to abstract async operations - Still, we must be able to support these "async-specialized implementation" cases like `dash::copy_async`, too.

We need both, and I think / hope the proposed interface could do.

devreal:
> But on the other hand, how would you replace the current `dash::copy_async` by tasks?

Not at all, it does not make sense to implement `dash::copy_async` in terms of tasks as the time is mostly spent in data transfer. I was just wondering about the other DASH algorithms :)

> We need both, and I think / hope the proposed interface could do.

Absolutely, I just wanted to make sure we are on the same page here. I will present my current status in Garching and based on that we can discuss the details of the tasking abstraction in it's full beauty :)

fuchsto:
Ah, I got a good feeling about this.

Ehhmm, when exactly is the SPPEXA name-dancing in Garching again?

fuchsto:
Ah, here we go:

<http://www.sppexa.de/sppexa-activities/annual-plenary-meeting/2017.html>

That's a good time frame. Did you happen to review HPX and Kokkos (also: Alpaka), yet?
I'm about to make DASH views and Kokkos views fancy each other.

devreal:
I had a look at Kokkos (see my comments in #193) and will attend a workshop on HPX here at HLRS next week. I haven't looked at Alpaca, yet. Could that be relevant as well?

fuerlinger:
Yes, Alpaka is relevant as well. I'm just here at the Alpaka Kickoff meeting, will report back tomorrow.

devreal:
@fuerlinger Gotcha. If you can provide us with some material I'll be happy to look at it before the Garching meeting.

devreal:
Can we get this into 0.3.0? If not, I would start working on a quick fix for `dash::accumulate` and document the deficiencies so we can close #216.

fuchsto:
@devreal I can get this into 0.3.0 this week.

pascalj:
@fuchsto May I hijack this PR to add the few lines needed for Alpakas `for_each` and maybe `transform` example? It currently does not rely on these changes, but it wouldn't be much work to base the MEPHISTO-stuff (in the MEPHISTO repo) on `dash/include/dash/Execution.h`

PR #302, opened 2017-02-25
Many stability fixes coming up from static code analysis

fmoessbauer:
This PR fixes all errors (DART and DASH) reported by clang-analyze except the following:
- locality module (both DASH and DART)
- dead initialization (=unused variable) in examples and tests (this cannot be 100% avoided as different build modes read them)

This PR also coveres some serious bugs and a huge amount of memory leaks. Also have a look at the reports:
- [before fixes](https://felixmoessbauer.de/temp/build-report/)
- [after fixes](https://felixmoessbauer.de/temp/build-report-new/)

@devreal please review the fixes on the DART side carefully, especially `DART_GPTR_COPY` and the level calculation in dart buddy allocator.

If you want to generate the build report on your machine, just execute `build.analyze.sh` and get a cup of coffee. It will take roughly 15 minutes to complete. However you can speed it up by disabling tests and examples.

### DIFF
--- a/build.analyze.sh
+++ b/build.analyze.sh
@@ -31,7 +31,7 @@ if [ "$SCANBUILD_OPTS" = "" ]; then
   SCANBUILD_OPTS="-analyze-headers -plist-html"
 fi
 SCANBUILD_OPTS="-o $REPORT_DIR ${SCANBUILD_OPTS}"
-SCANBUILD_OPTS="--force-analyze-debug-code ${SCANBUILD_OPTS}"
+SCANBUILD_OPTS="--force-analyze-debug-code -v ${SCANBUILD_OPTS}"
 
 which $BUILD_WRAPPER ||
   (echo "This build requires $BUILD_WRAPPER. Set env. var SCANBUILD_BIN" \
@@ -123,8 +123,8 @@ mkdir -p $BUILD_DIR/$REPORT_DIR
                         -DENABLE_HDF5=ON \
                         \
                         -DBUILD_EXAMPLES=OFF \
-                        -DBUILD_TESTS=ON \
-                        -DBUILD_DOCS=ON \
+                        -DBUILD_TESTS=OFF \
+                        -DBUILD_DOCS=OFF \
                         \
                         -DIPM_PREFIX=${IPM_HOME} \
                         -DPAPI_PREFIX=${PAPI_HOME} \
--- a/build.analyze.sh
+++ b/build.analyze.sh
@@ -9,12 +9,14 @@
 BUILD_DIR=./build.analyze
 REPORT_DIR=report            # relative to BUILD_DIR
 BUILD_WRAPPER="${SCANBUILD_BIN}";
-ANALYZE_OPTS="-o $REPORT_DIR -analyze-headers -plist-html"
 
 
 # try to find build wrapper
 if [ "$BUILD_WRAPPER" = "" ]; then
   BUILD_WRAPPER="scan-build"
+fi
+if [ "$SCANBUILD_OPTS" = "" ]; then
+  SCANBUILD_OPTS="-o $REPORT_DIR --analyze-headers --plist-html"
 fi
 
 which $BUILD_WRAPPER ||
@@ -73,7 +75,7 @@ fi
 # Configure with default release build settings:
 rm -Rf $BUILD_DIR/*
 mkdir -p $BUILD_DIR/$REPORT_DIR
-(cd $BUILD_DIR && $BUILD_WRAPPER $ANALYZE_OPTS \
+(cd $BUILD_DIR && $BUILD_WRAPPER $SCANBUILD_OPTS \
                   cmake3 -DCMAKE_BUILD_TYPE=Release \
                         -DENVIRONMENT_TYPE=default \
                         -DENABLE_COMPTIME_RED=OFF \
@@ -116,7 +118,7 @@ mkdir -p $BUILD_DIR/$REPORT_DIR
                         -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
                         ../ && \
  await_confirm && \
- $BUILD_WRAPPER $ANALYZE_OPTS make -j 4) && \
+ $BUILD_WRAPPER $SCANBUILD_OPTS make -j 4) && \
  (cp $BUILD_DIR/compile_commands.json .) && \
 exit_message
 
--- a/build.analyze.sh
+++ b/build.analyze.sh
@@ -71,10 +71,10 @@ fi
 # installed.
 
 # Configure with default release build settings:
-mkdir -p $BUILD_DIR/$REPORT_DIR
 rm -Rf $BUILD_DIR/*
+mkdir -p $BUILD_DIR/$REPORT_DIR
 (cd $BUILD_DIR && $BUILD_WRAPPER $ANALYZE_OPTS \
-                  cmake -DCMAKE_BUILD_TYPE=Release \
+                  cmake3 -DCMAKE_BUILD_TYPE=Release \
                         -DENVIRONMENT_TYPE=default \
                         -DENABLE_COMPTIME_RED=OFF \
                         \
@@ -106,7 +106,7 @@ rm -Rf $BUILD_DIR/*
                         -DENABLE_PLASMA=ON \
                         -DENABLE_HDF5=ON \
                         \
-                        -DBUILD_EXAMPLES=ON \
+                        -DBUILD_EXAMPLES=OFF \
                         -DBUILD_TESTS=ON \
                         -DBUILD_DOCS=ON \
                         \
--- a/build.analyze.sh
+++ b/build.analyze.sh
@@ -16,8 +16,10 @@ if [ "$BUILD_WRAPPER" = "" ]; then
   BUILD_WRAPPER="scan-build"
 fi
 if [ "$SCANBUILD_OPTS" = "" ]; then
-  SCANBUILD_OPTS="-o $REPORT_DIR --analyze-headers --plist-html"
+  SCANBUILD_OPTS="-analyze-headers -plist-html"
 fi
+SCANBUILD_OPTS="-o $REPORT_DIR ${SCANBUILD_OPTS}"
+
 
 which $BUILD_WRAPPER ||
   (echo "This build requires $BUILD_WRAPPER. Set env. var SCANBUILD_BIN" \
--- a/build.analyze.sh
+++ b/build.analyze.sh
@@ -74,9 +74,14 @@ fi
 mkdir -p $BUILD_DIR
 mkdir -p $REPORT_DIR
 rm -Rf $BUILD_DIR/*
+rm -Rf $REPORT_DIR/*
 (cd $BUILD_DIR && $BUILD_WRAPPER $ANALYZE_OPTS \
                   cmake -DCMAKE_BUILD_TYPE=Release \
                         -DENVIRONMENT_TYPE=default \
+                        \
+                        -DCMAKE_C_COMPILER=clang \
+                        -DCMAKE_CXX_COMPILER=clang++ \
+                        \
                         -DDART_IF_VERSION=3.2 \
                         -DINSTALL_PREFIX=$HOME/opt/dash-0.3.0/ \
                         -DDART_IMPLEMENTATIONS=mpi \
@@ -98,7 +103,7 @@ rm -Rf $BUILD_DIR/*
                         -DENABLE_LIKWID=OFF \
                         -DENABLE_HWLOC=ON \
                         -DENABLE_PAPI=ON \
-                        -DENABLE_MKL=ON \
+                        -DENABLE_MKL=OFF \
                         -DENABLE_BLAS=ON \
                         -DENABLE_LAPACK=ON \
                         -DENABLE_SCALAPACK=ON \
@@ -115,7 +120,7 @@ rm -Rf $BUILD_DIR/*
                         -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
                         ../ && \
  await_confirm && \
- $BUILD_WRAPPER $ANALYZE_OPTS make -j 4) &&
+ $BUILD_WRAPPER $ANALYZE_OPTS make -j 4) && \
  (cp $BUILD_DIR/compile_commands.json .) && \
 exit_message
 
--- a/build.analyze.sh
+++ b/build.analyze.sh
@@ -29,7 +29,7 @@ fi
 await_confirm() {
   if ! $FORCE_BUILD; then
     echo ""
-    echo "   To build using these settings, hit ENTER"
+    echo "   To build and analyze using these settings, hit ENTER"
     read confirm
   fi
 }
@@ -71,16 +71,12 @@ fi
 # installed.
 
 # Configure with default release build settings:
-mkdir -p $BUILD_DIR
-mkdir -p $REPORT_DIR
+mkdir -p $BUILD_DIR/$REPORT_DIR
 rm -Rf $BUILD_DIR/*
-rm -Rf $REPORT_DIR/*
 (cd $BUILD_DIR && $BUILD_WRAPPER $ANALYZE_OPTS \
                   cmake -DCMAKE_BUILD_TYPE=Release \
                         -DENVIRONMENT_TYPE=default \
-                        \
-                        -DCMAKE_C_COMPILER=clang \
-                        -DCMAKE_CXX_COMPILER=clang++ \
+                        -DENABLE_COMPTIME_RED=OFF \
                         \
                         -DDART_IF_VERSION=3.2 \
                         -DINSTALL_PREFIX=$HOME/opt/dash-0.3.0/ \
--- a/build.analyze.sh
+++ b/build.analyze.sh
@@ -11,6 +11,18 @@ REPORT_DIR=report            # relative to BUILD_DIR
 BUILD_WRAPPER="${SCANBUILD_BIN}";
 
 
+
+## !! NOTE !!
+#
+#  See documentation of scan-build for details on recommended build
+#  configuration:
+#
+#  https://clang-analyzer.llvm.org/scan-build.html#recommended_debug
+#
+##
+
+
+
 # try to find build wrapper
 if [ "$BUILD_WRAPPER" = "" ]; then
   BUILD_WRAPPER="scan-build"
@@ -19,7 +31,7 @@ if [ "$SCANBUILD_OPTS" = "" ]; then
   SCANBUILD_OPTS="-analyze-headers -plist-html"
 fi
 SCANBUILD_OPTS="-o $REPORT_DIR ${SCANBUILD_OPTS}"
-
+SCANBUILD_OPTS="--force-analyze-debug-code ${SCANBUILD_OPTS}"
 
 which $BUILD_WRAPPER ||
   (echo "This build requires $BUILD_WRAPPER. Set env. var SCANBUILD_BIN" \
@@ -78,7 +90,7 @@ fi
 rm -Rf $BUILD_DIR/*
 mkdir -p $BUILD_DIR/$REPORT_DIR
 (cd $BUILD_DIR && $BUILD_WRAPPER $SCANBUILD_OPTS \
-                  cmake3 -DCMAKE_BUILD_TYPE=Release \
+                  cmake3 -DCMAKE_BUILD_TYPE=Debug \
                         -DENVIRONMENT_TYPE=default \
                         -DENABLE_COMPTIME_RED=OFF \
                         \
@@ -120,7 +132,7 @@ mkdir -p $BUILD_DIR/$REPORT_DIR
                         -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
                         ../ && \
  await_confirm && \
- $BUILD_WRAPPER $SCANBUILD_OPTS make -j 4) && \
+ $BUILD_WRAPPER $SCANBUILD_OPTS make) && \
  (cp $BUILD_DIR/compile_commands.json .) && \
 exit_message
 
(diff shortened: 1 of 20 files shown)


dart-impl/mpi/include/dash/dart/mpi/dart_globmem_priv.h
```diff
@@ -6,6 +6,7 @@
 #define DART_GPTR_COPY(gptr_, gptrt_)                       \
   do {                                                      \
     gptr_.unitid = gptrt_.unitid;                           \
+    gptr_.flags  = gptrt_.flags;                            \
```
fuchsto: woooaah :D

fuchsto (approved):
Marvelous!

fuchsto:
Reports in the "Dead store" category are false positives, mostly variables for logging that appear unused because the scan was build with logging disabled.
The nullptr-derefs, however ...
Well, let's say I'm eager to create a pull request,

dart-impl/mpi/include/dash/dart/mpi/dart_globmem_priv.h
devreal: Why do we need this macro anyway? `gptr` can be directly assigned so there is no need to have a dedicated copy macro (which is not used anyway afair).

dart-impl/mpi/src/dart_mem.c
```diff
@@ -166,7 +171,7 @@ dart_buddy_alloc(struct dart_buddy * self, size_t s) {
 		}
 		for (;;) {
 			level--;
-			length *= 2;
+//    length *= 2;
```
devreal: Are you sure about this? What was the warning from the static analyzer?

dart-impl/mpi/src/dart_team_group.c
```diff
@@ -27,9 +27,10 @@
 
 static struct dart_group_struct* allocate_group()
 {
-  struct dart_group_struct* group = malloc(sizeof(struct dart_group_struct));
+  struct dart_group_struct* group = (struct dart_group_struct *) malloc(
+                                      sizeof(struct dart_group_struct));
   return group;
-}
+};
```
devreal: The semi-colon is not needed here.

dart-impl/mpi/src/dart_team_group.c
```diff
@@ -27,9 +27,10 @@
 
 static struct dart_group_struct* allocate_group()
 {
-  struct dart_group_struct* group = malloc(sizeof(struct dart_group_struct));
+  struct dart_group_struct* group = (struct dart_group_struct *) malloc(
```
devreal: Cast of `malloc` return type is not needed in ANSI C, see http://c-faq.com/malloc/cast.html

dart-impl/mpi/src/dart_team_group.c
```diff
@@ -373,7 +374,7 @@ dart_ret_t dart_group_locality_split(
   /* create a group for every domain in the specified scope: */
 
   int total_domains_units           = 0;
-  dart_domain_locality_t ** domains = malloc(
+  dart_domain_locality_t ** domains = (dart_domain_locality_t **) malloc(
```
devreal: See above.

dart-impl/mpi/src/dart_team_group.c
```diff
@@ -395,15 +396,19 @@ dart_ret_t dart_group_locality_split(
...
+  }
   if (num_groups == (size_t)num_domains) {
     /* one domain per group: */
     for (size_t g = 0; g < num_groups; ++g) {
       int                  group_num_units = domains[g]->num_units;
       dart_global_unit_t * unit_ids        = domains[g]->unit_ids;
 
       /* convert relative unit ids from domain to global unit ids: */
-      int * group_global_unit_ids = malloc(group_num_units * sizeof(int));
+      int * group_global_unit_ids = (int *) malloc(group_num_units * sizeof(int));
```
devreal: Oh, I think I know where this is leading to...

dart-impl/mpi/src/dart_team_group.c
```diff
@@ -480,9 +485,13 @@ dart_ret_t dart_group_locality_split(
       for (int d = group_first_dom_idx; d < group_last_dom_idx; ++d) {
         group_num_units += domains[d]->num_units;
       }
-      dart_global_unit_t * group_team_unit_ids =
-                              malloc(sizeof(dart_global_unit_t) *
-                                     group_num_units);
+
+      dart_global_unit_t * group_team_unit_ids = NULL;
+      if(group_num_units != 0){
```
devreal: Not necessary, man page states: 

> NULL may also be returned by a successful  call  to malloc() with a size of zero, or by a successful call to calloc() with nmemb or size equal to zero.

dart-impl/mpi/src/dart_team_private.c
```diff
@@ -285,11 +285,12 @@ dart_ret_t dart_allocate_shared_comm(dart_team_data_t *team_data)
...
     MPI_Comm_group(DART_COMM_WORLD, &group_all);
 
-    int * dart_unit_mapping = malloc(
+    int * dart_unit_mapping = (int *) malloc(
         team_data->sharedmem_nodesize * sizeof(int));
-    int * sharedmem_ranks = malloc(
+    int * sharedmem_ranks = (int *) malloc(
         team_data->sharedmem_nodesize * sizeof(int));
-    team_data->sharedmem_tab = malloc(size * sizeof(int));
+    team_data->sharedmem_tab = (dart_team_unit_t *) malloc(
```
devreal: gnarf.

devreal (changes requested):
There are some important fixes, thanks for that. However, some I disapprove of (casting return value of `malloc` in C and checking for zero being passed to `malloc`). There are some subtle differences between C and C++ and we should not mix them.

dart-impl/mpi/src/dart_mem.c
fmoessbauer: It's a dead increment: https://felixmoessbauer.de/temp/build-report/report-fa474c.html#EndPath

devreal: Ahh, sorry, didn't fully read you initial post. Will look into that.

dart-impl/mpi/src/dart_team_group.c
devreal: There is a bug filed against Clang static analyzer on this: https://bugs.llvm.org//show_bug.cgi?id=15708

dart-impl/mpi/include/dash/dart/mpi/dart_globmem_priv.h
fmoessbauer: It's used multiple times in `dart_synchronization.c`:

```
dart-impl/mpi/src/dart_synchronization.c:69:    DART_GPTR_COPY((*lock) -> gptr_tail, gptr_tail);
dart-impl/mpi/src/dart_synchronization.c:70:    DART_GPTR_COPY((*lock) -> gptr_list, gptr_list);
dart-impl/mpi/src/dart_synchronization.c:97:    DART_GPTR_COPY(gptr_tail, lock -> gptr_tail);
dart-impl/mpi/src/dart_synchronization.c:98:    DART_GPTR_COPY(gptr_list, lock -> gptr_list);
dart-impl/mpi/src/dart_synchronization.c:156:   DART_GPTR_COPY(gptr_tail, lock -> gptr_tail);
dart-impl/mpi/src/dart_synchronization.c:196:  DART_GPTR_COPY(gptr_tail, lock -> gptr_tail);
dart-impl/mpi/src/dart_synchronization.c:197:  DART_GPTR_COPY(gptr_list, lock -> gptr_list);
dart-impl/mpi/src/dart_synchronization.c:256:   DART_GPTR_COPY(gptr_tail, (*lock) -> gptr_tail);
dart-impl/mpi/src/dart_synchronization.c:257:   DART_GPTR_COPY(gptr_list, (*lock) -> gptr_list);
```

dart-impl/mpi/src/dart_team_group.c
fmoessbauer: I thought that this is not necessary as well, but `clang-analyze` does not agree on that (all build flags we use are passed): `Unix API Allocator sizeof operand mismatch`. Anyway: Does this hurt?  We should at least use the cast always or never.

dart-impl/mpi/src/dart_team_group.c
fmoessbauer: Yes, wonder how it got there.

dart-impl/mpi/src/dart_team_group.c
fmoessbauer: This is correct, but if and only if using an up-to-date unix which does not suffer from CERT MEM04-C and CWE-131:

`Undefined allocation of 0 bytes (CERT MEM04-C; CWE-131)`

fmoessbauer: However there is a mailing list entry on llvm that this error message is misleading: https://bugs.llvm.org//show_bug.cgi?id=15708

dart-impl/mpi/src/dart_team_group.c
rkowalewski: It is more a matter of style and you used probably the clang static analyzer with some kind of c++ configuration because c++ requires strictly to cast the result of malloc, while C does not. In C it is actually discouraged because you may suppress compiler warnings if, for example, you forgot to include `stdlib.h`.

devreal: > Unix API Allocator sizeof operand mismatch

I couldn't immediately find this warning in your _before_ analyzer output. However, I think what it refers to is a (potential) difference in the sizes of the type used in `sizeof` and the assigned type (such `dart_unit_t` vs `dart_global_unit_t`). So the cast actually hides that. 

> Does this hurt?

I think it definitely hurts readability as it clutters the code even more, unnecessarily so, and repeats the type three times in a single statement. On the positive side, it might trigger a warning if the target type changes but the cast does not. A clean solution could be:
```
struct dart_group_struct* group = malloc(sizeof(*group));
```
which would account for any changes in the type of `group`. 

I agree that we have to be consistent on this throughout DART. I have been removing the casts in pretty much all the code I touched but haven't thought about the solution outlined above, which I got from http://stackoverflow.com/questions/605845/do-i-cast-the-result-of-malloc

dart-impl/mpi/include/dash/dart/mpi/dart_globmem_priv.h
devreal: Not anymore :)

devreal:
I have plowed through some of the warnings and fixed a bunch of problems in DART. There  are about a dozen problems left in the locality code that I cannot immediately fix. @fuchsto Can you please have a look? I also added a `TODO` on missing conversion between team-relative and global unit IDs in `dart_group_locality_split`.

dart-impl/mpi/src/dart_team_group.c
fmoessbauer: @devreal : of course you are right ;) The problems are related to a mismatch of `dart_unit_t` and `dart_global_unit_t` and the cast hides them. `clang-analyze` perfectly accepts malloc without cast iff the types are equal.

dart-impl/mpi/src/dart_mem.c
devreal: This is strange: the second `if` below breaks the `for(;;;)` loop and thus the value _is_ read at the beginning of the enclosing `while` loop. Which Clang version do you use? I do not see this warning with 3.8.0.

fuchsto:
@fmoessbauer @devreal Yes, leave the changes in dart/locality to me. Will resolve this later today.

fuchsto:
@fmoessbauer The `build.analyze.sh` does not work with precompiled headers for me on projekt03 with clang-5.0, any hints?

Also, if I'm not mistaken the analyze-build only works for `CMAKE_C(XX)_COMPILER=clang(++)`, right?

fuchsto:
... also, the `REPORT_DIR` was not cleared and a newline was missing an escape.
Even then, I get:

~~~
...
-- Configuring done
-- Generating done
-- Build files have been written to: /home/fuchsto/workspaces/dash-bug-static-analysis/build.analyze
scan-build: Analysis run complete.
scan-build: Analysis results (plist files) deposited in '/home/fuchsto/workspaces/dash-bug-static-analysis/build.analyze/report/2017-02-27-153115-11966-1'
scan-build: Removing directory '/home/fuchsto/workspaces/dash-bug-static-analysis/build.analyze/report/2017-02-27-153115-11966-1' because it contains no reports.
scan-build: No bugs found.

   To build using these settings, hit ENTER
~~~

That does not seem right. Ideas?

fmoessbauer:
I did only test the script using clang 3.8 but it should also work when using gcc. The scan build script modifies CC and CXX, and sets it to clang-analyze. Then all compiler runs are intercepted and after that the real compiler is started.

As cmake hard-codes the compilers set in CC in the Makefile, scan-build has to be executed twice. The first output hence does not contain any reports. So in fact, the behavior is correct. Just hit enter.

Here is an updated version of the build script: If it works, just pull it. Branch: [`bug-static-analysis-prop`](https://github.com/dash-project/dash/compare/bug-static-analysis...bug-static-analysis-prop)

build.analyze.sh
```diff
@@ -74,9 +74,14 @@ fi
 mkdir -p $BUILD_DIR
 mkdir -p $REPORT_DIR
 rm -Rf $BUILD_DIR/*
+rm -Rf $REPORT_DIR/*
```
fmoessbauer: Not necessary as `$REPORT_DIR` is relative to `BUILD_DIR`. In fact, the mkdir statement is wrong: it should be `mkdir -p $BUILD_DIR/$REPORT_DIR`

build.analyze.sh
```diff
@@ -74,9 +74,14 @@ fi
...
 (cd $BUILD_DIR && $BUILD_WRAPPER $ANALYZE_OPTS \
                   cmake -DCMAKE_BUILD_TYPE=Release \
                         -DENVIRONMENT_TYPE=default \
+                        \
+                        -DCMAKE_C_COMPILER=clang \
+                        -DCMAKE_CXX_COMPILER=clang++ \
+                        \
```
fmoessbauer: That is not necessary, but propably this does not work with `gcc` due to the precompiled headers (which are only enabled in gcc builds).  Better disable `ENABLE_COMPTIME_RED`in this builds.

build.analyze.sh
fuchsto: Ah, that could fix both issues.

fmoessbauer:
Just tested with the new `build.analyse.sh` with gcc, g++ and clang from llvm5.0 and it works like charm.

fuchsto:
Can't quite believe it does.
The script in your other branch has:

~~~
mkdir -p $BUILD_DIR/$REPORT_DIR
rm -Rf $BUILD_DIR
~~~

Let me know how you make it run on projekt03, I didn't find a working environment setup.

fmoessbauer:
Oops, I committed a staged version without the latest changes. Of course this has to be the other way round. It's fixed now. However surprisingly this even works, as the report dir is automatically created by clang...

It worked without any special modification:
```
module list > 1) prun/1.1   2) ohpc   3) gnu/5.4.0   4) impi/2017.1   5) phdf5/1.8.17
SCANBUILD_BIN=scan-build ./build.analyze.sh
```

This does not link correctly due to the scalapack problem, but that has no impact on the static code analysis.

fuchsto:
@fmoessbauer Still no luck on projekt03 ...

Why is it reporting "Analysis run complete" before the actual build?

~~~

-- Generating done
-- Build files have been written to: /home/fuchst/workspaces/dash-bug-static-analysis/build.analyze
scan-build: Analysis run complete.
scan-build: Analysis results (plist files) deposited in '/home/fuchst/workspaces/dash-bug-static-analysis/build.analyze/report/2017-02-27-191029-22267-1'
scan-build: Removing directory '/home/fuchst/workspaces/dash-bug-static-analysis/build.analyze/report/2017-02-27-191029-22267-1' because it contains no reports.
scan-build: No bugs found.

   To build and analyze using these settings, hit ENTER
~~~

Also, I get compiler errors (using the same module environment setup you posted):

~~~
Array.h:645:22: error: no member named 'is_trivially_copyable' in namespace 'std'
  static_assert(std::is_trivially_copyable<ElementType>::value,
~~~

Which clang do you use on projekt03?

fuchsto:
@fmoessbauer I don't quite understand some of the reported nullptr-derefs, for example:

https://felixmoessbauer.de/temp/build-report-new/report-1bd6e8.html#EndPath

Any idea why it would report this?

fuchsto:
Another nuisance: Some versions of `scan-build` expect `--analyze-headers` and `--plist-html` (double-dash flags).

fuchsto:
Now works on my private server:

<http://62.75.245.182:8181>

fuchsto:
Aaaha. Apparently the `DART_ASSERT*` macros are not understood by `scan-build`.

Reordered checks in definitions of assertion macros to simplify evaluation for static analysis, seems to be fixed now.

fuchsto:
Also, documentation of `scan-build` strongly recommends to use debug builds:

<https://clang-analyzer.llvm.org/scan-build.html#recommended_debug>

Disabled test sources from analysis by default as they take most of analysis time but do not contribute to reported errors.

fuchsto:
Almost done, a few dubious reports are left, like:

~~~
dart-impl/base/src/internal/host_topology.c:690:27: warning: Call to 'realloc' has an allocation size of 0 bytes
~~~

Well, I know, because I assume behavior of `realloc` as documented in the C99 standard:

> if size is equal to zero, and ptr is not NULL, then the call is equivalent to free(ptr)

Not sure if there is an actual problem.

fmoessbauer:
@fuchsto : Now I see the problems on project03: The wrong STL implementation is used (e.g. 4.9.x). It seems that this depends on various environment variables and the order of the includes. This was just a matter of luck that my script ran successfull (it even did not compile, but the analysis worked) This seems to be a bug. This is another advantage of using the docker containers: You have an absolutely clean environment.

> Why is it reporting "Analysis run complete" before the actual build?

That is because build-scan is run twice as described above. Just ignore the first output.

> I don't quite understand some of the reported nullptr-derefs

clang-analyze marks every pointer possibly null, if the opposite cannot be proven. We could ignore that, add an assertion or [use annotations](https://clang-analyzer.llvm.org/faq.html).

Thank you for putting time into this. The static code analysis looks to me like a highly experimental field of application...

fuchsto:
@fmoessbauer Thank you, I think I got it figured out now. Trying to eliminate some remainng false positives, but wouldn't care too much if `scan-build` cannot be convinced to interpret them correctly.

@devreal @fmoessbauer 
I don't quite follow the reasoning behind the remaining reported issues.

Find the output of my analysis run here: 62.75.245.182:8181

dart-impl/mpi/include/dash/dart/mpi/dart_globmem_priv.h
fuchsto: @devreal @fmoessbauer 

I can't make up a "mechanical" reason for this macro, either. It would only make sense if `dart_gptr_t` contained dynamically allocated fields, and that again would be ... preposterous.
Presumably it has been added to round up the `DART_GPTR_*` macro interface consisting of:

`DART_GPTR_NULL`, `DART_GPTR_IS_NULL`, `DART_GPTR_EQUAL`, `DART_GPTR_SET_FLAG`, `DART_GPTR_GET_FLAG`.

... while all of these should be inline functions. Macros don't buy anything here apart from nicely obfuscated compiler errors.

fuchsto:
@devreal @fmoessbauer Travis has some issues, please review my recent changes anyways.

dart-impl/mpi/src/dart_team_group.c
```diff
@@ -482,41 +482,35 @@ dart_ret_t dart_group_locality_split(
...
+        gout[g] = NULL;
         continue;
       }
       int group_unit_idx = 0;
       for (int d = group_first_dom_idx; d < group_last_dom_idx; ++d) {
-        for (int u = 0; u < domains[d]->num_units; ++u) {
-          group_team_unit_ids[group_unit_idx + u] = domains[d]->unit_ids[u];
+        for (int du = 0; du < domains[d]->num_units; ++du) {
+          int u = group_unit_idx + du;
+          group_unit_ids[group_unit_idx + u] = domains[d]->unit_ids[du].id;
```
devreal: `group_unit_idx` is used twice here. Do you mean:

```
int u = group_unit_idx + du;
group_unit_ids[u] = domains[d]->unit_ids[du].id;
```
?

dart-impl/base/src/internal/host_topology.c
```diff
@@ -691,10 +691,17 @@ dart_ret_t dart__base__host_topology__create(
...
+      // Note: realloc with zero-size is argued unsafe in certain scenarios:
+      // https://www.securecoding.cert.org/confluence/display/c/\
+      //   MEM04-C.+Beware+of+zero-length+allocations
+      if (host_units->num_units > 0) {
+        host_units->units = realloc(host_units->units,
+                                    host_units->num_units *
+                                      sizeof(dart_global_unit_t));
+        DART_ASSERT(host_units->units != NULL);
+      } else {
+        free(host_units->units);
```
devreal, line 703: Please `NULL` `host_units->units` to avoid double `free`s

dart-impl/mpi/src/dart_team_group.c
fuchsto: Indeed! Good catch!

dart-impl/mpi/include/dash/dart/mpi/dart_globmem_priv.h
devreal: 1. `DART_GPTR_NULL` cannot be an inline function since it is a constant. 
2. `DART_GPTR_EQUAL` and `DART_GPTR_IS_NULL` can be.
3. `DART_GPTR_SET_FLAG` and `DART_GPTR_GET_FLAG`: They only exist in the 4.0 interface and are outdated (there is `dart_gptr_getflags` and `dart_gptr_setflags`) to handle 16 and 8 bit flags in the segment and the `gptr`, respectively. We really need to talk about the flags and how they are accessed/handled.

devreal (approved):

PR #304, opened 2017-02-28
fixed ambiguous ctor of BlockPattern

fmoessbauer:
removed ambiguous constructors in BlockPattern and BlockPattern1D


fuchsto (changes requested):
The issue is not about an "ambiguous constructor" in BlockPattern, it's a missing `explicit` in the constructor of `TeamSpec`.

Both constructor variants for BlockPattern are required:

~~~c++
// A:
BlockPattern(sizespec, distspec, team = ALL);
// B:
BlockPattern(sizespec, distspec, teamspec, team = ALL);
~~~

Now, the constructor `TeamSpec(Team)` is not marked explicit. That's why a call of c'tor overload A matches B:

~~~c++
BlockPattern(sizespec, distspec, team)
// -> implicit conversion team -> teamspec and last param has default value:
BlockPattern(sizespec, distspec, TeamSpec(team), team = ALL)
~~~

See c'tors here:

- <https://github.com/dash-project/dash/blob/development/dash/include/dash/TeamSpec.h#L49>
- <https://github.com/dash-project/dash/blob/development/dash/include/dash/TeamSpec.h#L142>
- <https://github.com/dash-project/dash/blob/development/dash/include/dash/TeamSpec.h#L212>

fmoessbauer:
But the Ctors are ambiguous anyway, as both provide defaults for all except the first parameter. If it's not necessary we can close this PR. I just noticed the problem while trying to instantiate a BlockPattern using just SizeSpec and there, clang reported ambiguous ctor.

Here was the problem: https://github.com/dash-project/dash/blob/feat-coarray/dash/include/dash/Co_array.h#L98 . I'm going to continue my work on the co_array on Thursday so this fix is not mission critical at all.

fuchsto:
Nonono, the PR itself is fine.
I overlooked that both c'tors have a common ... param-list-prefix.
Still, the missing explicits are bugging me way more.

Just push it and I'll approve it quickly before someonce notices.

fmoessbauer:
Unfortunately this conversion is already used, so adding `explicit` will generate a chain of other fixes:

```c++
/opt/dash/dash/include/dash/Array.h:790:5: error: no matching constructor for initialization of 'dash::BlockPattern<1, dash::MemArrange::ROW_MAJOR, long>'
    m_pattern(
    ^
/opt/dash/dash/include/dash/Array.h:809:5: note: in instantiation of member function 'dash::Array<int, long, dash::BlockPattern<1, dash::MemArrange::ROW_MAJOR, long> >::Array' requested here
  : Array(nelem, dash::BLOCKED, team)
    ^
/opt/dash/dash/examples/ex.09.view-1dim/main.cpp:40:11: note: in instantiation of member function 'dash::Array<int, long, dash::BlockPattern<1, dash::MemArrange::ROW_MAJOR, long> >::Array' requested here
  Array_t array(nunits * block_size);
          ^
/opt/dash/dash/include/dash/pattern/BlockPattern1D.h:159:13: note: candidate constructor [with Args = <const dash::DistributionSpec<1> &, dash::Team &>] not viable: no known conversion from 'SizeSpec_t'
      (aka 'SizeSpec<1, unsigned long>') to 'SizeType' (aka 'unsigned long') for 1st argument
  constexpr BlockPattern(
            ^
/opt/dash/dash/include/dash/pattern/BlockPattern1D.h:221:3: note: candidate constructor not viable: no known conversion from 'dash::Team' to 'const TeamSpec_t' (aka 'const TeamSpec<NumDimensions, long>')
      for 3rd argument
  BlockPattern(
```

fuchsto:
Hm, we just could not provide a default for the `team` parameter in the extended constructor?

fmoessbauer:
But if we leave both constructors they are ambiguous even if we remove the default of the second parameter. Anyway, by prohibiting the conversion, the current dash code will not work anymore.

fuchsto:
Sorry, I just remembered that I defined these constructors for implicit conversion intentionally to support use cases exactly like these.
A conversion from `Team` to `TeamSpec` is always valid. With the implicit conversion, functions with a `TeamSpec` parameter are more flexible in invocation as it can be specified explicitly or initialized from  a team.

**However** this is a problem with signatures like these:

~~~c++
void foo(TeamSpec ts, team = ALL);
~~~

Because:

~~~c++
// assume nunits = 24
auto & myteam = dash::Team::All().split(2);

foo(TeamSpec<2>(2,6));
// same as:
foo(TeamSpec<2>(2,6), Team::All()); // team spec specified for different team 
~~~

devreal:
> Sorry, I just remembered that I defined these constructors for implicit conversion intentionally to support use cases exactly like these.

We should document this deliberate choice. 

It seems counter-intuitive having to pass a `TeamSpec` and a `Team` to a function, esp. if it allows me to pass different teams. Why not have the `TeamSpec` expose the `Team` used to create it and draw the `Team` from there?

fuchsto:
No, I added a specific c'tor in TeamSpec for this case:
https://github.com/dash-project/dash/blob/development/dash/include/dash/pattern/BlockPattern.h#L261

TeamSpec is a CartesianIndexSpace, it does not depend on `Team`.
That is to say: it can be initialized from extents only, without specifying a team, so it can't provide a `Team` getter.

(initializer list uses Team::All() to initialize `_myid` but this will be removed in the next clean up).

devreal:
So I still can do the following: 
```
auto pattern = BlockPattern(
                 SizeSpec(), 
                 DistributionSpec(), 
                 TeamSpec(some_distribution, Team::All().split(2)));
```
And the team in the `TeamSpec` of BlockPattern will be overriden by dash::Team::All? That sounds like it's made to be confusing...

fuchsto:
Yes, that's the use case I meant in my comment. It's just not a robust interface like that.

fuchsto:
No, that's fine:

https://github.com/dash-project/dash/blob/development/dash/include/dash/pattern/BlockPattern.h#L261

Again, TeamSpec does not depend on a team. The team is only used for the first configuration of the progress grid extents.

devreal:
You end up with extents computed for the split team (they are taken from the passed `TeamSpec` in the `TeamSpec` copy-ctor) and the unit ID from `Team::All()`. Is that valid? Are the extents completely overriden in the copy-ctor? What role does the team play that is passed to `TeamSpec` then?

fuchsto:
That's what is *passed* to the constructor, but the TeamSpec member in BlockPattern is initialized with both: The Team and the TeamSpec.
From the code location I linked,

~~~c++
 BlockPattern(
    const SizeSpec_t         & sizespec,
    const DistributionSpec_t & dist     = DistributionSpec_t(),
    const TeamSpec_t         & teamspec = TeamSpec_t::TeamSpec(),
    dash::Team               & team     = dash::Team::All())
  : _distspec(dist),
    _team(&team),
    _teamspec(  // <==
      teamspec,
      _distspec,
      *_team),
// ...
~~~

devreal:
OK, so I traced it down and I see that the initial extents in `TeamSpec` are overridden by the team specified in the copy ctor (which is the defaulted `Team::All`). I stand my ground on the claim that the fact that a posterior defaulted argument overrides the team passed to an earlier argument of the `BlockPattern` ctor is not intuitive. I understand that the `TeamSpec` should not contain the team but hey, I announced the intended team already (in the `TeamSpec`) so why bother announcing it again? This inconsistency should be fixed.

fuchsto:
Which copy constructor do you mean?
The extents specified by the caller are not overridden. BlockPattern ends up here:

<https://github.com/dash-project/dash/blob/development/dash/include/dash/TeamSpec.h#L104>

But perhaps we are talking about different cases?

About the repeated specification of the team:
Well, you *may* have specified the team already, but TeamSpec can be initialized without specifying a team as well. 
In initialization of TeamSpec with a team, only numeric the attributes (size and gaps mostly) of the team are considered.
TeamSpec is not about "wrapping a team with specs" but "providing hints for topological arrangement of an aribtrary team":

~~~c++
Pattern(
  sizespec,
  distspec,
  TeamSpec<2>(34,45),   // <-- "use this as a hint on how to arrange
  team);                // <--  this team, if possible; otherwise rebalance
                        // or make it work somehow"
~~~

With that in mind, it should not be too irritating that the team can be specified twice, and differently, even.

I'm definitely not satisfied with the interface as it is, but I ... inherited 🧀  use cases that had to be supported.

fmoessbauer:
Can we merge this PR or is there still something to discuss and / or to be changed?

dash/include/dash/pattern/BlockPattern.h
```diff
@@ -253,7 +253,7 @@ class BlockPattern
     /// dimensions
     const DistributionSpec_t & dist     = DistributionSpec_t(),
     /// Cartesian arrangement of units within the team
-    const TeamSpec_t         & teamspec = TeamSpec_t::TeamSpec(),
+    const TeamSpec_t         & teamspec = typename TeamSpec_t::TeamSpec(),
```
fuchsto: `typename` is wrong here, remove it

dash/include/dash/pattern/BlockPattern1D.h
```diff
@@ -225,7 +225,7 @@ class BlockPattern<1, Arrangement, IndexType>
     /// Defaults to BLOCKED.
     const DistributionSpec_t dist     = DistributionSpec_t(),
     /// Cartesian arrangement of units within the team
-    const TeamSpec_t         teamspec = TeamSpec_t::TeamSpec(),
+    const TeamSpec_t         teamspec = typename TeamSpec_t::TeamSpec(),
```
fuchsto: same here, `typename` is wrong

fuchsto (changes requested):

fuchsto (approved):

PR #305, opened 2017-02-28
fixed circular includes in matrix classes

fmoessbauer:
removed circular includes of `dash/Container.h` and updated missing includes

### DIFF
--- a/dash/include/dash/matrix/MatrixRef.h
+++ b/dash/include/dash/matrix/MatrixRef.h
@@ -9,7 +9,9 @@
 #include <dash/Pattern.h>
 #include <dash/GlobRef.h>
 #include <dash/HView.h>
-#include <dash/Container.h>
+
+#include <dash/view/Local.h>
+#include <dash/view/Sub.h>
 
 #include <dash/iterator/GlobIter.h>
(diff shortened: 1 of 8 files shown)


dash/include/dash/matrix/MatrixRef.h
```diff
@@ -9,7 +9,9 @@
 #include <dash/Pattern.h>
 #include <dash/GlobRef.h>
 #include <dash/HView.h>
-#include <dash/Container.h>
+
+#include <dash/view/Local.h>
```
fuchsto: Are these required? The implementation of `dash::Matrix` does not depend on view expressions, does it?

fmoessbauer: Unfortuately there are many many cross dependencies and the easiest way was to separate the includes to specific files. And yes, this is required.

fuchsto: What for? Which definition would be missing otherwise?

fmoessbauer: Just `dash/Sub.h` is sufficient, as `local` and `sub` are used.

devreal (approved):
Looks fine to me

PR #306, opened 2017-02-28
Fixed type conflict of dart_buddy_alloc on 32bit machines

ddiefenthaler:
Reflected type change of dart_buddy_alloc in the header from commit 500abc9

### DIFF
--- a/dart-impl/mpi/include/dash/dart/mpi/dart_mem.h
+++ b/dart-impl/mpi/include/dash/dart/mpi/dart_mem.h
@@ -42,7 +42,7 @@ void     dart_buddy_delete(struct dart_buddy *);
  * \return The offset relative to the starting adddress of the external
  *         memory block where the allocated memory begins.
  */
-uint64_t dart_buddy_alloc(struct dart_buddy *, size_t size);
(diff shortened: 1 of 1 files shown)


fuchsto:
The `dart_buddy_alloc` interface is dubios in general.
<https://github.com/dash-project/dash/blob/development/dart-impl/mpi/src/dart_mem.c#L128>

So in the C source, the signature is `size_t dart_buddy_alloc(...)` but the implementation returns `-1` on error (without a cast to indicate that this is intentional).

Then, failure of `dart_buddy_alloc` is tested accordingly:
<https://github.com/dash-project/dash/blob/development/dart-impl/mpi/src/dart_globmem.c#L159>

Semantically, `dart_buddy_alloc` is supposed to return an offset:
<https://github.com/dash-project/dash/blob/development/dart-impl/mpi/src/dart_mem.c#L141>

... and offsets are signed. Always. It's the law.

Oh, and there also is `dart_buddy_new`, for good measure:
<https://github.com/dash-project/dash/blob/development/dart-impl/mpi/src/dart_mem.c#L58>

It doesn't do any harm at the moment, but feel free to clean up the interface.

fuchsto (approved):

PR #308, opened 2017-03-01
Re-use segment ID of free'd segments

### LINKED ISSUE #221: Segment IDs and dynamic global data structures

devreal:
Fixes #221

### DIFF
--- a/dart-impl/mpi/include/dash/dart/mpi/dart_segment.h
+++ b/dart-impl/mpi/include/dash/dart/mpi/dart_segment.h
@@ -30,34 +30,64 @@ typedef struct
 typedef struct dart_seghash_elem dart_seghash_elem_t;
 
 typedef struct {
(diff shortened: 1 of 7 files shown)


fuchsto (approved):
Good stuff!

PR #309, opened 2017-03-01
Fix test output

devreal:
`AtomicInterface:AtomicTest` => `AtomicTest:AtomicInterface` and invalid number of units


fmoessbauer (approved):

fmoessbauer (commented):
TestGroup.Test would be even better as it reflects googles test naming scheme.

fmoessbauer (approved):
Perfect ;)

PR #310, opened 2017-03-02
[WIP] Finish Global Dynamic Memory and Allocator Concepts

fuchsto:
Brief (as possible) summary of the concept definitions that emerged from my GDM evaluation for discussion:

Memory Space and Allocator Concepts
=============================

Memory Space
------------

The principle task of memory spaces is to maintain a global address space.
This consequently implies the following aspects:

-	**Pointer Semantics**

	A memory space provides a dependent pointer type that realizes random
	access semantics within its address range. Memory space pointers have
	two principle concerns:

	- Random access iteration (conventional pointer arithmetics) in
	  the address range of their memory space
	- Access to values in the memory space (dereference)
	- Conversion to compatible pointers, e.g. between local and
	  global pointers

	Pointers depend on a memory space instance to provide these operations.
	For global address space, pointer types must be defined as
	specializations of `dash::GlobPtr` and specified in
	`dash::memory_space_traits` as dependent type:

	```c++
	using MemSpaceType<T, ...>::pointer = GlobPtr<Tc, MemSpaceType<T, ...>>`
	```

	The compatible operations for pointer types of two different memory
	spaces is specified by the pointer's implementation.

-   **Resource Interface for Allocators**

	Allocators depend on memory spaces to attach local memory in the global
	address space. In this, memory space types might support access from
	multiple allocator instances on the same memory space.
	The memory space concept therefore must provide methods to query total
	and available capacity.
	For example, a static memory space is initialized with a fixed capacity
	which might be shared by multiple allocators. Dynamic memory spaces
	represent a global heap that automatically grows and shrinks on
	allocation and de-allocation.

-	**Synchronization of Distributed Memory Subspaces**

	Changes on a memory segment usually are not visible at remote locations
	immediately. Dynamic memory spaces may, among other aspects, differ in
	the communication of memory segment sizes between units (collective epoch
	synchronization, deferred point-to-point synchronization, ...) or
	data modifications.
	For example, a memory space could issue asynchronous pointers that
	perform an implicit flush operation (= future) when dereferenced at a
	remote address.

Allocator
---------

-	**Allocation Balance**

	The number of elements allocated at the single units depends on the
	allocator, not the memory space type.
	More specific, allocator semantics define whether `AllocType.allocate(n)`
	will reserve `n` elements at every unit (balanced) or `n` elements at
	the active unit (unbalanced).
	An allocator's memory space might support irregular (unbalanced) address
	ranges but a balanced allocator may still perform balanced allocation
	on it.

-	**Unit Participation in Allocation**

	Allocator semantics define whether a call of `AllocType.allocate` is
	a one-sided, collective, or partially collective operation.
	Allocators are not responsible for communicating allocated memory
	regions across units, however.
	For example, a call of `AllocType.allocate(l)` could be unbalanced
	(local allocation with `l` varying between units) and collective.
	The propagation of the local sizes `l` between units relies to the
	memory space.

Traits
------

Not all combinations of allocator- and memory space types are compatible,
of course. Compatibility can be checked at compile time by matching allocator
traits and memory space traits, just like already implemented in pattern traits.

The same principles used to deduce pattern types from container- and algorithm
constraints can applied to deduce allocator types for memory spaces, or vice
versa, or depending on algorithm execution- and launch policies, etc.

The aspects listed above represent the primary dimensions (= categories) of allocator
and memory space traits.

Interaction
-----------

From application perspective, memory can only be acquired from an allocator
instance that is initialized with an underlying memory space.
Memory spaces can again utilize allocators for memory management of their
sub-spaces. A general call dependency is:

~~~
AllocOuterT
  .allocate(n)
    -> MemSpaceOuterT
         .allocate(n)
           - mso_add_capacity = n - free_capacity
           - mso_bucket =
               grow(mso_add_capacity) 
               -> AllocInnerT.allocate(mso_add_capacity)
                  // ... arbitrarly nested
               <- msi_pointer
           - attach(msi_pointer)
           - capacity += mso_add_capacity
             (might commit attached memory here)
         <- mso_bucket
    <- mso_pointer = mso_bucket.begin()
  <- mso_pointer
~~~


fuchsto:
@rkowalewski In the current state of the implementation of scoped allocators, the principle of how local allocation is delegated presumably works, but there is no concept of affine address space sizes.
For example, containers require `MemorySpace:;lbegin / lend` and `MemorySpace::local_size(unit)` as in the `GlobDynamicMem` concept in development branch.
As this is related to global heaps / epoch memory spaces etc. anyways, I will push my proposal to PR #333

dart-if/include/dash/dart/if/dart_globmem.h
```diff
@@ -90,8 +113,20 @@ typedef struct
...
+
+  bool operator==(const dart_gptr &rhs) const noexcept
+  {
+    return DART_GPTR_EQUAL((*this), rhs);
+  }
+  bool operator!=(const dart_gptr &rhs) const noexcept
+  {
+    return !(*this == rhs);
+  }
+#endif
```
devreal: I'm not happy with having additional C++ code in DART. Maybe implement these operators somewhere in DASH, potentially in a header wrapping `dart_globmem.h `.

dash/include/dash/allocator/internal/Types.h
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  memory_block(void *ptr, size_t length) noexcept
+    : ptr(ptr)
+    , length(length)
+  {
+  }
+
+  // Move Constructor
+  memory_block(memory_block &&x) noexcept { *this = std::move(x); }
```
devreal, line 25: Shouldn't this `.reset()` `x`?

dash/include/dash/map/UnorderedMap.h
```diff
@@ -643,6 +611,68 @@ class UnorderedMap
     return result;
   }
 
+  iterator insert(
+    const_iterator hint,
+    const value_type & value)
+  {
+    Timer::timestamp_t ts_enter = Timer::Now(), ts_insert, ts_find, d_insert, d_find;
```
devreal: The timestamps are only needed for debugging. They should not be used in builds without debug output.

dash/include/dash/map/UnorderedMap.h
```diff
@@ -792,12 +822,19 @@ class UnorderedMap
                    "unit:",   unit,
                    "key:",    value.first);
     auto result = std::make_pair(_end, false);
+    /* rkowalewski:
+     * Why do we increment local size and _local_cumul_size for the corresponding unit at the same time?
+     * This seems strange to me!!
+    */
```
devreal: Is this still an issue?

dash/include/dash/memory/GlobHeapMem.h
```diff
@@ -1118,77 +1132,108 @@ class GlobHeapMem
...
-    dash::copy(_num_attach_buckets.begin(),
-               _num_attach_buckets.end(),
+    dash::copy(_num_attach_buckets.begin(), _num_attach_buckets.end(),
                num_unattached_buckets.data());
-    // Attach array of local unattached bucket sizes to allow remote units to
-    // query the sizes of this unit's unattached buckets.
+
+#ifdef DASH_ENABLE_TRACE_LOGGING
+    std::for_each(std::begin(_num_attach_buckets),
+                  std::end(_num_attach_buckets),
```
devreal, line 1140: Iterating over `num_unattached_buckets` may be more efficient.

dash/include/dash/memory/MemorySpace.h
```diff
@@ -1,18 +1,268 @@
...
+ *
+ * Clarify:
+ * The STL memory_resource concept assumes that any type can be
+ * allocated with its size specified in number of bytes so untyped
+ * allocation with void pointers is practicable.
+ * However, the DART allocation routines are typed for correctness
+ * considerations and to optimize communication.
+ * If the memory space concept complies to the STL memory_resource
+ * interface, only DART_TYPE_BYTE instead of the actual value type
+ * is available. This would therefore harm stability and performance.
```
devreal, line 149: Note: the type argument in DART allocation functions is not used and you can safely pass `DART_TYPE_BYTE` to them. It do not affect alignment, which will be taken care of later (DART memory allocation overhaul). In fact, passing the type to the underlying MPI implementation only affects the indexing in access functions (which is in bytes for us, for reasons).

dash/test/allocator/EpochSynchronizedAllocatorTest.cc
```diff
@@ -0,0 +1,58 @@
...
+  GlobDynAllocTraits::pointer gp = GlobDynAllocTraits::allocate(dynAlloc, n);
+  //Each process allocates all the local elements
+  GlobDynAllocTraits::deallocate(dynAlloc, gp, n);
+
+  using OtherDynAllocTraits = GlobDynAllocTraits::rebind_traits<struct my_type>;
+
+  OtherDynAllocTraits::allocator_type otherDynAlloc{dash::Team::All()};
+  OtherDynAllocTraits::pointer gp2 = OtherDynAllocTraits::allocate(otherDynAlloc, n);
+  OtherDynAllocTraits::deallocate(otherDynAlloc, gp2, n);
+}
```
devreal: Missing newline.

dash/test/allocator/EpochSynchronizedAllocatorTest.h
```diff
@@ -0,0 +1,12 @@
...
+
+#include "../TestBase.h"
+
+/**
+ * Test fixture for class dash::GlobDynamicMem
+ */
+class EpochSynchronizedAllocatorTest : public dash::test::TestBase {
+};
+
+#endif
```
devreal: No newline.

dash/test/memory/SimpleMemoryPoolTest.h
```diff
@@ -0,0 +1,96 @@
...
+#define DASH__TEST__UNIT_ID_TEST_H__INCLUDED
+
+#include "../TestBase.h"
+
+#include <dash/memory/SimpleMemoryPool.h>
+#include <dash/Exception.h>
+
+/**
+ * Test fixture for class DASH unit id types.
+ */
```
devreal: That comment seems strange ;)

devreal (changes requested):
Since I am on my way to the airport, I don't have time to fully grasp all the details. Comments inline, good otherwise :+1:

dash/include/dash/map/UnorderedMap.h
rkowalewski: no it is not. will remote these comments. :)

dash/include/dash/map/UnorderedMap.h
rkowalewski: This was actually a commit by accident. Will remove this either. It is something which does not really belong to this pull request.

dash/include/dash/map/UnorderedMap.h
```diff
@@ -643,6 +611,68 @@ class UnorderedMap
...
+    if (found != _end) {
+      DASH_LOG_TRACE("UnorderedMap.insert", "key found");
+      // Existing element found, no insertion:
+      res = found;
+    } else {
+      DASH_LOG_TRACE("UnorderedMap.insert", "key not found");
+      // Unit mapped to the new element's key by the hash function:
+      DASH_LOG_TRACE("UnorderedMap.insert", "target unit:", unit);
+      // No element with specified key exists, insert new value.
+      ts_insert = Timer::Now();
```
fuchsto: Benchmarking / debugging leftovers? If these changes are intended to improve performance, we should add a benchmark example.

fmoessbauer:
Update: Not necessary anymore, as already fixed by @devreal 

While implementing this, please also fix the `GlobRef<T>.member()` accessor. This is broken due to the recent changes in `GlobPtr`. I will make an issue for this after I am back again.

dash/include/dash/allocator/internal/Types.h
fmoessbauer: This looks quite strange. If this works, better use `memory_block(memory_block &&) noexcept = default;`

dash/include/dash/allocator/internal/Types.h
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Bool operator to make the Allocator code better readable
+   */
+  explicit operator bool() const noexcept
+  {
+    return length != 0 && ptr != nullptr;
+  }
+
+  void *ptr;
```
fmoessbauer, line 62: should be encapsulated in `std::unique_ptr`, as otherwise the ptr is not set to `nullptr` on default move.

PR #311, opened 2017-03-02
Feat patblockviz tile

devreal:
Add support for `TilePattern` to pattern-block-visualizer and add proper help output

### DIFF
--- a/dash/examples/ex.06.pattern-block-visualizer/main.cpp
+++ b/dash/examples/ex.06.pattern-block-visualizer/main.cpp
@@ -189,15 +189,15 @@ make_block_pattern(
 {
   // Example: -n 30 30 -u 4 1 -t 10 10
   typedef dash::Pattern<2> pattern_t;
-  dash::Pattern<2> pattern(sizespec,
-                           dash::DistributionSpec<2>(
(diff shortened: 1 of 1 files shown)


fuchsto (approved):

ddiefenthaler (approved):
Looks good,
I'll drag my own changes along soon.

dash/examples/ex.06.pattern-block-visualizer/main.cpp
```diff
@@ -240,11 +301,16 @@ cli_params parse_args(int argc, char * argv[])
...
+    } else {
+      print_usage(argv);
+      exit(EXIT_FAILURE);
     }
   }
+
   return params;
 }
 
+static
```
fuchsto, line 313: Just wondering: why the statics? These are functions in main.cpp, I don't see how external linkage would hurt?

devreal: I'd put it the other way around: there is really no sense in exposing anything but `main` in `main.cpp` here so there is no reason not have them `static`.  I think it's good practice to only expose those functions that actually need external linkage.

PR #315, opened 2017-03-07
Introduce Shared Libraries and Generic Build

### LINKED ISSUE #265: Portable Dash Version
The current version of dash use the compiler flags  "-march=native -mtune=native". These are hidden in the sub-CMake files [config/coreglitch.cmake ](https://github.com/dash-project/dash/blob/development/config/coreglitch.cmake#L12) and [config/default.cmake](https://github.com/dash-project/dash/blob/development/config/default.cmake#L8). Without editing these files it is not possible to build a portable dash version. It would be nice if there is a cmake flag to enable a cross compiled version. 
At the moment it is pretty hard to see, that dash is build for a specific hardware. Even in ccmake there isn't a hint.

devreal:
Introduces the following two CMake options:
1. `BUILD_GENERIC` to disable architecture specific optimizations (`-mtune=native` and `-march-native`)
2. `BUILD_SHARED_LIBS` to build shared libraries (static libraries are still the default)

Should we include shared libraries in the tests? If so, where? When using shared libraries, the test script has to set the `LD_LIBRARY_PATH`.

UPDATE: I added support for limited visibility of internal symbols, e.g,, `dart_segment*` and global variables. See https://gcc.gnu.org/wiki/Visibility for details. 

Fixes #265


fuchsto (approved):
That's a clean solution, perfect!

Issue #109, opened 2016-11-15
Thread-safety of DART functions


devreal:
The documentation states the following:

> A note on thread safety:
> All functions in the DART interface shall be implemented in a thread safe way. 

This is problematic for two reasons:
- We cannot guarantee that the underlying communication library supports multi-threaded access (MPI provides `MPI_Init_thread` to check whether this is supported) neither do we enforce multi-threaded initialization at the moment. 
- There are global data-structures in DART that are manipulated, e.g., the team or segment data structures. We could guard them with mutexes by default but that will likely cause some overhead even for single-threaded applications. 

I propose to remove this sentence for now until we have support for tasking/multi-threading in DART in a later version. I would favor a relaxed policy, e.g., all communication operations should be thread-safe. 

Related to #106

fuchsto:
Thread-safe by default _without any restriction on safe operations whatsoever_ is a bad idea, that's why the STL doesn't do it.
But I think the issue here is just the missing definition of thread safety.

> All functions in the DART interface shall be implemented [...] thread safe

... does not specify if this refers to function calls on shared resources.
Calling `dart_group_add_member` in parallel is perfectly thread safe! But not on the same `dart_group_t` object.

I think that is what this spec is about. There must not exist any hidden shared states (global vars etc.)
that lead to data races the caller has no chance knowing about.

This also would correspond to the STL's understanding of thread safety in containers.
Concurrent reads must be safe, and concurrent writes with exactly one thread modifying the state.

Here's the meat:

> An object is thread-safe for reading from multiple threads. For example, given an object A, it is safe to > read A from thread 1 and from thread 2 simultaneously.
> If an object is being written to by one thread, then all reads and writes to that object on the same or
> other threads must be protected. For example, given an object A, if thread 1 is writing to A, then
> thread  2 must be prevented from reading from or writing to A.
> It is safe to read and write to one instance of a type even if another thread is reading or writing to a
> different instance of the same type. For example, given objects A and B of the same type, it is safe
> when A is being written in thread 1 and B is being read in thread 2.

Note that the second part

> If an object is being written to by one thread, then all reads and writes to that object on the same or
> other threads must be protected. 

... refers to the responsibility of the _programmer_, not the STL.

fuchsto:
@fuerlinger @knuedd @colinglass @jgraciahlrs
What is the stance of the members of the honorable DART Interface Review Board (DIRB) on this?

fuerlinger:
[...]

PR #316, opened 2017-03-07
Introduce Log-levels in DART (2)

devreal:
Giving this another shot (refer to #247 for previous discussion). This PR adds the following on top of it:

1. Resolve `DART_LOG*` functions to a single function call (`printf`-style argument checking is preserved). This yields some significant savings in binary size (151kB vs 240kB for `libdart-mpi.a` with debug output enabled)
2. Ensure mutual exclusion between threads to avoid interleaved log output. 
3. A bunch of fixes to format specifier.
4. Do not silently drop overlong log messages but print truncate line. It's up to the developer to fix them.

Related to #109


dart-impl/base/src/logging.c
```diff
@@ -22,3 +39,84 @@ const int dart__base__unit_term_colors[DART_LOG_TCOL_NUM_CODES-1] = {
...
+  char      msg_buf[maxlen];
+  pid_t     pid = getpid();
+  sn_ret = vsnprintf(msg_buf, maxlen, format, argp);
+//  if (sn_ret < 0 || sn_ret >= maxlen) {
+//    break;
+//  }
+  dart_global_unit_t unit_id;
+  dart_myid(&unit_id);
+  // avoid inter-thread log interference
+  dart_mutex_lock(&logmutex);
```
fuchsto: Acquiring a mutex for every log message has significant overhead, but more important, it also leads to implicit synchronization. This is only needed for multi-threaded logging, right? In this case, I suggest to wrap the `dart_mutex_*` calls in `#ifdef DART_ENABLE_THREADSUPPORT`.

dart-impl/base/src/logging.c
```diff
@@ -22,3 +39,84 @@ const int dart__base__unit_term_colors[DART_LOG_TCOL_NUM_CODES-1] = {
...
+  fprintf(DART_LOG_OUTPUT_TARGET,
+    "[ %*d %.5s ] [ %*d ] %-*s:%-*d %.3s DART: %s\n",
+    DASH__DART_LOGGING__UNIT__WIDTH, unit_id.id,
+    loglevel_names[level],
+    DASH__DART_LOGGING__PROC__WIDTH, pid,
```
fuchsto: These macros could be renamed to `DART_LOGGING__PROC_WIDTH` etc., I used the prefix `DASH__` back then but it's pretty stupid to write `DASH__DART...` in hindsight. Spelled out, it's `DASH DASH Runtime` ;)

fuchsto (changes requested):

dart-impl/base/src/logging.c
devreal: This is actually handled in `dart_mutex_*`, which resolves to an empty `static inline` call if `DART_ENABLE_THREADSUPPORT` is not defined.

fuchsto: That's what I assumed, but `dart_mutex_*` is part of the DART interface, the "empty" implementation is just in DART-MPI. So any other implementation of `dart_mutex_*`, e.g. in `DART-SHMEM` or `DART-GASPI` would depend on this assumption, too. (Stichwort Schichtenbruch).

dart-impl/base/src/logging.c
devreal: Done. Also, there is no need to have it in the header file anymore.

dart-impl/base/src/logging.c
devreal: `dart_mutex_*` is actually part of the `dart_base` interface, not the public DART interface. Thus, it should be safe to use in other modules (`shmem`, `tasking`, ...)

fuchsto: Well, yes, but that adds to my point. That is: we need the `#ifdef` guards in both places, the logging functions and the implementation of `dart_mutex_*`.

devreal: No it doesn't. Both `logging.h` and `mutex.h` are part of the `dart-base` interface. `logging.c` is compiled as part of `dart-base`. The modules that use the `DART_LOG*` macros will never even see the `dart_mutex_*` calls in `dart__logging__message`. So if there is no support for for threads, `logging.c` will be compiled with an empty version to `dart_mutex_*` and anything linking against `dart_base` will call that compiled function. Of course, if you decide to build `dart_base` and `dart_shmem` with different threading settings other things are likely to break...

fuchsto: Ok, I see. But shouldn't `dart_mutex_*` be renamed to `dart_base_mutex_*` then? Not sure how I did this with the locality stuff in dart/base, though.

fuchsto: Yes, I used `dart__base__locality*` for the common functions in dart/base. We should do the same for `dart_mutex*`.

devreal: It's now `dart__base__mutex*` (although I am not a big fan of the double underscore namespace thingy...)

fuchsto (approved):

dart-impl/base/src/logging.c
fuchsto: Me neither, but it's all we got. Life is hard without colons.

PR #317, opened 2017-03-09
Introduce barriers to fix TransformTest

devreal:


fuchsto:
@devreal Good catch, just wondering why this lead to a problem. `dash::transform` should be owner-computes only, so initialization of non-local values should be irrelevant?

fuchsto:
Ah, implicit `dart_flush` in the atomic / thread-safe implementation after `dart_accumulate`.
The missing barriers have been intentional, IIRC, to express the local-only nature of this algorithm. This assumption might be wrong, output range could be non-local ... will have to think about this some more, but the PR is fine.

devreal:
In `ArrayGlobalPlusLocalBlocking`, all units update the values on all other units (in the `block_idx` loop). That's where `dart_accumulate` comes into play. I initially thought that there is a bug in `MPI_Accumulate` but the problem is that some units start updating while others are still filling their local part...

fuchsto:
Ah, gotcha. Yes, the local-only assumption is intuitive, but not correct in the general case. But you opened an issue on `dash::accumulate` and `dash::transform` anyways, will take care about this there.

fuchsto (approved):

fuchsto:
Related: #216

PR #318, opened 2017-03-10
Mark DART functions as nothrow 

devreal:
Mark DART functions as `nothrow` and inline `dart_gptr_incaddr` and `dart_gptr_setunit` (marked as possibly unused). 

No (current) DART function can throw a C++ exception, hence the `nothrow` hint allows the compiler to omit exception handling code around the calls. This likely has only an effect when building with `ENABLE_ASSERTIONS=OFF`, since the assertions themselves throw exceptions.

### DIFF
--- a/dart-if/include/dash/dart/if/dart_communication.h
+++ b/dart-if/include/dash/dart/if/dart_communication.h
@@ -2,6 +2,7 @@
 #define DART__COMMUNICATION_H_
 
 #include <dash/dart/if/dart_types.h>
+#include <dash/dart/if/dart_util.h>
 #include <dash/dart/if/dart_globmem.h>
 
 /**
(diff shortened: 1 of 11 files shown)


fuchsto (changes requested):
Good to see you have overcome your hatred against white space changes ;)

But shouldn't this be `noexcept`? If I get your motivation right, you want to annotate DART functions for C++ to indicate that no exceptions are thrown. That's a good idea, but `__declspec (nothrow)` is bad style now.

It can be argued that DART might, in some parallel universe, be used by C++98-projects that don't support `noexcept`, but then we should use an ifdef-guard.

devreal:
> Good to see you have overcome your hatred against white space changes ;)

You got me on that one :white_flag: 

> But shouldn't this be noexcept? If I get your motivation right, you want to annotate DART functions for C++ to indicate that no exceptions are thrown. That's a good idea, but __declspec (nothrow) is bad style now.

Unfortunately, we cannot use `noexcept` in `extern C` code. The `nothrow` *attribute* is not the same as `throw()` in C++. It's more like an inofficial hint to the compiler that he might ignore as well. I will check whether all compilers we support actually support this attribute, might have to add guards.

fuchsto (approved):

fuchsto:
D'ah, you are right of course. Forgot about linkage.

Issue #319, opened 2017-03-10
constexpr function return is non-constant


devreal:
In my humble try to compile DASH with all compilers we support, I also tried to use the latest Cray CCE to compile the DASH part. While this has not been successful so far (some symbols are missing during linking), the following warnings could be relevant nevertheless:  
```
CC-12251 crayc++: WARNING File = /zhome/academic/HLRS/hlrs/hpcjschu/src/dash/dash-development/dash/include/dash/pattern/BlockPattern1D.h, Line = 724
  constexpr function return is non-constant
      return global(_team->myid(), std::array<IndexType, 1> {{ l_index }})[0];
             ^
          A template was detected during header processing.
            The instantiation of "std::array<IndexType, 2UL>
                      dash::BlockPattern<1, Arrangement,
                      IndexType>::initialize_local_range(dash::BlockPattern<1,
                      Arrangement, IndexType>::SizeType) const [with
                      Arrangement=(dash::MemArrange)1,
                      IndexType=dash::default_index_t]" at line 319
            The instantiation of "dash::BlockPattern<1, Arrangement,
                      IndexType>::BlockPattern(dash::BlockPattern<1,
                      Arrangement, IndexType>::SizeSpec_t,
                      dash::BlockPattern<1, Arrangement,
                      IndexType>::DistributionSpec_t, dash::Team &) [with
                      Arrangement=(dash::MemArrange)1,
                      IndexType=dash::default_index_t]" at line 769 of
                      "/zhome/academic/HLRS/hlrs/hpcjschu/src/dash/dash-develop
                      ment/dash/include/dash/Array.h"
            The instantiation of "dash::Array<ElementType, IndexType,
                      PatternType>::Array(dash::Team &) [with
                      ElementType=node_t, IndexType=dash::default_index_t,
                      PatternType=dash::BlockPattern<1, (dash::MemArrange)1,
                      dash::default_index_t>]" at line 85 of
                      "/zhome/academic/HLRS/hlrs/hpcjschu/src/dash/dash-develop
                      ment/dash/examples/app.01.mindeg/main.cpp"

CC-12251 crayc++: WARNING File = /zhome/academic/HLRS/hlrs/hpcjschu/src/dash/dash-development/dash/include/dash/iterator/GlobIter.h, Line = 613
  constexpr function return is non-constant
      return self_t(
             ^
          A template was detected during header processing.
            The instantiation of "bool dash::Array<ElementType, IndexType,
[...]

PR #321, opened 2017-03-13
Making pattern methods constexpr

fuchsto:
See #319

### DIFF
--- a/dash/include/dash/pattern/BlockPattern1D.h
+++ b/dash/include/dash/pattern/BlockPattern1D.h
@@ -664,39 +664,22 @@ public:
    *
    * \see  DashPatternConcept
    */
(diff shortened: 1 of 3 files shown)


dash/include/dash/pattern/BlockPattern1D.h
```diff
@@ -664,39 +664,23 @@ class BlockPattern<1, Arrangement, IndexType>
    *
    * \see  DashPatternConcept
    */
-  std::array<IndexType, NumDimensions> global(
+  constexpr std::array<IndexType, NumDimensions> global(
     team_unit_t unit,
```
fmoessbauer, line 668: Pass by const reference

fmoessbauer (approved):

Issue #324, opened 2017-03-14
Consistency of NArray bracket accesses


fmoessbauer:
During the implementation of the dash coarray for atomic types, I noticed that the following accesses return different types:

```cpp
dash::NArray<int,2> matrix(5,10);
auto a = matrix(0,0);  // returns a GlobRef<int>
auto b = matrix[0][0]; // returns a MatrixRef<int, 2, 0, PatternT>
```

Is this really desired? I would appreciate it if both accesses return a GlobRef, as the GlobRef has specializations for atomic accesses. Otherwise we have to implement the atomic specializations also for the `MatrixRef<int,0,...>`

devreal:
Afaics, `MatrixRef::operator[]()` returns again a `MatrixRef` to keep the `operator[]` chain going:
```
  MatrixRef<ElementT, NumDimensions, NumViewDim-1, PatternT>
    operator[](index_type n);
```

I guess we could provide an overload that returns a `GlobRef` if `(NumViewDim-1) == 0)`?

fmoessbauer:
I am currently working on exactly this functionality. Seems to be not too hard, but anyway I am interested if this was a design decision or just a "bug".

For my Coarray implementation this functionality is essential, as I have to proxy all views otherwise...

fuchsto:
Fixed in #326

PR #326, opened 2017-03-14
Return GlobRef for last matrix dimension

fmoessbauer:
This PR addresses #324 . The key idea is to return a `GlobRef` instead of a `MatrixRef<T,N,0,Pattern>` for the last free dimension. Furthermore this enables the `NArray` to work correctly with atomics, as the actual accesses are performed using the specialized `dash::GlobRef<Atomic<T>>`.

```cpp
dash::NArray<int,2> matrix(5,10);
auto a = matrix(0,0);  // returns a GlobRef<int>
auto b = matrix[0][0]; // returns a GlobRef<int> (previously MatrixRef<int, 2, 0, PatternT>)
```

Among these changes, I discovered and fixed some Bugs in the TilePattern1D specialization as well.


devreal (approved):

fuchsto:
Good proposal. I remember there was a valid reason for `MatrixRef<T, 1>`, but I think it is obsolete.
With your changes, this now also nicely corresponds to the behavior of `dash::Array` where `array[i]` returns a `GlobRef<T>`.

Actually, this reminds me: once we have generic unit test fixtures, all tests of `dash::Array<T>` should be run on `dash::NArray<T, 1>`. Semantics should be exactly the same.

fuchsto (approved):
Good proposal. I remember there was a valid reason for `MatrixRef<T, 1>`, but I think it is obsolete.
With your changes, this now also nicely corresponds to the behavior of `dash::Array` where `array[i]` returns a `GlobRef<T>`.

Actually, this reminds me: once we have generic unit test fixtures, all tests of `dash::Array<T>` should be run on `dash::NArray<T, 1>`. Semantics should be exactly the same.

PR #328, opened 2017-03-16
Set thread-support globally in DART

devreal:
Partly revert changes from #314 that enabled thead-support only for dart-mpi. It should be global in DART since other modules than dart-mpi might require thread-support (logging, tasking, younameit)


fuchsto:
@devreal As briefly discussed yesterday, global DART build options should be specified in dart/CMakeLists.txt

fuchsto (approved):

Issue #327, opened 2017-03-15
Implement `dash::mutex` for mutual exclusion 


fmoessbauer:
Mutex similar to `std::mutex` based on the synchronization interface of DART.

fuchsto:
Is there a specific use case where a mutex is required or is this a general feature request?

fmoessbauer:
It's both. In [Johnson, Coarray C++, 6.3](http://www.pgas2013.org.uk/sites/default/files/finalpapers/Day2/R4/2_paper6.pdf) he proposes a mutex which can be placed in his coarray and behaves similar to a `std::mutex`.

My idea was to implement it roughly the same way as `dash::Atomic` by specializing `dash::GobRef<dash::Mutex>`. Together with #326 this would enable the following use case:

```cpp
dash::CoArray<dash::Mutex> m;
m(i).lock()
// access data on image i
m(i).unlock();
```

fuchsto:
No, please don't. The `GlobRef<Atomic>` is a **phantom** and never explicitly defined by / for application developers.
Specializing `GlobRef` for a visible / explicit type is like `std::vector<bool>`. Looks convenient, is plain hell.

fmoessbauer:
This is a dilemma: The CoArray should behave almost exactly like the stored value, so for mutexes it's like the following:

```cpp
std::vector<std::mutex> mxstd(10);
dash::CoArray<dash::mutex> mxdash;

mxstd[0].lock();   // OK
mxdash(0).lock(); // Not OK ??
```

But after sleeping a night over this, you are right that this is not a good decision: I could make it possible to mimic the interface, but almost all semantics will change as a `dash::mutex` requires a copy per unit, while `std::mutex` requires exactly one copy for all threads. So what about introducing a `dash::CoMutex` which can be accessed similar to the `dash::CoArray<T>` but as it is its own container, this is well defined:

```cpp
dash::CoMutex comx;
comx(0).lock();
comx(0).unlock();
```

fmoessbauer:
Implemented in #329

PR #329, opened 2017-03-16
implemented dash::Mutex for mutual exclusions

fmoessbauer:
Addresses #327 

The `dash::Mutex` behaves similar to a `std::mutex` but ensures mutual exclusion within a whole dash team. Internally the `dart_lock*` interface is used. As the mutex requires (collective) initialization, they are not trivially copyable and hence cannot be placed in DASH containers.

Example:
```cpp
dash::Mutex mx; // mutex for dash::Team::All();
dash::Array<int> arr(10);
{
  std::lock_guard<dash::Mutex> lg(mx);
  int tmp = arr[0];
  arr[0] = tmp + 1;
  // TODO: this almost certainly requires a flush
}
dash::barrier();
// postcondition: arr[0] == dash::size();
```

For use cases requiring mutual accesses per unit, the mutexes can be placed in a `std::vector` storing a mutex for each unit. By wrapping this into a `dash::CoArray<dash::Mutex>` specialization this also covers the described use case.

```cpp
std::vector<dash::Mutex> mxarr(dash::size());
mxarr[0].lock();
mxarr[0].unlock();
```


dash/include/dash/Mutex.h
```diff
@@ -0,0 +1,102 @@
...
+   */
+  explicit Mutex(Team & team = dash::Team::All()):
+    _team(team)
+  {
+    dart_ret_t ret = dart_team_lock_init(_team.dart_id(), &_mutex);
+    DASH_ASSERT_EQ(DART_OK, ret, "dart_team_lock_init failed");
+  }
+  
+  Mutex(const Mutex & other)               = delete;
+  self_t & operator=(const self_t & other) = delete;
```
devreal: How about move? That should be allowed, or not?

dash/test/AtomicTest.cc
```diff
@@ -223,3 +225,53 @@ TEST_F(AtomicTest, AtomicInterface){
...
+  int tmp = shared.get();
+  shared.set(tmp + 1);
+  LOG_MESSAGE("Before %d, after %d", tmp, static_cast<int>(shared.get()));
+  // I guess here a flush is required, blocked by issue 322
+  mx.unlock();
+  
+  dash::barrier();
+  
+  while(!mx.try_lock()){
+    std::this_thread::sleep_for(std::chrono::microseconds(100));
```
devreal: Since the critical region is quite short, I would avoid the sleep to increase pressure on the lock. 

A sleep may be interesting between the `get()` and the `set()` above to better detect if the lock actually works.

devreal (commented):
Looks good to me (except two minor thing). 

However, I would block this PR until #330 is fixed.

dash/include/dash/Mutex.h
fmoessbauer: I did not provide one, as the implementation follows `std::mutex` which does not provide one as well. Because I specified a destructor, the move Ctor is not generated by the compiler. So in short: **No move.**

dash/test/AtomicTest.cc
fmoessbauer: Yes, you are right. Will change that. The main idea for the sleep was to avoid busy waiting in CI (or other scenarios where more units are spin up than cores are available).

dash/src/Mutex.cc
```diff
@@ -0,0 +1,34 @@
...
+  DASH_ASSERT_EQ(DART_OK, ret, "dart_lock_try_acquire failed");
+  return static_cast<bool>(result);
+}
+
+void Mutex::unlock(){
+  dart_ret_t ret = dart_lock_release(_mutex);
+  DASH_ASSERT_EQ(DART_OK, ret, "dart_lock_acquire failed");
+}
+
+} // namespace dash
```
fuchsto: No newline at EOF

dash/test/AtomicTest.cc
```diff
@@ -223,3 +225,53 @@ TEST_F(AtomicTest, AtomicInterface){
...
+    shared.set(tmp + 1);
+  }
+  
+  dash::barrier();
+  
+  if(dash::myid() == 0){
+    int result = shared.get();
+    EXPECT_EQ_U(result, static_cast<int>(dash::size())*3);
+  }
+}
```
fuchsto: No newline at EOF (fix your vimrc?)

fuchsto (changes requested):

dash/test/AtomicTest.cc
fmoessbauer: fixed

dash/src/Mutex.cc
fmoessbauer: fixed

dash/include/dash/Mutex.h
fmoessbauer: Ha, that was too hasty: I need copy move to be able to emplace_back a mutex in a `std::vector<dash::Mutex>`. Added it.

fuchsto (approved):

PR #332, opened 2017-03-17
Added debug build script

ddiefenthaler:
With pull request #232 the developer build setting now produce a huge amout of warnings.
There was this comment ([link](https://github.com/dash-project/dash/pull/232#issuecomment-271174182)) in the pull request:

>The "excessive" compiler warnings are only enabled in developer (not debug) build config, so we don't have to fix all of them right now. There just should be a way to make them visible.

This pull request adds a script with default debug build settings, so that the difference between the two build modes is easier to spot.

The script also builds to build.dev as apart from the warnings the actual build result should be similar.


ddiefenthaler:
@fmoessbauer On the topic of build scripts is there a reason why HDF5 is disabled in the current build.dev script?

fmoessbauer:
@ddiefenthaler I don't know any, so from my side it's ok to re-enable it.

build.debug.sh
```diff
@@ -0,0 +1,100 @@
...
+# installed.
+
+# Configure with default debug build settings:
+mkdir -p $BUILD_DIR
+rm -Rf $BUILD_DIR/*
+(cd $BUILD_DIR && cmake -DCMAKE_BUILD_TYPE=Debug \
+                        -DENVIRONMENT_TYPE=default \
+                        -DINSTALL_PREFIX=$HOME/opt/dash-0.3.0-dev/ \
+                        -DDART_IMPLEMENTATIONS=mpi \
+                        -DENABLE_THREADSUPPORT=OFF \
```
fuchsto: Why disable thread support?

fuchsto (changes requested):

build.debug.sh
ddiefenthaler: No particular reason, I just copied it from the other build scripts,
but I can activate it for them as well.

ddiefenthaler: I enabled thread support for
- build.sh,
- build.dev.sh and
- build.debug.sh

fuchsto (approved):

devreal (approved):

ddiefenthaler:
While looking at the readme I noticed that the table with the build scripts only lists 3 out of the 7 current scripts. Should this list get completed?

The table from the [README.md](https://github.com/dash-project/dash/blob/development/README.md#building-from-source):
>Script file name | Description 
>:--------------- | :-------------------------------------
>`build.sh`       | Standard release build
>`build.dev.sh`   | Development / debug build
>`build.mic.sh`   | Release build for Intel MIC (Xeon Phi)

ddiefenthaler:
Possible new table:

Script file name | Description 
:--------------- | :-------------------------------------
`build.sh`       | Standard release build
`build.minimal.sh` | Minimal build settings without external libraries and tests
`build.debug.sh`   | Debug build with logging and debug symbols
`build.dev.sh`   | Development / debug build with additional compiler warnings
`build.nasty.sh` | Build and link against a special version of MPI to reveal errors
`build.cov.sh`   | Build script with code coverage of the tests
`build.analyze.sh` | Build script with static code analysis
`build.mic.sh`   | Release build for Intel MIC (Xeon Phi)

Is this too much or would this make sense in a readme file?

BenjaProg:
Is `build.minimal.sh` supposed to build examples as well?
Currently they are enabled.

fuchsto:
@BenjaProg Yes, that's fine, the "minimal" refers to optional features, not the number of build targets.

Issue #325, opened 2017-03-14
Why does GlobPtr depend on the pattern type


fmoessbauer:
The `dash::GlobPtr` has to be declared with a pattern type, but I cannot find a single method which requires this.

```cpp
template<
  typename ElementType,
  class    PatternType  = Pattern<1> >
class GlobPtr {...};
```

fuchsto:
This is a good one, and it raises almost philosophical concerns.
The reason for the dependency on a pattern is the possibility to define pointer arithmetics corresponding to iteration space.

The ancient question behind this is the difference between `GlobPtr` and `GlobIter`.
With `GlobRef` it is easy because references cannot be re-directed, just like vanilla reference types.

Now assuming a global pointer to an element in a matrix block where block elements are non-contiguous (e.g. BlockPattern, opposed to TilePattern). When incrementing the global pointer, should it traverse elements in iteration order like `GlobIter` or not?

My suggestion is that it should not. Just like incrementing a pointer to an element of `std::list`, it would not refer to the next element in the list as pointers are agnostic of iteration order.
Pointers are dull, non-validated value-representations of addresses, iterators respect the structure of values in memory.

A often-cited formal concern is that pointers *are* iterators, or rather: iterators are a more general concept of pointers, so they should behave the same. In my understanding, `GlobPtr` is the PGAS equivalent of `T *`, so with a nod in the direction of category theory, we should be fine without the pattern dependency.

@dash-project/developers Everyone, let me know if you have objections on this.

fuerlinger:
> My suggestion is that it should not. Just like incrementing a pointer to an element of std::list, it would not refer to the next element in the list as pointers are agnostic of iteration order.
Pointers are dull, non-validated value-representations of addresses, iterators respect the structure of values in memory.

:+1:

devreal:
Agreed. It is up to the user to be careful when juggling with (global) pointers, or just don't do it. We should document though that a `dash::Array` is different from a `std::array` in that you cannot iterate over the whole array by continuously incrementing a `GlobPtr` as you would do with a `T *` and a `std::array`. 

Just one remark: The definition of `self_t operator+=(gptrdiff_t n)` in `GlobPtr` seems wrong, it should rather be returning a reference to itself. The same holds true for `operator-=`.

fuchsto:
@devreal The semantic deviation of `GlobPtr` from `T *`, which also affects global pointer distance, is bugging me since I started working on DASH. The proposal in PR #333 tries to fix it.

[...]

Issue #246, opened 2017-01-20
Why doesn't dash::copy support GlobPtr as input?


BenjaProg:
Hey, is there a reason why dash::copy doesn't accept GlobPtr as in/output?
I stumbled about this as i wanted to copy to a globmem allocated with memalloc on a remote unit.
[it's just in the comments right now](https://github.com/dash-project/dash-apps/blob/master/cowichan/winnow/main.cpp#L152) and I try to solve the problem another way, but I am still curious why ;-)

fuchsto:
`dash::copy` is an algorithm interface, that is: it has several implementations depending on the direction (local-global, global-local, global-global) with adapter wrappers for different parameter types.
GlobPtr ranges should be supported and, at least for some variants, already are.
Apparently an adapter specialization for GlobPtr is missing for your use case. I will add it.

fmoessbauer:
As discussed in #325 a `GlobPtr` does not respect the structure of values in memory. Passing a `GlobPtr` to `dash::copy` should not be allowed. Taken again the `std::list` example, this demonstrates why: Passing a pointer to a list element to an STL algorithm does not work because the data is not contiguous in memory. There you need full blown iterators.

Keeping that in mind, I guess we can close this issue.

fuchsto:
@fmoessbauer That's right in principle, but `std::copy` accepts pointers and so should we.

devreal:
We could annotate `GlobPtr` type to signal whether it points to contiguous memory (i.e., it points to a `dash::Matrix` or `dash::Array`) and at runtime error out if the user tries to copy more than one element and the memory region pointed to is not contiguous. This might be more restrictive than necessary but prevents from running total havoc in the transport layer.

fuchsto:
I think it should not. `GlobPtr` is a dull representation of an address. Or thinking the other way around: In which situation would application programmers use `dash::copy` or `std::copy` (which is explicitly supported!) on a `GlobPtr`? To avoid overhead from validation and index calculations.
Also, `GlobPtr` has no concept of containers (other than `GlobIter` types) so its implementation has no means to reflect container memory, like contiguous data. Hey, perhaps copying across the iteration space it's exactly what the application developer wants to do? It's not the concern of DASH algorithms or `GlobPtr`.

fuchsto:
Update: `GlobPtr<T>` now provides robust global random access pointer semantics for any memory space so `dash::copy` can be safely extended to support global pointers now.

BenjaProg:
[...]

PR #333, opened 2017-03-17
GlobPtr and global memory space concept (GlobHeapMem, GlobStaticMem, ...)

fuchsto:
Related to #325 #246

I'm trying to ease my discontent with pointer semantics of `GlobPtr`. In brief, I want it to comply to shared-memory pointer semantics.

A pointer in NUMA address space behaves nicely when it is incremented past a NUMA domain address space. Also, the distance between two pointers in separate NUMA domains (numa_dom_a_ptr - numa_dom_b_ptr) is well-defined.

The corresponding DASH global pointers, on the other hand, are not that polite at all. When a `GlobPtr` is incremented past the unit's local address range, it refers to an undefined address in global memory. Also, `GlobPtr` provides distance operators, but these only give sane results if both operands refer to the same unit memory space.

**Implemented in this PR:**

In the proposed implementation in this PR, `GlobPtr` depends on a global memory space (usually an instance of `GlobMem` or `GlobDynamicMem`, the latter to be renamed to `GlobHeap` for reasons I want to discuss next week).
The type dependency of `GlobPtr` to a pattern type is removed.

In some situations, a `GlobPtr` is initialized but no reference to a global memory space is available, for example in `GlobRef` constructors. A global pointer without a coupled global memory space is valid if no pointer arithmetics are performed on it.
In the vanilla shared-memory world, this corresponds to a const pointer (`T * const`, not `const T *` which is pointer-to-const).
For this, I introduced the `GlobConstPtr` concept which provides all operations of `GlobPtr` apart from those that would not be well-defined without a reference global memory space.
A `GlobConstPtr` can be **explicitly** converted to a local pointer. Pointer arithmetics on the converted local pointer then again are allowed as the restriction to local address space is obvious:

~~~c++
// Initialized from dart_gptr_t without reference to a GlobMem:
GlobConstPtr<T> gcp(some_dart_gptr);

++gcp; // fails, pointer arithmetics on const pointer not accessible

// Conversion to local pointer requires explicit cast / conversion op:
T * lptr = static_cast<T *>(gcp); // or: gcp.local();
// Application developer performed explicit conversion and definitely
// knows that the pointer's address space is limited to local memory:
++lptr; // works
~~~

Global const pointers also allow locality tests:
~~~c++
bool glob_const_ptr_is_local = GlobConstPtr<T>(some_dart_gptr).is_local();
~~~

The `GlobMem` concept is renamed to `GlobalMemorySpace` and extended by the following methods:
~~~c++
local_index_type GlobalMemorySpace::at(index_type global_index)
// -> { team_unit_t unit, index_type index }

size_type GlobalMemorySpace::local_size(unit = myid())
~~~

Global memory space maintains the local sizes of the associated units' local memory space.
As a result, `GlobPtr` is agnostic of iteration space (different from `GlobIter` types), but could traverse across the units' local memory, comparable to a bucket iterator.

The virtual global address translation required for robust global pointer arithmetics is illustrated in this article in the DASH wiki:

<http://doc.dash-project.org/internal/Publications/DASH-GDM>

glob_ptr_address_translation


dash/include/dash/memory/GlobUnitMem.h
```diff
@@ -0,0 +1,470 @@
+#ifndef DASH__GLOB_UNIT_MEM_H__INCLUDED
```
fmoessbauer, line 1: Unfortunately you copied the content of the file istead of the file. So we lose the history.

fmoessbauer (commented):
Looks very good to me. Now we should have a clean memory interface.

dash/include/dash/memory/GlobUnitMem.h
fuchsto: No, that's what git does when you `git mv`. Use `git log --follow` to see the full history.

fmoessbauer (approved):

fuchsto:
@devreal @rkowalewski 
To document the conclusion of our discussions earlier this week:

- `dart_gptr_incaddr` could/should be removed as it's just a simplicistic wrapper for `dart_gptr.addr += x` and does not support traversal across unit address ranges

Sequential coherence of global address space can easily exceed the complexity we want to delegate to DART, therefore:

- global pointer arithmetic and traversal across unit address ranges is in responsibility of DASH, more specific, the `dash::GlobPtr` types
- maintaining the sizes of the single units' address ranges is in responsibility of memory spaces, more specific, GlobMemorySpace types like `dash::GlobHeapMem`, `dash::GlobStaticMem`, `dash::GlobUnitMem`

Also see #325

rkowalewski:
> dart_gptr_incaddr could/should be removed as it's just a simplicistic wrapper for dart_gptr.addr += x and does not support traversal across unit address ranges

After reviewing the changes regarding pointer arithmetic I am not happy with this for two reasons.
- We have to expose the internal structure of a `dart_gptr_t` to DASH and possibly other users of DART. This is anything but good design. What happens if we modify the internal structure of a `dart_gptr_t`?
- It is kind of inconsistent that we used to introduce DART types and now do not exploit them for address arithmetic. What if `(sizeof(DART_TYPE) != sizeof(element_type))`?

There may possibly be more issues about that.

fuchsto:
`dart_gptr_t` is a public type and its structure is documented. If it changes, we will adapt DASH to it.
The stride for pointers in memory is not related to the DART types.
Otherwise, local memory could not be iterated in DASH, like:

```c++
dash::Array<int> array(...);
array.local[4] += 4; // <- you see my point. DART size is irrelevant.
```

fuchsto:
... so we rely on MPI/DART to detect type sizes correctly.
But you still are right in principle: If type sizes differ between DART and C++, we are in trouble.
Not sure if we validate this at the moment. We should.

rkowalewski:
Pointer arithmetics on the local view. Yes you are right, maybe I was thinking too complicated. And in general we can still overload the corresponding operators in C++. But that will be a different issue. 

Regarding the documentation of `dart_gptr_t`. OK, but anyway. It can be an error-prone effort to refactor a large code base if the internal structure has, let's say, more complicated changes.

PR #335, opened 2017-03-21
Optimizations in the MPI backend

devreal:
Reduce the number of MPI calls necessary (reported by @fuerlinger):

1. Cache rank and size in any team and globally.
2. Use `memcpy` on puts and gets if source and target is the same unit.

Also in the MPI backend:
 
3. Use `MPI_Rget` and `MPI_Wait` to decouple `dart_get_blocking` from other non-blocking operations (`dart_put_blocking` still has to call `MPI_Win_flush` to ensure remote completion)


dhinf (approved):

rkowalewski:
I think the change in [commit](https://github.com/dash-project/dash/pull/335/commits/3fc32d3b8e565dd1b080c4a8a65dff4812d6009a) are now obsolete after the discussion today since we use now a window per container and not per team, right? While semantically it makes not really a difference a [paper](http://dl.acm.org/citation.cfm?id=2798443.2780584) by the RMA Working group mentions that request based operations are generally slower due to the request management in MPI libraries.

But anyway, I understand the point that we need fine-grained completion semantics which are independent from other operations in flight. If it cause some performance related issues we can always roll back to the flush approach.

rkowalewski (commented):
looks good

dart-impl/mpi/src/dart_team_group.c
```diff
@@ -705,16 +714,23 @@ dart_ret_t dart_myid(dart_global_unit_t *unitid)
...
-  MPI_Comm_size (DART_COMM_WORLD, &s);
-  (*size) = s;
+  static int mpi_size = -1;
+  if (dart_initialized()) {
+    if (mpi_size == -1) {
+      MPI_Comm_size (DART_COMM_WORLD, &mpi_size);
+    }
+    (*size) = mpi_size;
+  } else {
+    *size = 0;
```
fuchsto: Function always returns `OK`.
For example in this line, should return `DART_ERR_OTHER`

dart-impl/mpi/src/dart_team_group.c
```diff
@@ -695,8 +700,12 @@ dart_ret_t dart_team_clone(dart_team_t team, dart_team_t *newteam)
...
 {
+  static dart_unit_t mpi_id = DART_UNDEFINED_UNIT_ID;
   if (dart_initialized()) {
-    MPI_Comm_rank(MPI_COMM_WORLD, &(unitid->id));
+    if (mpi_id == DART_UNDEFINED_UNIT_ID) {
+      MPI_Comm_rank(MPI_COMM_WORLD, &(mpi_id));
+    }
+    unitid->id = mpi_id;
   } else {
     unitid->id = -1;
```
fuchsto, line 711: Should return `DART_ERR_OTHER`

fuchsto (changes requested):

dart-impl/mpi/src/dart_team_group.c
devreal: Fixed.

fuchsto: Same in `dart_myid`

devreal: Fixed there as well, see https://github.com/dash-project/dash/pull/335/commits/e62ddc4dc45f9248f450efd755f2f0afbe89271d

fuchsto:
@devreal @rkowalewski 

Tests in CI failed for NastyMPI:
~~~
[Nasty-MPI ERROR] (src/nasty_runtime/win_storage.c:113) window is not properly initialized (no nasty id!
~~~

See [build 1676](https://circleci.com/gh/dash-project/dash/1676)

devreal:
Mhh strange, will look into that.

dart-impl/mpi/src/dart_communication.c
```diff
@@ -863,22 +951,24 @@ dart_ret_t dart_get_blocking(
...
       != MPI_SUCCESS) {
-    DART_LOG_ERROR("dart_get_blocking ! MPI_Get failed");
+    DART_LOG_ERROR("dart_get_blocking ! MPI_Rget failed");
     return DART_ERR_INVAL;
   }
-  DART_LOG_DEBUG("dart_get_blocking: MPI_Win_flush");
-  if (MPI_Win_flush(team_unit_id.id, win) != MPI_SUCCESS) {
-    DART_LOG_ERROR("dart_get_blocking ! MPI_Win_flush failed");
+  DART_LOG_DEBUG("dart_get_blocking: MPI_Wait");
+  if (MPI_Wait(&req, MPI_STATUS_IGNORE) != MPI_SUCCESS) {
```
fuchsto, line 970: 🥇

fuchsto (approved):

fuchsto:
Don't. I cannot reproduce the error reported by NastyMPI, it did not occur before and has not been observed since. Also, I can't identify any changes related to window initialization, or anything else I could nitpick about for that matter.

fuerlinger:
:+1:  Looks good, here is the updated report:

[cost-mpi.pdf](https://github.com/dash-project/dash/files/868986/cost-mpi.pdf)

One issue that seems to remain is that `MPI_Type_size` is still executed repeatedly (not cached) in the shared window case (called regular in the report).

fuerlinger:
Looks good! Here is the new report:

[cost-mpi.pdf](https://github.com/dash-project/dash/files/869778/cost-mpi.pdf)

devreal:
I removed one of the two calls to `MPI_Win_sync` in `dart_flush_all`. One thing left is the double barrier during deallocation: One happens in the d'tor of `dash::Array` and one happens in the `_deallocate` function of the `SymmetricAllocator`. @rkowalewski already noted that `dart_team_memfree` is not collective although the name suggests it to be. We could add the barrier there (slightly more efficient) before the actual deallocation and remove it from the previously mentioned places. However, `dash::Array` then has no control over the synchronization anymore, not sure if that is relevant when using other allocators.

fuchsto:
@devreal Fine for me, `dart_team_memfree` is only used in `dash::SymmetricAllocator` as expected.
Also, a non-collective deallocation of collectively allocated memory sounds suspicious to me as units might reference remote memory locations that can be deallocated without publishing the removed memory segment.

For non-collective deallocation I would expect a one-sided or epoch-synchronized memory space, and there is the non-collective (as in: one-sided) `dart_memfree` for that.
That's what I use in `GlobUnitMem` (one-sided, asymmetric) and `GlobHeapMem` (collective, imbalanced).

Are you done with this branch? I'd like to merge the PR.

devreal:
Please go ahead and merge. We will soon overhaul the memory allocation routines in DART anyway.

Issue #336, opened 2017-03-23
dash::Shared broken for composite types


fuerlinger:
The following does not compile any more because `dash::Shared` insists on keeping a  `GlobRef<Atomic<ElementType>>`  as a member.

~~~
struct test
{
  int a;
  double b;
};

int main(int argc, char* argv[])
{
  dash::init(&argc, &argv);

  dash::Shared<test> s;
  
  dash::finalize();
}
~~~

fuchsto:
@fmoessbauer Type traits for `dash::Atomic` are missing. Then, `dash::Shared<T>::atomic_ref_type` should be:

```c++
using atomic_ref_type = typename std::conditional<
                           dash::is_atomic_value<T>::value,
                           GlobRef<Atomic<T>>,
                           void
                        >::type

```

fuchsto:
... or, fundamentally: `dash::Shared` should not have an `atomic` accessor. A shared global either is atomic, or it is not. Otherwise the owner unit could perform non-atomic changes to the shared value.

For atomic operations on a `dash::Shared`, use `dash::Shared<dash::Atomic<T>>`.

fmoessbauer:
I agree on that, but IMO this should not be implemented or forced using a type trait. Please have a look at my comment in [PR 338](https://github.com/dash-project/dash/pull/338#discussion_r107706607)

fuchsto:
In the current version of the PR, I just removed `dash::Shared.atomic`.

fuchsto:
Fixed in PR #338, also added test of `dash::Shared` for composite value types.

PR #338, opened 2017-03-23
Added atomic value type trait, removed dash::Shared.atomic

fuchsto:
See #336 

- New implementation of `dash::Shared<T>` now automatically adapts semantics to `GlobRef<T>`, in particular for specializations like `GlobRef<Atomic<T>>`
- `dash::Shared` now has correct atomic value semantics
- Removed `dash::Shared.atomic`: for atomic operations on shared values, use `dash::Shared<dash::Atomic<T>>`
- Adjusted tests of `dash::Atomic` and `dash::Shared` accordingly
- Added test of `dash::Shared` for composite value types
- Added type trait `dash::is_atomic_compatible<T>` to test if values of `T` can be used for `dash::Atomic`
- Restricted numeric operators (++, --, +=, -=, fetch_and_op, ...) of `GlobAtomicRef` to integral types

**Important side note**:

As discussed in my previous comments on atomic value semantics (https://github.com/dash-project/dash/issues/249#issuecomment-275514703), atomicity must be bound to a type.
That is: either a value is atomic, or it is not. If a value could be used for atomic operations in some context and for non-atomic operations in another, consistency can be spuriously violated and operation semantics cannot be ensured at compile time.

Example:

~~~c++
dash::Shared<int> shared;

if (myid == 0) {
  shared.atomic.fetch_and_add(10); // atomic in this context
} else {
  shared.get() += 10; // not atomic here
}
~~~

With changes in this PR:

~~~c++
dash::Shared<dash::Atomic<T>> shared; // atomicity bound to type

if (myid == 0) {
  shared.get().fetch_and_add(10);
} else {
  shared.get() += 10; // calls fetch_and_add(10)
}
~~~


dash/include/dash/Meta.h
```diff
@@ -90,6 +90,19 @@ struct is_container_compatible :
...
+ * Type trait indicating whether a type can be used for global atomic
+ * operations.
+ */
+template <typename T>
+struct is_atomic_compatible
+: public std::integral_constant<
+           bool,
+              dash::is_container_compatible<T>::value
+           && sizeof(T) <= sizeof(std::size_t)
+         >
```
fmoessbauer: I'm not too happy with this implementation as `std::atomic<T>` supports any type that is trivially copyable. To check if atomic operations are used instead of locks, [`is_lock_free()`](http://en.cppreference.com/w/cpp/atomic/atomic/is_lock_free) or `is_always_lock_free` (c++17) can be used. We should change the implementation in this way. However this could be done past 0.3.0.

devreal: With the current implementation, this should be 

```
std::is_integral<T>
```
instead of 
```
sizeof(T) <= sizeof(std::size_t)
```

fuchsto: No, atomic operations are not related to integral types. For example, type punning from any 64bit value is fine for lock-free CAS.

fuchsto: @fmoessbauer And lock-freedom is not related to atomic operations, neither. Atomicity is semantics, locks are implementation.

fuchsto: We restrict to 64bit because MPI does not support wider types for by-value-atomic reduce operations. As a side-effect, atomic operations in local memory usually will be executed bia `CMPXCHG` or some other lock-free instruction. But: `std::is_lock_free` does not ensure that MPI supports reduce operations on that type.

dash/include/dash/Shared.h
```diff
@@ -7,7 +7,7 @@
 #include <dash/GlobRef.h>
 #include <dash/Allocator.h>
 
-#include <dash/atomic/GlobAtomicRef.h>
+#include <dash/Atomic.h>
```
devreal: The include seems not needed atm.

dash/include/dash/Meta.h
devreal: > No, atomic operations are not related to integral types. For example, type punning from any 64bit value is fine for lock-free CAS.

Correct in general but not in our case: MPI does not allow any other types than integral types in atomic operations. So if you try to use `dash::Atomic` on a `double` you will end up with a runtime error... (coming out of DART, not MPI)

fuchsto: I know, but as I mentioned above: type punning can be automatically applied then. The trait is fine.

devreal: That sounds like a horrible idea to me: the user will expect an atomic add on a `double` to behave like an add, not like an add on a pointer type punned to int64_t. This semantic is only useful for `compare_and_swap`, which does not do type-punning at the moment.

fuchsto: No, why? Type punning is only used when required, it's not used for doubles.
And if the user specifies a composite type for atomics, it's obvious that fetch-and-add is nonsensical.

This trait states, as decumented in it's doxygen, whether a type is eligible for atomic operations. `struct { char; char; }` is.
It does not express semantic stability of **any** atomic operation. That would be a different trait.

fuchsto: Or simply put, the trait answers the question: "is it okay to have an atomic type for this T?".
Operators for fetch-and-add can then be disabled if T is not a numeral, but that's unrelated.

devreal: What is the sense in allowing `dash::Atomic` on any type that will fail to perform correct atomic operation? 

The name of the trait is `dash::is_atomic_compatible` and  neither a `struct` nor `double` are *compatible* because the operations on the are not well defined.

fuchsto: CAS, atomic read, atomic write. Operators like += can be disabled but `Atomic<CompositeChars>` is fine. The current implementation in development just "eats" any type which is definitely less expressive than this trait.
The restriction to numerics in MPI is an implementation detail of dart-mpi - which needs to be addressed - but does not affect atomic type semantics in general.

devreal: Please implement, preferably in this PR.

fuchsto: Done, see `dash/atomic/GlobAtomicRef.h`.

fmoessbauer: What a discussion. @fuchsto It is still not clear to me, why large structs cannot be wrapped with `dash::Atomic`. As you mentioned, atomic refers to the semantics, but not to the implementation. So if neglecting performance, atomic operations on data of any size should be possible by using a `lock()` and `unlock()`. 

If I get the proposed implementation right, we are going to support only types up to `std::size_t`. Why that?

devreal:
Please do not merge, I will commit some changes.

dash/include/dash/Meta.h
fuchsto: @fmoessbauer Absolutely! But currently, a DART reduce operation would not work for types with size greater than 64 bit, right?
Would compare-exchange on `dash::Atomic<T>` currently work for `T = struct { short foo[128] }` ?

fuchsto: ... So: yes, eventually the atomic type traits should depend on types supported by atomic operations in DART or even split into `dash::is_atomic_compatible` and `dash::is_global_atomic_compatible`, but for now the definition in this PR should be correct.

fmoessbauer: So in fact the documentation / specification of dart atomic operations is missing. I'm fine with it if dart works only on integral types or types of a given size (and this is documented!), however this is not a reason to restrict the functionality on the DASH side. DASH is intended to follow the STL as close as possible.

I see it similar to using `std::atomic` on CPUs which do not support lock-free atomic instructions for a given type: `std::atomic` still guarantees the semantics, but is implemented differently. I propose to add a trait to check if the type is natively supported by DART atomic. If so, the `GlobRef<Atomic<T>>` uses the dart atomic functions, if not it uses locks.

Of course, some functions are only available for integral types, but that is another topic.

fuchsto:
@devreal What are your changes about? Will you push them today?

dash/include/dash/Meta.h
fuchsto: Right, that's what I meant with `dash::is_global_atomic_compatible` to distinguish between "atomic operations supported by STL" and "atomic operations supported by DART".
Because some types are fine for `std::atomic<T>` but not supported by DART reduce operations, like `struct DoublePair { double; double; }`.
It would be okay to use `dash::Atomic<DoublePair>`, though:
Compare-exchange would then be supported for local memory (probably lock-based) but not across units.
Still, I stand by my point:

- With the current implementation of DART and dash::Atomic, the type trait `is_atomic_compatible` in this PR is correct.
- `std::is_integral<T>` would be incorrect
- `std::is_lock_free<T>` is unrelated to type semantics
- Type punning is fine and numeric operators (++, --, ...) are only defined for integral types (as implemeted in this PR)

devreal:
I will fix the type traits (after having consulted the MPI standard again) and add type punning. Will try to push them tonight.

dash/include/dash/Meta.h
devreal: Type punning and correct checks should now be there. The trait you introduced simply is not strong enough to guarantee correctness on all possible types so my comment was reasonable. However, I admit that std::is_integral is too restrictive if applied globally.

fuchsto:
@devreal 

Your implementation of type punning has some issues:
- `internal::atomic_punned_type` is not implemented as type trait and could therefore not be specialized
- the use of `static_assert` inside methods prevents SFINAE
- value type is tested against `std::is_integral` which is too strict
- `GlobRef<Atomic>.exchange` would only work for integral and floating point types which is too strict
- compatiblity of value type for atomic non-arithmetic operations is already ensured by `dash::Atomic` and should not be re-checked (with different constraints)

Also, in the unit test for atomic operations with type punning:
- `operator<<` for type `container` is missing and breaks development builds
- CAS is only tested for old value = new value

See my changes in https://github.com/dash-project/dash/pull/338/commits/53280921087d61acdefdba11f98a9c9b2fce3244 https://github.com/dash-project/dash/pull/338/commits/72601e14c74f324ebc0b2b82fda3fa437018d1cc

dash/include/dash/Meta.h
fuchsto: Again, the type trait `is_atomic_compatible` does **not** state that all existing atomic operations on a type are supported. It just tests if a type can be used as value type of `dash::Atomic` *at all*.
Supported atomic operations are unrelated and now tested in DASH operation types like `dash::plus` (requires arithmetic type) or `dash::second` (no restrictions).

fuchsto:
With my recent changes:

- `dash::Atomic<T>` ensures, via `dash::is_atomic_compatible`, that values of `T` can be used as atomics such that atomic operations compare-exchange, fetch and exchange are supported (**not** providing any guarantee that all atomic operations are supported)
- DASH operation types like `dash::plus<T>`, `dash::second<T>`, ... ensure that the respective operation can be performed on values of type `T`. For example, `dash::sum<std::pair<int, int>>` would raise a compiler error
- Type punning, e.g. from composite types to `DART_TYPE_LONGLONG`, is implemented in `GlobRef<dash::Atomic>` via an internal trait `atomic_punned_type` but does not check type compatibility because:
    - `dash::Atomic` already checks availability of non-arithmetic atomic operations like `compare_exchange`
    - `GlobRef<dash::Atomic>.fetch_and_op` expects a DASH operation type that would raise a compiler error if the requested operation cannot be performed on `T`

fuchsto:
Finally, with additional changes in my recent commits:

- `dash::Atomic<T>` ensures, via `dash::is_atomic_compatible`, that values of `T` can be used as atomics such that atomic operations compare-exchange, fetch and exchange are supported (**not** providing any guarantee that all atomic operations are supported)
- DASH operation types like `dash::plus<T>`, `dash::second<T>`, ... ensure that the respective operation can be performed on values of type `T`. For example, `dash::sum<std::pair<int, int>>` would raise a compiler error
- Conversion from native data type to corresponding DART data type for reduce operations (implies atomic ops) now provided as trait `dash::dart_punned_datatype<T>` which only applies type punning if `T` cannot be used for reduce operations directly (e.g. `double` is not punned, `char [4]` is punned to `DART_TYPE_INT`)
- `GlobRef<dash::Atomic>` uses `dash::dart_punned_datatype` and is not responsible for type compatibility checks because:
    - `dash::Atomic` already checks availability of non-arithmetic atomic operations like `compare_exchange`
    - `GlobRef<dash::Atomic>.fetch_and_op` expects a DASH operation type that would raise a compiler error if the requested operation cannot be performed on `T`

fuchsto:
Finally, with additional changes in my recent commits:

- `dash::Atomic<T>` ensures, via `dash::is_atomic_compatible`, that values of `T` can be used as atomics such that atomic operations compare-exchange, fetch and exchange are supported (**not** providing any guarantee that all atomic operations are supported)
- DASH operation types like `dash::plus<T>`, `dash::second<T>`, ... ensure that the respective operation can be performed on values of type `T`. For example, `dash::sum<std::pair<int, int>>` would raise a compiler error
- Conversion from native data type to corresponding DART data type for reduce operations (implies atomic ops) now provided as trait `dash::dart_punned_datatype<T>` which only applies type punning if `T` cannot be used for reduce operations directly (e.g. `double` is not punned, `char [4]` is punned to `DART_TYPE_INT`)
- `GlobRef<dash::Atomic>` uses `dash::dart_punned_datatype` and is not responsible for type compatibility checks because:
    - `dash::Atomic` already checks availability of non-arithmetic atomic operations like `compare_exchange`
    - `GlobRef<dash::Atomic>.fetch_and_op` expects a DASH operation type that would raise a compiler error if the requested operation cannot be performed on `T`

fuchsto:
Finally, with additional changes in my recent commits:

- `dash::Atomic<T>` ensures, via `dash::is_atomic_compatible`, that values of `T` can be used as atomics such that atomic operations compare-exchange, fetch and exchange are supported (**not** providing any guarantee that all atomic operations are supported)
- DASH operation types like `dash::plus<T>`, `dash::second<T>`, ... ensure that the respective operation can be performed on values of type `T`. For example, `dash::sum<std::pair<int, int>>` would raise a compiler error
- Conversion from native data type to corresponding DART data type for reduce operations (implies atomic ops) now provided as trait `dash::dart_punned_datatype<T>` which only applies type punning if `T` cannot be used for reduce operations directly (e.g. `double` is not punned, `char [4]` is punned to `DART_TYPE_INT`)
- `GlobRef<dash::Atomic>` uses `dash::dart_punned_datatype` and is not responsible for type compatibility checks because:
    - `dash::Atomic` already checks availability of non-arithmetic atomic operations like `compare_exchange`
    - `GlobRef<dash::Atomic>.fetch_and_op` expects a DASH operation type that would raise a compiler error if the requested operation cannot be performed on `T`

fuchsto:
Finally, with additional changes in my recent commits:

- `dash::Atomic<T>` ensures, via `dash::is_atomic_compatible`, that values of `T` can be used as atomics such that atomic operations compare-exchange, fetch and exchange are supported (**not** providing any guarantee that all atomic operations are supported)
- DASH operation types like `dash::plus<T>`, `dash::second<T>`, ... ensure that the respective operation can be performed on values of type `T`. For example, `dash::sum<std::pair<int, int>>` would raise a compiler error
- Conversion from native data type to corresponding DART data type for reduce operations (implies atomic ops) now provided as trait `dash::dart_punned_datatype<T>` which only applies type punning if `T` cannot be used for reduce operations directly (e.g. `double` is not punned, `char [4]` is punned to `DART_TYPE_INT`)
- `GlobRef<dash::Atomic>` uses `dash::dart_punned_datatype` and is not responsible for type compatibility checks because:
    - `dash::Atomic` already checks availability of non-arithmetic atomic operations like `compare_exchange`
    - `GlobRef<dash::Atomic>.fetch_and_op` expects a DASH operation type that would raise a compiler error if the requested operation cannot be performed on `T`

devreal:
Thanks for the correct type trait. I think we are getting there :)

> DASH operation types like `dash::plus<T>`, `dash::second<T>`, ... ensure that the respective operation can be performed on values of type T. For example, `dash::sum<std::pair<int, int>>` would raise a compiler error

This breaks the following valid use-case:

```
struct type{
typedef struct type self_t;
int a, b;
self_t operator+(const self_t& other);
}

[...]
dash::accumulate(..., dash::plus<struct type>())
```

`std::accumulate` does not enforce `std::is_arithmetic` and it is not required in our case, either.

fuchsto:
Hm, well.
`dash::accumulate` does not enforce `is_arithmetic`, the restriction is in `dash::plus<T>`.
And DASH operations may of course be more restrictive than their STL counterparts, that's why I introduced them.

Your example is only valid with the current implementation of `dash::accumulate` because it happens to resort to local-only operations.
However, there is no global reduce operation available for `sum<struct type>` and future implementations of `dash::accumulate` will use DART reduce operations.

Currently, `dash::accumulate` is rather proof-of-conceptish and based on `dash::Array`, a comfortable but arguably expensive solution. For a reasonable large team size (4096 is not even close to exascale) MPI reduce operations are way more efficient than allocating a `dash::Array` with all its windows and barriers.

Finally, it's a trait! Traits are meant to be specialized for custom types if the default deduction does not fit. In your case:

~~~c++
template <>
struct is_arithmetic<MyType> : public std::integral_constant<bool, true>
{ };
~~~

devreal:
> Currently, dash::accumulate is rather proof-of-conceptish and based on dash::Array, a comfortable but rather expensive solution.

Agreed. But it's the only solution we have for custom types. Using DART functionality for basic types should be an optimization that we can employ later. 

> Finally, it's a trait! Traits are meant to be specialized for custom types if the default deduction does not fit. In your case:

I could do that, of course. That would also allow me to pass it to `GlobRef<dash::Atomic>::fetch_op` and watch my data go corrupted silently thanks to type punning. `fetch_op` simply should not employ type punning. DART operations and C++ operators are two different stories.

fuchsto:
> Agreed. But it's the only solution we have for custom types.
> Using DART functionality for basic types should be an optimization that we can employ later.

The fact that `dash::accumulate` is currently implemented with local operations is an implementation detail. In general, DASH algorithms require that they can be executed as global (as in: RDMA, because PGAS) operations.

If an application would rely on the *implementation* of a DASH algorithm ("hey, my custom type works even if MPI doesn't support it, nice"), the application would break with any future minor version release of DASH that optimized the algorithm.

Actually there is an open request to re-implement it with DART reduce operations, it might make it into 0.3.0 even.

This will be relaxed with the introduction of execution policies: `dash::execute_local` would pick a variant of the algorithm that only employs local operations and map any DASH operation type like `dash::plus` to its local equivalent like `std::plus`, consequently losing the type constraint `is_arithmethic`.

> fetch_op simply should not employ type punning. DART operations and C++ operators are two 
> different stories.

Yes, that's why we don't use C++ operators but DASH operators.
There is no problem with type punning in `fetch_op`. It depends on the operation whether it is legal, like `fetch_op(dash::second<T>)`, or not. That's why the type constraints are specified exactly by these.

devreal:
> The fact that dash::accumulate is currently implemented with local operations is an implementation detail. In general, DASH algorithms require that they can be executed as global (as in: RDMA, because PGAS) operations.

Not an argument here. To quote you: RDMA is an implementation detail of PGAS. 

A user that specifies the `is_arithmetic` type trait for his custom type should be able to use a version of the DASH algorithms that produces a correct result or receive a compiler error. The type trait `is_arithmetic` can be true for any type (because the user specified the correct operators and is sure that in C++ semantics the trait is fulfilled).  If the size of that type is <= 8 byte `dash::sum` will happily accept it and produce incorrect results in `fetch_op`. The fact whether a type satisfies arithmetic properties in C++ is independent of whether we can apply DART atomic arithmetic operations on it...

fuchsto:
> Not an argument here. To quote you: RDMA is an implementation detail of PGAS.

I wrote "as in RDMA" to illustrate what I mean by "global operation".
That is: PGAS is possibly implemented by means of RDMA, that is we have to restrict to global operations.
I did not imply the inverse implication.

In brief, I explained why your example is not a valid use case and why it would be harmful, even.
Can you explain if and why you object?

Yes, if a user could define a custom type as arithmetic (like `pair<int, int>` for reasons) which must be *explicitly stated* and then use `fetch_and_op(dash::multiplies)` on it which *again must be explicitly stated*, then it could break.

Regarding your other remarks, please read my previous comments.

Which concrete changes do you suggest?

devreal:
> Which concrete changes do you suggest?

We cannot allow `fetch_op` on any types other than integral and floating point (except for `DART_REPLACE` which works through type punning). 

However, I tried to make my point clear and the discussion is not getting anywhere here. For the record: this PR breaks support for `dash::accumulate` on custom types and opens the door for data corruption in `fetch_op` (with any other operation than `dash::second`) for custom types with the `is_arithmetic` trait defined. If we are fine with this, please go ahead and merge. There are more pressing issues at hand...

fuchsto:
Yes, and I argued that support for DASH-algorithms on custom types indeed *must not be supported* and would lead to incompatibilty with future versions of DASH (see https://github.com/dash-project/dash/pull/338#issuecomment-288963203) if they cannot be executed globally.
So: breaking support for global operations on custom types that are not compatible with these global operations is, no matter how hard I think about it, a good thing.

Again, what is your objection?

I can't think of a custom type that is not arithmetic but should be supported in `dash::accumulate` or `dash::transform` etc. with arithmetic DASH operations. The constraint `is_arithmetic` used in DASH operation types can well be replaced, of course, but I don't see a semantic problem so I asked for concrete alternatives.

Yes, custom type traits can break correctness if they "lie" about actual behavior, but that's a ... metaphysical problem.

If there are no concrete alternatives presented, I will merge.

devreal (approved):
After giving this some more thought I think I am OK with this PR. I realized that specializing `std::is_arithmetic` is a bad idea in general, so the traits are fine. We should aim for DASH algorithms to support all types supported by DASH containers (given appropriate operators defined on them) at some point even if that means that DART primitives cannot be used (at the cost of a performance penalty compared to arithmetic types). But that is something for future discussions...

fuchsto:
@devreal After giving this some more thought I defined `dash::is_arithmetic<T>` that ensures that `T` matches a plain DART data type without punning ;)
There are some more non-arithmetic ops now: `dash::bit_and`, `dash::bit_or`, `dash::bit_xor`. A unit test for `fetch_op` with punned value type has been added.

Issue #342, opened 2017-03-24
Problem with dash::memalloc 


devreal:
I am using `dash::memalloc` in my `ThreadSafetyTest`s. I'm pretty sure it had worked before but now every call returns a global pointer with offset 0: `dash::GlobPtr<i>(000000|00|0000|0000|0000000000000000)`. 

I have created a simple  test case that currently fails: https://github.com/dash-project/dash/blob/feat-dart-reducempicalls/dash/test/memory/LocalAllocatorTest.cc#L19

Output:
```
[  ERROR   ] [UNIT 0] in /home/joseph/src/dash/dash/dash/test/memory/LocalAllocatorTest.cc:28
Expected: (ptr1) != (ptr2), actual: dash::GlobPtr<i>(u000000|f00|s0000|t0000|o0000000000000000) vs dash::GlobPtr<i>(u000000|f00|s0000|t0000|o0000000000000000)
```

Two allocations should never get the same pointer if there has not been a free in between. 

The same test case with DART functionality succeeds: https://github.com/dash-project/dash/blob/feat-dart-reducempicalls/dash/test/dart/DARTMemAllocTest.cc#L40

That test succeeds. Could this be related to the memory spaces that have been added recently?

fuchsto:
Pretty sure that's on me. `dash::memalloc` used to create a "free-floating" memory space with rather arbitrary global pointer iteration. The type dependencies are fine now, but apparently I broke something else there.
Unit tests passed on my machines with some different MPI versions and compilers, though. I will check your test case now.

rkowalewski:
I am looking into  this.

fuchsto:
I might have tracked it down. @devreal can you please re-build with logging enabled and attach the log snippet around the `dash::memalloc` call?

rkowalewski:
After inspecting what's going on I can see that there is a [TODO](https://github.com/dash-project/dash/blob/feat-dart-reducempicalls/dash/include/dash/GlobPtr.h#L137) by @fuchsto. We do not bind the GlobUnitMem temporary to a global pointer.

So please fix this. I am wondering why these refactorings of the global memory concept are actually at least partially included in this release.

fuchsto:
It's not related to global memory concepts, these are preparations for fixes of global pointer arithmetics.
[...]

PR #344, opened 2017-03-24
Re-added ownership check in GlobUnitMem

fuchsto:
Addresses #342

### DIFF
--- a/dash/include/dash/memory/GlobUnitMem.h
+++ b/dash/include/dash/memory/GlobUnitMem.h
@@ -194,14 +194,12 @@ public:
   /**
    * Copy constructor.
    */
-  GlobUnitMem(const self_t & other)
-    = default;
+  GlobUnitMem(const self_t & other)      = delete;
 
   /**
    * Assignment operator.
    */
-  self_t & operator=(const self_t & rhs)
-    = default;
+  self_t & operator=(const self_t & rhs) = delete;
 
   /**
    * Equality comparison operator.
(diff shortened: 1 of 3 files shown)


dash/include/dash/memory/GlobUnitMem.h
```diff
@@ -187,7 +185,9 @@ class GlobUnitMem
   ~GlobUnitMem()
   {
     DASH_LOG_TRACE_VAR("GlobUnitMem.~GlobUnitMem()", _begptr);
-    _allocator.deallocate(_begptr);
+    if (_owns_mem) {
+      _allocator.deallocate(_begptr);
+    }
```
devreal, line 190: With something like that, shouldn't the copy constructor of `GlobUnitMem` be `deleted` to avoid dangling references if one of the copies goes out of scope?

rkowalewski:
This approach looks rather like a hacky workaround to me. Depending on which constructor is called a GlobUnitMem instance is either the owner or not. But anyway it will be redesigned in future.

rkowalewski (approved):
My comments are described in the issue.

fuchsto:
@rkowalewski I see your point but the differentiation by constructor is a common principle:

- If `GlobMem` is initialized with an existing global pointer, then the caller owns the allocated memory and `GlobMem` is just an interface to it. Only the caller knows how the memory referenced by the `dart_gptr_t` has been aquired so only the caller can decide how and when to release it

- If `GlobMem` is initialized with a size and a team, it allocates the memory itself and of course takes ownership

The copy/move semantics mentioned by @devreal is an actual issue, however. I will fix this before merging.

PR #345, opened 2017-03-25
Global pointer arithmetics and pointer distance in global memory

fuchsto:
Introduces correct random access on global address space. `GlobPtr` now supports iteration across unit memory spaces for balanced, unbalanced and dynamic memory spaces.

See:
https://github.com/dash-project/dash/blob/a26b27ea142449d0a9f4beb4d201f7a912aad002/dash/test/memory/GlobStaticMemTest.cc#L33

The full story is discussed in https://github.com/dash-project/dash/pull/333

Before the changes in this PR, `GlobPtr` continued iteration past the local memory space instead of the subsequent unit's address range.
Some places in DASH rely on the previous "relaxed" semantics of global pointer arithmetics, so some additional adjustments will soon be pushed.

Update: Done and ready for review now


dash/include/dash/GlobPtr.h
```diff
@@ -543,9 +488,57 @@ class GlobPtr
...
-    dart_team_myid(_dart_gptr.teamid, &luid);
-    return _dart_gptr.unitid == luid.id;
+    DASH_ASSERT_RETURNS(
+      dart_team_myid(_rbegin_gptr.teamid, &luid),
+      DART_OK);
+    return _rbegin_gptr.unitid == luid.id;
+  }
+
+private:
+  void increment(index_type offs) {
```
rkowalewski, line 493: This sequence looks really ugly. I would suggest to modify the code  for better readabality with appropriate variables or even a small macros. Something like this:
~~~cpp
//Local Variables
auto const & raddr = _rbegin_gptr.addr_or_offs.offset;
auto const lpos = _raddr / sizeof(value_type);
// Alternatively
#define _LPOS (addr) (addr / sizeof(element_type))
auto const lpos = _LPOS(r_addr);
~~~

I remember we decided to omit `dart_gptr_incaddr` since it does currently nothing but incrementing address and offset. This might be actually problematic IMHO. Will write something in #333

dash/include/dash/memory/GlobHeapMem.h
```diff
@@ -429,6 +429,23 @@ class GlobHeapMem
     return unit_local_size;
   }
```
rkowalewski: See my comments to `GlobPtr`. The same applies here. Maybe we can centralize the pointer arithmetic stuff.

dash/include/dash/memory/GlobStaticMem.h
```diff
@@ -294,6 +288,23 @@ class GlobStaticMem
     return _nlelem;
   }
 
+  index_type distance(const pointer & gbegin, const pointer & gend) const {
```
rkowalewski: Even more duplicated stuff here.

rkowalewski (changes requested):

dash/include/dash/memory/GlobStaticMem.h
fuchsto: The calculation of global pointer distance depends on `Glob*Mem` as discussed in #333 so it cannot be implemented as a single independent function.
For comparison, see `dash::GlobUnitMem`.

We could specialize `dash::distance` for `GlobPtr` and its dependent `Glob*Mem` and define `GlobPtr::operator-(const self_t &)` as `dash::distance(*this, other)`.

dash/include/dash/memory/GlobHeapMem.h
fuchsto: What's wrong with `GlobHeapMem::local_size`? It's fundamentally different from the implementation in any other memory space type, so why should it be centralized?

rkowalewski: Öhm actually the comment refers to the wrong line. What I mean is `GlobHeapMem::distance`. And the problem is not the replicated logic itself. It is clear that `GlobHeapMem::distance` is different from that in `GlobStaticMem`. What I mean is the `gptr.addr_or_offset.offset / sizeof(element_type)` which is repeated in at least 3 or 4 files. We could provide a simple macro for this.

fuchsto: Ah, yes. Actually I want to get rid of the division ops somehow and multiply the byte-based offsets instead. As we know from ParComp, division is horribly expensive and pointer increments should have as little overhead as possible.

dash/include/dash/GlobPtr.h
```diff
@@ -104,12 +110,24 @@ class GlobPtr
...
    */
-  constexpr GlobPtr(
+  GlobPtr(
     const MemorySpace * mem_space,
-    dart_gptr_t gptr)
-  : _dart_gptr(gptr)
+    dart_gptr_t         gptr,
+    index_type          rindex = 0)
+  : _rbegin_gptr(gptr)
   , _mem_space(reinterpret_cast<const MemorySpace *>(mem_space))
```
devreal: What is this `reinterpret_cast` good for?

dash/include/dash/GlobPtr.h
```diff
@@ -290,47 +305,41 @@ class GlobPtr
...
    */
   constexpr index_type operator-(const self_t & rhs) const noexcept
   {
-    return _dart_gptr - rhs._dart_gptr;
+    return ( _rbegin_gptr.unitid == rhs._rbegin_gptr.unitid ||
+             _mem_space == nullptr
+             ? ( _rbegin_gptr.addr_or_offs.offset
+                 - rhs._rbegin_gptr.addr_or_offs.offset )
+               / sizeof(value_type)
+             : _mem_space->distance(rhs, *this) );
```
devreal: correct ordering or arguments? Better: `_mem_space->distance(*this, rhs)`

dash/include/dash/memory/GlobHeapMem.h
devreal: It would be interesting to see whether this ends up to be a `div` in the object code. I'd expect the compiler to optimize it since `sizeof(element_type)` is a compile-time constant.

devreal (commented):
Comments inline. Looks good in general :+1:

dash/include/dash/GlobPtr.h
fuchsto: Ah, the argument type should be `const MemSpaceT`. Assignment and copy from other GlobPtr with different memory space type is essential to convert `GlobPtr<T>` to `GlobPtr <const T>`

dash/include/dash/GlobPtr.h
fuchsto: No, because `*this - rhs = distance(rhs, *this)`

dash/include/dash/memory/GlobHeapMem.h
fuchsto: ... and `sizeof(element_type)` is a power of two, so it should become a left-shift, right.

fuchsto: ... right-shift I mean

dash/include/dash/GlobPtr.h
devreal: You're right, my mistake.

fuchsto:
@rkowalewski I eliminated duplicate code using specialization of `dash::distance` for global pointer types and their dependent memory space type.
The increment/decrement operations in `GlobPtr` have been cleaned up a bit.

fuchsto:
@rkowalewski @devreal 

Changes following your requests:

- I managed to eliminate `GlobHeapMem.distance` and `GlobStaticMem.distance` (and thus their duplicate implementations) using specialization of `dash::distance` for global pointer types and their dependent memory space type.
- the increment/decrement operations in `GlobPtr` have been cleaned up a bit.

Also:

- updated other specializations of `dash::distance` accordingly
- updated doxygen documentation and removed resolved TODOs

fuchsto:
@rkowalewski @devreal 

Changes following your requests:

- I managed to eliminate `GlobHeapMem.distance` and `GlobStaticMem.distance` (and thus their duplicate implementations) using specialization of `dash::distance` for global pointer types and their dependent memory space type. (see https://github.com/dash-project/dash/blob/22b000777dd5fb444e0ccb9913151876fce8189b/dash/include/dash/GlobPtr.h#L793)
- the increment/decrement operations in `GlobPtr` have been cleaned up a bit.

Also:

- updated other specializations of `dash::distance` accordingly
- updated doxygen documentation and removed resolved TODOs

fuchsto:
@rkowalewski @devreal 

Changes following your requests:

- I managed to eliminate `GlobHeapMem.distance` and `GlobStaticMem.distance` (and thus their duplicate implementations) using specialization of `dash::distance` for global pointer types and their dependent memory space type. (see https://github.com/dash-project/dash/blob/22b000777dd5fb444e0ccb9913151876fce8189b/dash/include/dash/GlobPtr.h#L793)
- the increment/decrement operations in `GlobPtr` have been cleaned up a bit.

Also:

- updated other specializations of `dash::distance` accordingly
- updated doxygen documentation and removed resolved TODOs
- refactored `GlobBucketIter` to specialization of `GlobPtr` so the global pointer concept now is maintained for dynamic memory spaces

devreal (approved):
Looks good to me :)

Issue #116, opened 2016-11-17
Supporting strided ranges


fuchsto:
From issue #3:

Consider the following use cases:

~~~c++
// assuming row-major storage
dash::Matrix<double, 2> matrix(...);

auto lcol = matrix.local.col(0);
auto lrow = matrix.local.row(0);
auto n_lcol = std::distance(lcol.begin(), lcol.end());
auto n_lrow = std::distance(lcol.begin(), lrow.end());

std::vector<double> copy_lrow(n_lrow);
std::vector<double> copy_lcol(n_lcol);

// works:
dash::copy(lrow.begin(), lrow.end(), copy_lrow.data());
// fails:
dash::copy(lcol.begin(), lcol.end(), copy_lcol.data());
~~~

Slicing columns from a matrix with row-major storage order is fine, but copying this slice is not.
This is a missing feature: DASH does not support strided copying yet.

~~~
[ 0 1 2 3 ]  <- row contiguous in memory   | 8 | 4 | 0 |  <- column is strided, requires packing
[ 4 5 6 7 ]                                | 9 | 5 | 1 |     into temporary buffer [ 0 4 8 ] or
[ 8 9 0 1 ]        -- rot 90° -->          | 0 | 6 | 2 |     single copy operations for every
                                           | 1 | 7 | 3 |     value
~~~

This affects the specification of at least the following interfaces:

- Implementations of `dash::copy` should support strided input ranges, the interface must not change
- `dash::local_range` should be extended by `dash::local_ranges` which resolves a list of local ranges in
  a global input range
- `dash::local_range` should return the first local range in a global input range

PR #347, opened 2017-03-27
Fix Copy of C Array to Dash Array across multiple units

rkowalewski:
Addresses #346


rkowalewski:
It was not intended to submit the clang-format in this pull request. But now we have one.

rkowalewski:
@fuchsto My changes assume that local memory is always contiguous. So it will probably not work with patterns where local memory is fragmented. Please consider for review this special case. I am not sure if any of our copy algorithms can handle this.

dash/include/dash/algorithm/Copy.h
```diff
@@ -493,17 +493,42 @@ GlobOutputIt copy_impl(
...
-      out_first.dart_gptr(),
-      in_first,
-      ds.nelem,
-      ds.dtype),
-    DART_OK);
 
-  auto out_last = out_first + num_elements;
+  if (num_elements < 1) return out_first;
+
+  auto nremaining = static_cast<dash::default_size_t>(num_elements);
```
devreal: Dangerous! `dash::default_size_t` may be 32 bit wide while `num_elements` is of type `size_t`.

dash/include/dash/algorithm/Copy.h
```diff
@@ -493,17 +493,42 @@ GlobOutputIt copy_impl(
...
+  auto pattern = out_first.pattern();
+
+  while (nremaining) {
+    // global index to local unit and index
+    auto local_pos = pattern.local(out_first.pos());
+    // number of elements in unit
+    auto local_size = pattern.local_extents(team_unit_t{local_pos.unit});
+
+    auto ncopy = std::min(
+        std::initializer_list<dash::default_size_t>{local_size[0],
```
devreal: See above.

dash/include/dash/algorithm/Copy.h
```diff
@@ -493,17 +493,42 @@ GlobOutputIt copy_impl(
...
+        nremaining});
+
+    dart_storage_t ds = dash::dart_storage<ValueType>(ncopy);
+
+    DASH_ASSERT_RETURNS(
+        dart_put_blocking(
```
devreal: Why not use `dart_put` and a final `dart_flush` after the loop?

devreal (changes requested):
Comments inline.

dash/include/dash/algorithm/Copy.h
rkowalewski: `std::distance` cannot return `size_t` because it may be negative. And then I check explicitly for a non-negative number.
Apart from that I would expect that `dash::default_size_t` is the same as `size_t` if available on the underlying platform. Is there any argument against this?

dash/include/dash/algorithm/Copy.h
rkowalewski: Of course you are right, thanks. I will do this. Did not think too much about this, I simply extended the already existing code.

rkowalewski (commented):
@devreal I have considered your [comment](https://github.com/dash-project/dash/pull/347#discussion_r108153267) and modified it appropriately.

After having a more detailed look into it I noticed that `copy_async` was not modified according to the changes in the blocking variant. The approach is now a bit more elegant in my opinion. `dash::copy` on local ranges is now `dash::copy_async` on local ranges followed by an immediate wait on the returned future.

As a consequence we do temporarily not support `dash::GlobPtr` as input to `dash::copy` because we need the pattern information in `GlobIter`. This was a discussion in #246. However it is supported again after merging #345 which enables pointer arithmetics on `dash::GlobPtr` across units.

dash/include/dash/algorithm/Copy.h
devreal: You are right, it is `ssize_t`. The case I had in mind is the following: the user might decides to compile with the flag `DASH_ENABLE_DEFAULT_INDEX_TYPE_INT` (for whatever reason) and then copy more than 4GB into the array, which causes `num_elements` to be truncated into `nremaining`. He might even pass `size_t` as template parameter `IndexType` to the global data structure so it's a valid operation. I think using the pattern's index-type is the safest bet. But that is certainly a highly pathologic case :)

fuchsto:
Please wait until I evaluates effects on performance before merging.

fuchsto:
Support for strided / chunked local ranges has been requested in issue #116.

fuchsto:
Ah, now I see your changes.

No, the implementation of `dash::copy`, especially in this respect, is not by accident. Isn't it exactly these kind of cheesy shortcuts we hate about some MPI implementations?

Some time ago I gave a talk on a DASH project meeting on `dash::copy` and why the blocking variant is not implemented as `dash::async_copy + flush`. The overhead is rather significant. The non-blocking variant creates `dash::Future` objects with lambdas, request handles are stored in a `std::vector` which leads to additional heap allocation, etc.
The non-blocking `dash::copy_async` (soon to be provided via launch policy) is only beneficial for reasonably high work loads for latency hiding. The `dash::copy` function interface has a dedicated benchmark suite that has been used to evaluate these tradeoffs.

In addition, the blocking algorithms in DASH should rely on specialized blocking operations provided by DART because these again can exploit the communication routines for this specific purpose - even if these are currently nearly identical.

Issue #106, opened 2016-11-15
Update Doxygen


fuchsto:
Everyone (yes, you @dash-project/developers):

- Read the DASH API at https://codedocs.xyz/dash-project/dash/
- Complete it.

fuchsto:
@devreal Great effort, thank you a lot!

JosefSchaeffer:
Created a branch `sup-106-dash-doxygen` for doxygen documentation.

fuchsto:
@dash-project/developers 

**IMPORTANT**
Currently there is a bug in codedocs which prevents updates of the generated HTML files in browser cache.

To review doxygen output on https://codedocs.xyz/dash-project/dash either

- clear your browser cache (via browser settings, `<CTRL>+<F5>` does not suffice) or
- use incognito mode (starts with empty cache) (`<CTRL>+<SHIFT>+N` in chromium`)

fuchsto:
@dash-project/developers 
Examples are now nicely included in Doxygen.
For reference:
- http://dash-api.coreglit.ch/dash-0.3.0 (Doxygen generated on my server)
- https://codedocs.xyz/dash-project/dash/ (automatically generated by codedocs from latest `development` state)

Please help documenting all examples in `dash/examples`, just use the `\example` tag as in `dash/examples/ex.02.array/main.cpp`.

fmoessbauer:
@dash-project/developers 

Until now, there is not a single commit in `sup-106-dash-doxygen`. As discussed in the meeting, the documentation will be written in branch `doc-106-dash-doxygen` to bypass CI.

Please help us to complete the API documentation. Seriously!

fuchsto:
We use the following style for documentation:

~~~c++
/**
 * Multi-line comment opened with double-star /**.
 * First paragraph is brief description,
 *
 * Paragraphs are separated by empty comment lines. This is
 * the second paragraph and is not included in the brief
 * description.
 */
void wombatify(
   /// Single-line comments opened with *three* forward-slashes.
   /// This is a convenient way to describe function parameters.
   int     num_wombats,
   double  wombat_factor  ///< Postfix-style with <, comment placed *after* described parameter
) {
  // Any comment here is ignored, no matter how it is formatted.
}
~~~

ddiefenthaler:
Thanks for the introduction.

Is `/// parameter documentation` in any way preferred over `\param` or `@param`?

fuchsto:
[...]

PR #351, opened 2017-03-30
Document examples

devreal:
This PR adds the `\example` tag to most of the examples, as requested in #106. 

Additionally, it fixes the return values in `dash::any_of` and `dash::all_of` (`bool` instead of iterator) and fixes some glitches in the documentation across the project. 

Please check the documentation (as I haven't written the code it might be incorrect/incomplete) and feel free to extend it. 

@fuchsto Should the benchmarks be marked as `\example` as well? I have a patch for that but have not committed it yet.


fuchsto:
@devreal Thank you for taking care of this rather boring task! Benchmarks will move to `dash-project/dash-bench` so no need to document them as examples for now.

fuchsto (approved):

Issue #292, opened 2017-02-18
ThreadsafetyTest: Problems with OpenMPI 


devreal:
I am seeing issues on my local machine when running the `ThreadsafetyTest`s under OpenMPI. The used compiler is GCC 5.3.0, both in Debug and Release mode. 

__OpenMPI 1.10.5__
I am experiencing hangs in `ConcurrentPutGet`, where one or multiple threads are stuck in `MPI_Win_flush` called by `dart_put_blocking`:

```
Thread 4 (Thread 0x7fffe27a8700 (LWP 27525)):
#0  0x00007ffff62b1c47 in sched_yield () at ../sysdeps/unix/syscall-template.S:84
#1  0x00007fffebdfb715 in ompi_osc_pt2pt_flush_lock () from /home/joseph/opt/openmpi-1.10.5/lib/openmpi/mca_osc_pt2pt.so
#2  0x00007ffff752ebe4 in PMPI_Win_flush () from /home/joseph/opt/openmpi-1.10.5/lib/libmpi.so.12
#3  0x000000000095c23b in dart_put_blocking (gptr=..., src=0x7fffe27a7c44, nelem=1, dtype=DART_TYPE_INT) at /home/joseph/src/dash/dash/dart-impl/mpi/src/dart_communication.c:759
#4  0x00000000007cc2a3 in dash::GlobRef<int>::operator= (this=0x7fffe27a7cf0, val=2) at /home/joseph/src/dash/dash/dash/include/dash/GlobRef.h:201
#5  0x000000000089ae16 in dash::GlobRef<int>::operator= (this=0x7fffe27a7cf0, other=...) at /home/joseph/src/dash/dash/dash/include/dash/GlobRef.h:112
#6  0x0000000000898731 in ThreadsafetyTest_ConcurrentPutGet_Test::TestBody () at /home/joseph/src/dash/dash/dash/test/ThreadsafetyTest.cc:74
#7  0x00007ffff67bee46 in gomp_thread_start (xdata=<optimized out>) at ../../../src/libgomp/team.c:119
#8  0x00007ffff79b96ba in start_thread (arg=0x7fffe27a8700) at pthread_create.c:333
#9  0x00007ffff62ce82d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
```
For some reason, the test seems to pass on CircleCI. It seems to be a progress issue and I seem to remember that @fuchsto had investigated something along those lines in OpenMPI. Did you file a bug report back then?

[...]

PR #354, opened 2017-04-01
Fixes incorrect team in Pattern/Array by calling the wrong ctor unintended

### LINKED ISSUE #353: dash::Array storing the wrong team?
I am still debugging #292 (with increasing desperation). I found the following strange behavior

```
TEST_F(ArrayTest, TeamSplit)
{
  auto & team_all = dash::Team::All();
  auto   ext_x    = team_all.size();

  if(team_all.size() < 2){
    SKIP_TEST();
  }
  if(!team_all.is_leaf()){
    LOG_MESSAGE("team is already splitted. Skip test");
    SKIP_TEST();
  }

  dash::barrier();

  LOG_MESSAGE("Splitting team ...");
  auto & myteam = team_all.split(2);
  LOG_MESSAGE("... Team split finished");

  dash::Array<double> array_a(ext_x, myteam);
  ASSERT_EQ_U(array_a.team(), myteam); // << this test fails!

  array_a.barrier();
  // Check if array is allocated
  ASSERT_NE_U(array_a.lbegin(), nullptr);
  team_all.barrier();
}
```

I added the check `ASSERT_EQ_U(array_a.team(), myteam)`, which surprisingly fails. My understanding is that `array.team()` should yield the team the array was created with so the check should  succeed. Looking through the code, I don't see what could go wrong there. Ideas?

UPDATE: The error I get is

```
[  ERROR   ] [UNIT 0] in /home/joseph/src/dash/dash/dash/test/container/ArrayTest.cc:225
      Expected: array_a.team()
      Which is: dash::Team(0)
To be equal to: myteam
      Which is: dash::Team(1.0)
```

ddiefenthaler:
This pull request reverts the changes from PR #304 and solves it in a different way.

With the removal of the constructors from `BlockPattern` it got incompatible to the `CSRPattern`,
as the constructor of the `CSRPattern` differs in the signature.

To fix the ambiguity in the constructors this pull request removes some of the default paramters.
See also `TilePattern` for example, it is done there in the same way. [Parameter of `TilePattern` ctor](https://github.com/dash-project/dash/blob/development/dash/include/dash/pattern/TilePattern.h#L260#L265)

Due to the missing constructor the wrong constructor got called from `dash::Array` which lead to always using `dash::Team::All()` as the team of the array. This behaviour is observed in issue #353.

Fixes #353 and a part #292 (`ThreadsafetyTest.ConcurrentAlgorithm`).

### DIFF
--- a/dash/include/dash/pattern/BlockPattern.h
+++ b/dash/include/dash/pattern/BlockPattern.h
@@ -108,7 +108,6 @@ public:
   } local_coords_t;
 
 private:
-  PatternArguments_t          _arguments;
   /// Distribution type (BLOCKED, CYCLIC, BLOCKCYCLIC, TILE or NONE) of
   /// all dimensions. Defaults to BLOCKED in first, and NONE in higher
   /// dimensions
@@ -187,30 +186,7 @@ public:
     /// elements) in every dimension followed by optional distribution
     /// types.
     Args && ... args)
-  : _arguments(arg, args...),
-    _distspec(_arguments.distspec()),
-    _team(&_arguments.team()),
-    _teamspec(_arguments.teamspec()),
-    _nunits(_teamspec.size()),
-    _memory_layout(_arguments.sizespec().extents()),
-    _blocksize_spec(initialize_blocksizespec(
-        _arguments.sizespec(),
-        _distspec,
-        _teamspec)),
-    _blockspec(initialize_blockspec(
-        _arguments.sizespec(),
-        _distspec,
-        _blocksize_spec,
-        _teamspec)),
-    _local_memory_layout(
-        initialize_local_extents(_team->myid())),
-    _local_blockspec(initialize_local_blockspec(
-        _blockspec,
-        _blocksize_spec,
-        _distspec,
-        _teamspec,
-        _local_memory_layout)),
-    _local_capacity(initialize_local_capacity())
+  : BlockPattern(PatternArguments_t(arg, args...))
   {
     DASH_LOG_TRACE("BlockPattern()", "Constructor with argument list");
     initialize_local_range();
@@ -1425,6 +1401,33 @@ public:
   }
 
 private:
+
+  BlockPattern(const PatternArguments_t & arguments)
+  :  _distspec(arguments.distspec()),
+     _team(&arguments.team()),
+     _teamspec(arguments.teamspec()),
+     _nunits(_teamspec.size()),
+     _memory_layout(arguments.sizespec().extents()),
+     _blocksize_spec(initialize_blocksizespec(
+         arguments.sizespec(),
+         _distspec,
+         _teamspec)),
+     _blockspec(initialize_blockspec(
+         arguments.sizespec(),
+         _distspec,
+         _blocksize_spec,
+         _teamspec)),
+     _local_memory_layout(
+         initialize_local_extents(_team->myid())),
+     _local_blockspec(initialize_local_blockspec(
+         _blockspec,
+         _blocksize_spec,
+         _distspec,
+         _teamspec,
+         _local_memory_layout)),
+     _local_capacity(initialize_local_capacity())
+  {}
+
   /**
    * Initialize block size specs from memory layout, team spec and
    * distribution spec.
--- a/dash/include/dash/pattern/BlockPattern.h
+++ b/dash/include/dash/pattern/BlockPattern.h
@@ -289,6 +289,78 @@ public:
   }
 
   /**
+   * Constructor, initializes a pattern from explicit instances of
+   * \c SizeSpec, \c DistributionSpec and a \c Team.
+   *
+   * Examples:
+   *
+   * \code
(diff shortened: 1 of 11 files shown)


ddiefenthaler:
@devreal Can you add your adjusted unit test from issue #353?
Thanks for that by the way, it made debugging it quite a bit simpler
than having to jump between multiple threads to figure out where the team went wrong.

dash/include/dash/pattern/BlockPattern.h
```diff
@@ -249,11 +249,10 @@ class BlockPattern
     /// Pattern size (extent, number of elements) in every dimension
     const SizeSpec_t         & sizespec,
     /// Distribution type (BLOCKED, CYCLIC, BLOCKCYCLIC, TILE or NONE) of
-    /// all dimensions. Defaults to BLOCKED in first, and NONE in higher
-    /// dimensions
```
fmoessbauer: fix documentation (currently ends with "of")

fmoessbauer (commented):
If that is everything that is necessary to fix the ambiguity this PR is great! Can you please cross-check if this still works for the Coarray. Nevertheless @fuchsto  or @devreal should have a look at the changes too.

dash/include/dash/pattern/BlockPattern.h
ddiefenthaler: Thats actually just a tricky way of git displaying it.
I fell for it too yesterday and repeated that commit.

dash/include/dash/pattern/BlockPattern.h
```diff
@@ -249,11 +249,10 @@ class BlockPattern
...
-    const DistributionSpec_t & dist     = DistributionSpec_t(),
+    /// all dimensions.
```
ddiefenthaler: @fmoessbauer Look here, there is the rest of the line.

devreal:
@ddiefenthaler Great job, thanks a lot for the quick fix! I added the check to the `TeamSplit` test case and it succeeds. Since it's not my code, let's wait for the review from @fuchsto. 

While we are at it: Skimming through the code yesterday, I noticed that the `_arguments` in `BlockPattern` are stored in the pattern object even though they are only used in the c'tor. How about creating them temporarily and forwarding them to a private constructor? After all, they are rather heavy-weight. Maybe we could also apply some `constexpr` magic there because it's essentially just forwarding arguments (haven't tried though).

fmoessbauer:
While we are at it: I fixed some other bugs in the CSR pattern as well. If you like, pull in (cherry-pick) [0609a02fd58e7](https://github.com/dash-project/dash/commit/0609a02fd58e7affa396e4f2ce8f61db1c9dd484). Btw: we definitely need concept checking on the patterns.

ddiefenthaler:
I temporarily merged with [`feat-coarray`](https://github.com/dash-project/dash/tree/68e4459bb63ac91a244ab88645a7ee9c40e4633b), it was nearly automatic possible and test ran without problems.

If desired the constructor readded in this PR could get shortend by delegating to the other constructor
(like some constructors from [Array.h](https://github.com/dash-project/dash/blob/development/dash/include/dash/Array.h#L801)).

While being at cherry picking at CSRPattern I'll also add an constructor with the signatur
`CSRPattern(SizeSpec, DistSpec, TeamSpec, Team)`  so that it is more compatible to the other patterns.

ddiefenthaler:
Delegating the constructor would look like this: 
```
  BlockPattern(
    /// Pattern size (extent, number of elements) in every dimension
    const SizeSpec_t         & sizespec,
    /// Distribution type (BLOCKED, CYCLIC, BLOCKCYCLIC, TILE or NONE) of
    /// all dimensions. Defaults to BLOCKED in first, and NONE in higher
    /// dimensions
    const DistributionSpec_t & dist = DistributionSpec_t(),
    /// Team containing units to which this pattern maps its elements
    Team                     & team = dash::Team::All())
  : BlockPattern(sizespec, dist, TeamSpec_t(_distspec, *_team), team)
  {
    DASH_LOG_TRACE("BlockPattern()", "(sizespec, dist, team)");
    DASH_LOG_TRACE("BlockPattern()", "finished delegation constructor");
  }
```
against the [current implementation](https://github.com/dash-project/dash/blob/bug-353-array-team/dash/include/dash/pattern/BlockPattern.h#L324-L360).

@devreal, @fuchsto is this preferred? This would be applicable to most patterns.

~Otherwise this pull request would be ready to merge in my opinion.~
EDIT: Sorry @devreal missed your comment on the second reading...
Of course it is not a bad idea to get rid of the `_arguments` member as it should be only needed during the construction.

ddiefenthaler:
@fuchsto Here is a patch which would include the delegating constructors:
[Changed-ctor-of-pattern-to-delegate-to-similar-ctor.patch](https://github.com/dash-project/dash/files/888419/Changed-ctor-of-pattern-to-delegate-to-similar-ctor.patch.txt)

Apply it if you like,
Otherwise this pull request should be ready for review & merge

fuchsto (approved):

PR #356, opened 2017-04-03
Check variadic pattern arguments at compile time

devreal:
This PR changes `dash::internal::PatternArguments` such that variadic pattern argument integrity is checked at compile-time (except for tile-correctness). It also enables arbitrary mixing of `SizeType`, `Distribution`, and `Team|TeamSpec` arguments, as long as the argument list starts with a `SizeType`. Most notably, it catches mixed usage of `Team` and `TeamSpec`, `Distribution` and `DistributionSpec`, as well as `SizeType` and `SizeSpec` specifications at compile-time. 

My attempt to add some `constexpr` magic (as proposed in #354) failed since this style of parsing does not seem to be possible that way as it would require overloading constructor templates. 

As an addition, this PR also introduces `DASH_ASSERT_ALWAYS`, which always asserts even if assertions are disabled. This is required to ensure pattern validity (replaced plain `assert`).


dash/include/dash/pattern/internal/PatternArguments.h
```diff
@@ -122,71 +105,91 @@ class PatternArguments {
 
 private:
   /// BlockPattern matching for extent value of type IndexType.
-  template<int count>
-  void check(SizeType extent) {
+  template<int argc_size, int argc_dist, int has_team, typename ... Args>
```
ddiefenthaler: Some documentation would be nice for those parameters.

dash/include/dash/pattern/internal/PatternArguments.h
```diff
@@ -122,71 +105,91 @@ class PatternArguments {
...
-  void check(const TeamSpec_t & teamSpec) {
+  template<int argc_size, int argc_dist, int has_team, typename ... Args>
+  void check(const TeamSpec_t & teamSpec, Args && ... args) {
+    static_assert(has_team == 0,
+        "Cannot mix Team and TeamSpec definition in variadic "
+        "pattern constructor!");
     DASH_LOG_TRACE("PatternArguments.check(teamSpec)");
-    _argc_team++;
     _teamspec   = teamSpec;
+    check<argc_size, argc_dist, -1>(std::forward<Args>(args)...);
```
ddiefenthaler: No Team is set in this case, it was probably broken before as well...

dash/include/dash/pattern/internal/PatternArguments.h
```diff
@@ -122,71 +105,91 @@ class PatternArguments {
...
+    static_assert(has_team == 0,
+        "Cannot specify Team twice in variadic "
+        "pattern constructor!");
     DASH_LOG_TRACE("PatternArguments.check(team)");
-    if (_argc_team == 0) {
-      _team     = &team;
-      _teamspec = TeamSpec_t(_distspec, team);
-    }
+    _team     = &team;
+    check<argc_size, argc_dist, 1>(std::forward<Args>(args)...);
```
ddiefenthaler: No TeamSpec is created in this case,
as it is dependend on `_distspec` it probably has to go in the final `check()` call.

dash/include/dash/pattern/internal/PatternArguments.h
```diff
@@ -122,71 +105,91 @@ class PatternArguments {
...
   }
   /// BlockPattern matching for one optional parameter specifying the 
   /// size (extents).
-  template<int count>
-  void check(const SizeSpec_t & sizeSpec) {
+  template<int argc_size, int argc_dist, int has_team, typename ... Args>
+  void check(const SizeSpec_t & sizeSpec, Args && ... args) {
```
ddiefenthaler: Perhaps it makes sense to move this function up,
so that the it is grouped with the function taking individual extents.

dash/include/dash/pattern/internal/PatternArguments.h
```diff
@@ -122,71 +105,91 @@ class PatternArguments {
...
+    static_assert(argc_size < NumDimensions, "Number of size specifier exceeds"
+        "the number of dimensions in variadic pattern constructor!");
     DASH_LOG_TRACE("PatternArguments.check(extent)", extent);
-    _argc_size++;
-    _sizespec.resize(count, extent);
+    _sizespec.resize(argc_size, extent);
+    check<argc_size+1, argc_dist, has_team>(std::forward<Args>(args)...);
   }
   /// BlockPattern matching for up to \c NumDimensions optional 
   /// parameters specifying the distribution pattern.
```
ddiefenthaler: This comment seems to be shifted.

dash/test/pattern/BlockPatternTest.cc
```diff
@@ -18,14 +18,16 @@ TEST_F(BlockPatternTest, SimpleConstructor)
   int extent_x = 21;
   int extent_y = 37;
   int extent_z = 41;
-  int size = extent_x * extent_y * extent_z;
+  int size = extent_x * extent_x * extent_z;
```
ddiefenthaler: This change was probably uninteded, as it is currently breaking the tests.

dash/test/pattern/BlockPatternTest.cc
```diff
@@ -44,6 +46,23 @@ TEST_F(BlockPatternTest, SimpleConstructor)
...
+  EXPECT_EQ(ds_blocked_z, pat_ds_t.distspec());
+  EXPECT_EQ(size, pat_ds_t.capacity());
+  EXPECT_EQ(dash::Team::All().size(), pat_ds_t.num_units());
+
+  dash::BlockPattern<3> pat_spec(
+      sspec, ds_blocked_z,
+      dash::Team::All());
+  EXPECT_EQ(ds_blocked_z, pat_ds_t.distspec());
+  EXPECT_EQ(size, pat_ds_t.capacity());
+  EXPECT_EQ(dash::Team::All().size(), pat_ds_t.num_units());
```
ddiefenthaler: The wrong patterns are checked in those `EXPECT`s.

dash/test/pattern/BlockPatternTest.cc
```diff
@@ -18,14 +18,16 @@ TEST_F(BlockPatternTest, SimpleConstructor)
...
+  int size = extent_x * extent_x * extent_z;
   // Should default to distribution BLOCKED, NONE, NONE:
   dash::BlockPattern<3> pat_default(extent_x, extent_y, extent_z);
   EXPECT_EQ(dash::DistributionSpec<3>(), pat_default.distspec());
   EXPECT_EQ(dash::Team::All(), pat_default.team());
   EXPECT_EQ(dash::Team::All().size(), pat_default.num_units());
   EXPECT_EQ(size, pat_default.capacity());
 
+  dash::SizeSpec<3> sspec(
+      extent_x, extent_x, extent_z);
```
ddiefenthaler: Probably the wrong y-extent here, too.

dash/include/dash/pattern/internal/PatternArguments.h
```diff
@@ -122,71 +105,91 @@ class PatternArguments {
...
-  template<int count>
-  void check(const TeamSpec_t & teamSpec) {
+  template<int argc_size, int argc_dist, int has_team, typename ... Args>
+  void check(const TeamSpec_t & teamSpec, Args && ... args) {
+    static_assert(has_team == 0,
+        "Cannot mix Team and TeamSpec definition in variadic "
+        "pattern constructor!");
```
ddiefenthaler: This limitation seems a little bit odd,
but I think this is depended on TeamSpec.

dash/include/dash/pattern/internal/PatternArguments.h
```diff
@@ -71,27 +67,14 @@ class PatternArguments {
...
-        "expected " << NumDimensions << ", got " << _argc_size);
-    }
-    if (_argc_dist > 0 && _argc_dist != NumDimensions) {
-      DASH_THROW(
-        dash::exception::InvalidArgument,
-        "Invalid number of dist arguments for BlockPattern(...), " <<
-        "expected " << NumDimensions << ", got " << _argc_dist);
-    }
-    check_tile_constraints();
+    check<0, 0, 0>(std::forward<Args>(args)...);
```
ddiefenthaler: Meanwhile it is probably more a parse than a check.

ddiefenthaler (commented):
Looks good, I'll leave some comments inline

dash/include/dash/pattern/internal/PatternArguments.h
```diff
@@ -121,72 +104,160 @@ class PatternArguments {
...
+    static_assert(!(ArgcTeam > 0),
+        "Cannot specify Team twice in variadic "
+        "pattern constructor!");
     DASH_LOG_TRACE("PatternArguments.check(team)");
-    if (_argc_team == 0) {
-      _team     = &team;
-      _teamspec = TeamSpec_t(_distspec, team);
-    }
+    _team     = &team;
+    _teamspec = TeamSpec_t(team);
```
ddiefenthaler: Possibly overwrites previously set TeamSpec.
Should already be handled by check in line [255](https://github.com/dash-project/dash/pull/356/files#diff-703bc11c993d6436737bb7664d6b84a0R255).

dash/include/dash/pattern/internal/PatternArguments.h
```diff
@@ -121,72 +104,160 @@ class PatternArguments {
...
+        "Incomplete size specification in "
+        "variadic pattern constructor!");
+
+    /*
+     * TODO: we have no way to statically check whether a TeamSpec was created
+     *       using the team specified by the user.
+     */
+    if (ArgcTeam && !ArgcTeamSpec) {
+      _teamspec = TeamSpec_t(_distspec, *_team);
+    }
```
ddiefenthaler, line 259: I think it is possible to return without `_team` set, how about:
```c++
if(!ArgcTeam) {
  _team = &dash::Team::All()
}
if(!ArgcTeamSpec) {
  _teamspec = TeamSpec_t(_distspec, *_team);
}
```

I'm not sure about the use cases of this constructor,
so perhaps it is also fine to return without a team.

ddiefenthaler (approved):
It looks good now and should be ready to merge.

I found one more minor thing that is TeamSpec related,
but I think TeamSpec needs its own issue/pull request, so it shouldn't be important for now.

dash/include/dash/pattern/internal/PatternArguments.h
```diff
@@ -25,7 +27,7 @@ template<
   dim_t NumDimensions,
   typename IndexType = dash::default_index_t>
 class PatternArguments {
-private:
+public:
```
fuchsto: Why public?

dash/include/dash/pattern/internal/PatternArguments.h
```diff
@@ -121,72 +104,163 @@ class PatternArguments {
...
+    typename ... Args>
+  void parse(const DistributionSpec_t & ds, Args && ... args) {
+    static_assert(ArgcDist == 0, "Cannot mix DistributionSpec and inidividual"
+        "distributions in variadic pattern constructor!");
     DASH_LOG_TRACE("PatternArguments.check(distSpec)");
-    _argc_dist += NumDimensions;
-    _distspec   = ds;
+    _distspec = ds;
+    parse<ArgcSize, -1, ArgcTeam,
+      ArgcTeam>(std::forward<Args>(args)...);
```
ddiefenthaler: Is the use of `ArgcTeam` instead of `ArgcTeamSpec` intended here?

dash/include/dash/pattern/internal/PatternArguments.h
```diff
@@ -121,72 +104,163 @@ class PatternArguments {
...
-    DASH_LOG_TRACE("PatternArguments.check(args) ",
-                   "count", count,
-                   "argc", sizeof...(Args));
-    check<count>(std::forward<T>(t));
-    if (sizeof...(Args) > 0) {
-      check_recurse<count + 1>(std::forward<Args>(args)...);
-    }
+    _distspec[ArgcDist] = ds;
+    parse<ArgcSize, ArgcDist+1, ArgcTeam,
+      ArgcTeam>(std::forward<Args>(args)...);
```
ddiefenthaler: Here too.

fuchsto (changes requested):

devreal:
Thanks for the reviews, they should be addressed.

dash/include/dash/pattern/internal/PatternArguments.h
devreal: The `dash::Team::All` is used by default if no team has been set, see `team()`. The assumption here is that it only makes sense to overwrite the default `TeamSpec` if a team was set. However, I realized that I am not sure about the relation between `DistSpec` and `TeamSpec` and whether we have to create the latter if the former was explicitely given bu the user. Is it even valid to specify both at the same time? @fuchsto can you shed some light?

fuchsto (approved):

PR #358, opened 2017-04-18
Fixes to GlobAsyncRef

### LINKED ISSUE #357: GlobAsyncRef uses blocking dart operations
According to my understanding, `dash::GlobAsyncRef` should perform non-blocking puts and gets. However a first glance at the code this is not the case: See [GlobAsyncRef.h,L190](https://github.com/dash-project/dash/blob/development/dash/include/dash/GlobAsyncRef.h#L190).

Furthermore, the implementation does not fulfill the GlobRef concept and seems to be pretty experimental.

devreal:
1. Use blocking `get` to retrieve the value.
2. Use non-blocking `put` to write values.
3. Cache values from `get` and `put`.
4. Add `flush` method to synchronize cached and global value. 

Note that a non-blocking `get` does not make sense but a non-blocking `put` is perfectly fine. The result of the blocking `get` will be cached and another `get` will only be issued after a `flush`

Fixes #357.

### DIFF
--- a/dash/include/dash/GlobAsyncRef.h
+++ b/dash/include/dash/GlobAsyncRef.h
@@ -51,18 +51,14 @@ class GlobAsyncRef
 private:
   typedef GlobAsyncRef<T>
     self_t;
-  typedef typename std::remove_const<T>::type
-    nonconst_value_type;
 
 private:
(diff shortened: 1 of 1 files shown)


fuchsto (approved):
Thanks!

PR #359, opened 2017-04-18
More fixes for GlobAsyncRef

devreal:
Extends the interface of `dash::GlobAsyncRef` to match the interface of `dash::GlobRef`. Also fixes constness of `GlobRef` and adds checks for return values of DART functions. 

@fmoessbauer please review and let me know if any functionality is missing. 

See #357

### DIFF
--- a/dash/include/dash/GlobAsyncRef.h
+++ b/dash/include/dash/GlobAsyncRef.h
@@ -52,6 +52,12 @@ private:
   typedef GlobAsyncRef<T>
     self_t;
 
+  typedef typename std::remove_const<T>::type
+    nonconst_value_type;
+
+  typedef typename std::add_const<T>::type
+    const_value_type;
+
 private:
   /// Pointer to referenced element in global memory
   dart_gptr_t  _gptr        = DART_GPTR_NULL;
@@ -71,7 +77,7 @@ public:
    */
   explicit constexpr GlobAsyncRef(
     /// Pointer to referenced object in local memory
-    T * lptr)
+    nonconst_value_type * lptr)
   : _value(*lptr),
     _lptr(lptr),
     _is_local(true),
@@ -134,7 +140,7 @@ public:
   /**
    * Whether the referenced element is located in local memory.
    */
-  inline bool is_local() const
+  inline bool is_local() const noexcept
   {
     return _is_local;
   }
@@ -142,7 +148,7 @@ public:
   /**
    * Conversion operator to referenced element value.
    */
-  operator T() const
+  operator nonconst_value_type() const
   {
     DASH_LOG_TRACE_VAR("GlobAsyncRef.T()", _gptr);
     if (!_has_value) {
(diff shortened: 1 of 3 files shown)


dash/include/dash/GlobAsyncRef.h
```diff
@@ -258,16 +360,55 @@ class GlobAsyncRef
...
     --val;
     operator=(val);
     return result;
   }
 
+
+  /**
+   * Multiplication operator.
+   */
+  GlobRef<T> & operator*=(const_value_type& ref) {
```
fmoessbauer: Should be `GlobAsyncRef` ?

dash/include/dash/GlobAsyncRef.h
```diff
@@ -258,16 +360,55 @@ class GlobAsyncRef
...
+    nonconst_value_type val = operator nonconst_value_type();
+    val   *= ref;
+    operator=(val);
+    return *this;
+  }
+
+  /**
+   * Division operator.
+   */
+  GlobRef<T> & operator/=(const_value_type& ref) {
```
fmoessbauer: Should be `GlobAsyncRef` ?

dash/include/dash/GlobAsyncRef.h
```diff
@@ -258,16 +360,55 @@ class GlobAsyncRef
...
+    nonconst_value_type val = operator nonconst_value_type();
+    val   /= ref;
+    operator=(val);
+    return *this;
+  }
+
+  /**
+   * Binary XOR operator.
+   */
+  GlobRef<T> & operator^=(const_value_type& ref) {
```
fmoessbauer: Same here

fmoessbauer (commented):
Hi, thanks for fixing this. Just some minor points that should be addressed before merging:
- There are no unit tests for the arith. operators. 
- see inline comments

dash/include/dash/GlobAsyncRef.h
devreal: Good catch, will fix it.

devreal:
Thanks for the review. I discovered another problem here: the postfix increment operator is broken for both `GlobRef` and `GlobAsyncRef` because neither of them are copy-constructible. Besides that: what is the reasoning behind returning a *reference* to a value before that value gets incremented? The only solution I can think of on the spot is to have the postfix operator return a value of type `T` instead of `GlobRef<T>`. However, that would break use cases like `(gref++)++`, where the second increment would not be propagated to the global memory. Alternatively, we delete the postfix increment to avoid incorrect usage. @fuchsto please comment.

fmoessbauer:
Personally I would remove the postfix increment operator on all GlobRefs, even if we have to fix the examples and tests. This is also something to discuss tomorrow. 

Related to this topic a short google search took me to this quite interesting discussion: https://github.com/ericniebler/stl2/issues/137

fmoessbauer (approved):

PR #360, opened 2017-04-19
Bug globref conversion

fmoessbauer:
- Fixed conversion between `GlobRef` and `GlobAsyncRef` and added unit test.
- added missing `member` methods


dash/include/dash/GlobAsyncRef.h
```diff
@@ -157,6 +157,30 @@ class GlobAsyncRef
...
+   * specified offset
+   */
+  template<typename MEMTYPE>
+  GlobAsyncRef<MEMTYPE> member(size_t offs) const {
+    dart_gptr_t dartptr = _gptr;
+    DASH_ASSERT_RETURNS(
+      dart_gptr_incaddr(&dartptr, offs),
+      DART_OK);
+    GlobConstPtr<MEMTYPE> gptr(dartptr);
+    return GlobAsyncRef<MEMTYPE>(gptr);
```
devreal: This would trigger a new `get` even though the current `GlobAsyncRef` may already have cached the data, potentially leading to incoherency. How about a private ctor `GlobAsyncRef(dart_gptr_t, MEMTYPE*)`?

Also: the intermediate `GlobConstPtr` is not required, right?

devreal (changes requested):
Thanks for adding the `member` functions, I've been meaning to do that as well.

devreal (approved):

PR #361, opened 2017-04-20
Ensure progress in dart_flush*

devreal:
As discovered by @fmoessbauer, progress is not guaranteed if units issue only DART atomic and flush operations. To ensure progress, we can issue an `MPI_Iprobe` in the `dart_flush*` operations. This will ensure progress on all one-sided operations since they all require a flush at some point. Tested with both MPICH and OpenMPI. 

Also reduces the flush operation in `dash::GlobAtomicRef` to the target only.


rkowalewski:
What is the scenario for this pull request. Is there at least a short example describing this issue?

devreal:
The use-case is as follows: One unit uses `dash::GlobAtomicRef::add(1)` to send a signal to unit 1, which uses `dash::GlobAtomicRef::get()` to wait for the signal. Both operations issue DART atomic operations and `dart_flush`. 

This is part of the `CoEvent` `post`/`wait` implemented by @fmoessbauer. I will add a test case for it.

devreal:
I added a test case and verified that it hangs in `development`. The problem seems to be that the unit waiting for the signal (unit 0 here) does not issue any remote communication, which is why the other units do not see progress. If all units  issue a signal and wait for a signal from another unit, the test succeeds (as in the first version I committed). Nevertheless, it's a valid use-case to just wait for a signal to arrive locally without remote communication.

rkowalewski:
This is a specific solution for exactly this scenario, isn't it? In this case it works because the target calls `dart_flush` which in turn calls `MPI_Iprobe`. However, it may not work if the target does not call any DART routine or am I missing something?

What about the general idea to provide a `dart_sync_producer_consumer` interface to synchronize two units. This pattern is useful in many use cases. Then we can provide specific implementations because it may be better to use message-passing instead of one-sided communication for some MPI libraries. Moreover it causes less overhead in DASH as we could implement the relevant data structures in DART. The mysterious `I_probe` calls in `dart_flush` are then obsolete while we can keep it anyway.

devreal:
Well, if the user does not call DART functions then progress is not guaranteed. And this fixes any situation where the user does not trigger remote communication, e.g., local-only access through DART. 

Any one-sided communication does not block until flush is called. On the other hand, flush has to be called to complete the communication, so it's an integral part of our communication model. At the same time, collective operations guarantee progress. So in order to ensure progress, the user either has to call a collective operation (such das `dart_barrier`) or a flush. I would refrain from adding another call but we can document that the flush operations cause progress now. The user just has to call flush once in a while, which he is likely to do with any communication he performs. 

A separate `dart_sync_producer_consumer` interface does not solve the use-case presented here, which is a valid use-case after all.

rkowalewski (commented):
I am still not completely convinced since it is more a hack. Anyway, it fixes an issue and we will maybe revisit this topic if another exotic progress use case occurs. :)

devreal:
@rkowalewski you are right, it's kinda hacky. However, it bridges a gap between the progress rules of the MPI standard and the progress we have to ensure in our programming model. If anyone has a better idea please come forward :)

devreal:
@rkowalewski What's your opinion on this one now? Can we merge this PR? (you requested to do a review ;))

rkowalewski:
Well I have reviewed and approved this pull request with a not really good feeling as you can see in my previous message. But since there is no better solution right now...let us do it.

rkowalewski (approved):

PR #364, opened 2017-04-20
fixed copy and move semantics of HDF5 streams

fmoessbauer:


devreal (approved):
Looks good from my point of view.

Issue #368, opened 2017-04-25
Fix Copy / Move / Rule of five of Matrix


fmoessbauer:
Copy and move semantics are currently broken on (at least) `dash::Matrix`. This is due to the missing ctors and assignment operators.

fuchsto:
They are not broken, they are deleted. A `dash::Matrix` may not be copied as it's a typically huge object. Moving should work, though.

fuchsto:
Your changes in #369 look okay in this respect, though. Just don't implement assignment and copy ;)

fmoessbauer:
They were missing... Anyway, of course copy is deleted, but move should be perfectly well. I guess we have to fix that on other places in dash as well. I just stumbled upon this by making my `dash::Coarray` fulfill the container concept ;)

fmoessbauer:
Fixed in #369

PR #369, opened 2017-04-25
Fix move semantics of dash::Matrix

fmoessbauer:
Addresses #368

### DIFF
--- a/dash/include/dash/matrix/internal/Matrix-inl.h
+++ b/dash/include/dash/matrix/internal/Matrix-inl.h
@@ -94,6 +94,8 @@ inline Matrix<T, NumDim, IndexT, PatternT>
 {
     // do not free other globmem
     other._glob_mem = nullptr;
+    other._lbegin   = nullptr;
+    other._lend     = nullptr;
     DASH_LOG_TRACE("Matrix()", "Move-Constructed");
 }
 
@@ -124,7 +126,10 @@ Matrix<T, NumDim, IndexT, PatternT>
 
   // do not free other globmem
   other._glob_mem = nullptr;
+  other._lbegin   = nullptr;
+  other._lend     = nullptr;
   DASH_LOG_TRACE("Matrix.operator=(&&)", "Move-Assigned");
+  return *this;
 };
 
 template <typename T, dim_t NumDim, typename IndexT, class PatternT>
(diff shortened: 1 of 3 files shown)


dash/include/dash/matrix/internal/Matrix-inl.h
```diff
@@ -80,13 +80,54 @@ ::Matrix(
...
+  _lcapacity(other._lcapacity),
+  _begin(other._begin),
+  _pattern(other._pattern),
+  _glob_mem(other._glob_mem),
+  _lbegin(other._lbegin),
+  _lend(other._lend),
+  _ref(other._ref)
+{
+    // do not free other globmem
+    other._glob_mem = nullptr;
```
devreal, line 96: To be on the safe side, I'd also reset `_lbegin` and `_lend` etc in `other`, otherwise those pointer will dangle.

devreal (changes requested):

fmoessbauer:
@devreal Your suggestions are implemented.

dash/include/dash/matrix/internal/Matrix-inl.h
fuchsto: That's okayish as they are not deallocated. There is no (and must not be) ownership on `_lbegin`, `_lend`, so you can safe those cycles.

fuchsto: ... in general: To determine which pointers need to be nulled, look at the destructor. Only the destructor of a moved instance may be called after moving, so dangling is not an issue if the destructor does not depend on the pointers.

fmoessbauer: There are arguments for both sides. IMO nulling the pointers is useful to avoid accidental usage of the dead matrix object. With nulling the user gets an segfault upon first access instead of working on the wrong matrix. I prefer to leave the current implementation.

devreal (approved):

dash/include/dash/matrix/internal/Matrix-inl.h
fuchsto: Oh yes, considering the "overhead" of null-ing a pointer, there's no need for discussion. Just wanted to note that there is no risk of dangling pointers.

Issue #365, opened 2017-04-20
GlobAsyncRef flush via container.flush() does not work


fmoessbauer:
According to the documentation of `GlobAsyncRef` it is possible use `container.flush()` to publish all changes. However this does unfortunately not work:

```cpp
dash::Array<int> data(dash::size());
*(array.lbegin()) = 0;
array.barrier();

mutex.lock();
int tmp = array.async[0];
arrary.async[0] = tmp + 1;
array.flush(); // using gref.flush() works
mutex.unlock();
array.barrier();
// postcondition: array[0] == dash::size();
```

This has to be investigated further. My guess is that the problem is the value comparison in the assignment operator. Furthermore the async stuff is not threadsafe in any case.

Maybe we should document this as a known limitation in 0.3.0, as fixing this seems quite elaborate.

devreal:
> `arrary.async[dash::myid()] = tmp + 1;`

Shouldn't this be 
```
arrary.async[0] = tmp + 1;
            ^^^
```
?

The following (adapted) test succeeds:

```
  dash::Array<int> array(dash::size());
  *(array.lbegin()) = 0;
  array.barrier();

  dash::Mutex mutex;

  mutex.lock();
  int tmp = array.async[0];
  array.async[0] = tmp + 1;
  array.flush(); // using gref.flush() works
  mutex.unlock();
  array.barrier();

  // postcondition: array[0] == dash::size();
  ASSERT_EQ_U(dash::size(), array[0]);
```

> Furthermore the async stuff is not threadsafe in any case.

Mhh, as long as two threads do not share a `GlobAsyncRef` object it is thread-safe. Of course, you get race-conditions if two threads write to the same global memory through different references but that is not in our hands. What exactly do you have in mind?

The general problem is that the `GlobAsyncRef` is decoupled from the container so a flush will not force an update on pending async references, they become stale without anyone noticing. That should be documented, yes.

fmoessbauer:
You are right with `^^^` ;) Have you tried with nastympi? I first had the errors in CI and then was only able to generate them using nastympi.

devreal:
NastyMPI dies with a Segfault on my machine: 

```
[----------] 53 tests will be run.
[----------] run 1 out of 5 tests from GlobAsyncRefTest
[  RUN     ] GlobAsyncRefTest.ContainerFlush
[= -1 LOG =]       GlobAsyncRefTest.h :  13 | >>> Test suite: GlobAsyncRefTest 
[...]

PR #371, opened 2017-04-26
[WIP] Adjust propagation semantics of GlobAsyncRef

devreal:
Adds the test case from #365 and fixes the documentation of `GlobAsyncRef`. Note the change in semantics: changes to a globref are **not** propagated to the container unless a flush (either on the `GlobAsyncRef` or the container) is issued. The following is not legal (and probably never will be):

```
arr.async[0] = 123;
assert(arr.async[0] == 123); // !!! this is a different GlobAsyncRef object with empty cache
```

Changes are only cached in the `GlobAsyncRef` instance they were made on:
```

auto gar = arr.async[0];
gar = 123;
assert(gar == 123);
```

To make the first case valid, we'd have to use a blocking put, which essentially takes the async out of `GlobAsyncRef`.


dash/test/iterator/GlobAsyncRefTest.cc
```diff
@@ -200,3 +201,44 @@ TEST_F(GlobAsyncRefTest, RefOfStruct)
...
+
+  // async writes to array.async
+  for (int i = 0; i < num_iter; ++i) {
+    mutex.lock();
+    int tmp = array.async[0];
+    for (int j = 0; j < num_iter; ++j) {
+      tmp += 1;
+      array.async[0] = tmp;
+    }
+    array.flush(); // using gref.flush() works
```
fmoessbauer: Does this now work? What about the inline comment?

devreal: That was a leftover, removed it.

dash/include/dash/GlobAsyncRef.h
```diff
@@ -338,54 +334,22 @@ class GlobAsyncRef
    * types which are comparable
    */
   template<typename __T = T>
```
fuchsto: There is no need for a function template any more.

fuchsto (changes requested):
Some more fundamental considerations will follow.

dash/include/dash/GlobAsyncRef.h
devreal: Fixed.

rkowalewski:
While looking at the `GlobAsyncRef` implementation I have noticed that we can construct one by passing a native local pointer. What is the purpose behind this since I can pass any native pointer which does not necessarily refer to global memory. A `GlobRef` to a detached (non-global) memory region makes no sense or am I missing something?

Moreover the semantics are kind of confusing now. A `GlobRef`, whether synchronous or asynchronous,  is an alias to a particular memory address (array offset) in the global memory space. Now we have the following inconsistent invariants:

~~~cpp
auto alias_1 = array.async[0];
auto alias_2 = array.async[0];
alias_2 = 100;
assert(alias_1 == alias_2) //true which makes sense since it refers to the same dart_gptr_t
// However,
assert(static_cast<int>(alias_1) == alias_2) //...is false due to internal caching...

/* Another strange inconsistency. */
alias_2.flush(); //Let us flush the previous update
auto val_1 = alias_1.get();
assert(val_1 == 100); //OK since the previous operation is blocking
//However if we do the following...
int val_1;
alias_1.get(val_1);
assert(val_1 == 100); //false (What??) --> the previous operation is non-blocking.
~~~

In general, two references to the same underlying address (whatever address space we have) should always evaluate to the same value like in common C++. At least from the perspective of one process / thread.

The internal caching may cause various undefined behavior scenarios in more complex applications.  Caching is simply not the scope of a reference. The internal state machine with all these `has_value`, `value`, `local_ptr` and `gptr` looks a bit hacky and difficult to debug.
For caching we should provide a separate caching component which is aware of all references to a single address. There will follow a more detailed description in a separate concept.

If am right, the original main purpose of `GlobAsyncRef`was to simply provide non-blocking communication without a flush to enable computation-communication overlap. Now the implementation looks more like a mix of non-blocking communication and a future with additional internal caching. All of these are orthogonal concepts for difference purposes. 

Another question remains about the differences of Puts and Gets. Puts are now semantically non-blocking while Gets are blocking in `GlobAsyncRef` if we explicitly cast it to the `value_type`. This is kind of special and not always desireable from the programmer's point of view, is it?

devreal:
@rkowalewski I totally agree, the semantics are kinda messy and I am absolutely not happy with it. The problem boils down to the limited scope of the cache, which is only valid for a single `GlobAsyncRef` *object*, not the address. 

> In general, two references to the same underlying address (whatever address space we have) should always evaluate to the same value like in common C++. At least from the perspective of one process / thread.

They are not references in the classical sense as write operations are *asynchronous*. So any change in the value is guaranteed to complete *only after* a flush, no matter who is looking at the value (local or remote units). I agree that this breaks the common notion of cache-coherence and might be confusing but the only solution I can think of right now is to issue a flush before the `get` to ensure async `put` operations are complete and thus to provide a coherent view.

> For caching we should provide a separate caching component which is aware of all references to a single address. There will follow a more detailed description in a separate concept.

I thought about that but I'm afraid the overhead will blow up significantly if the user issues many `async` `put`s. We'd have to keep track of each address touched in between two flushes. That sounds like a performance nightmare. 

> If am right, the original main purpose of GlobAsyncRefwas to simply provide non-blocking communication without a flush to enable computation-communication overlap.

The original implementation was broken though, as it used *non-blocking gets* and *blocking puts*. 

> Another question remains about the differences of Puts and Gets. Puts are now semantically non-blocking while Gets are blocking in GlobAsyncRef if we explicitly cast it to the value_type. This is kind of special and not always desireable from the programmer's point of view, is it?

AFAICS, this is the only sensible option. We cannot return a reasonable value of the target type with a non-blocking `get`, can we? We could return a `dash::future`, though (I think this is what @fuchsto suggested earlier). Right now, `get()` is blocking while `get(nonconst_value_type *tptr)` is non-blocking, as documented. 

The cache value is necessary for the non-blocking `put` to prevent the source from going out of scope. Now, thinking about it this is broken as well: assuming the user discards the `GlobAsyncRef` object immediately, the cached value might go out of scope: 

```
array.async[0] = 1; // `GlobAsyncRef` object will be destroyed immediately after the assignment
```

We'd need at least local completion in the `dtor` of `GlobAsyncRef` before the cache value goes out of scope.

rkowalewski:
> I thought about that but I'm afraid the overhead will blow up significantly if the user issues many async puts. We'd have to keep track of each address touched in between two flushes. That sounds like a performance nightmare.

Performance is a different story and something we can (and should) solve as we are HPC scientists. :) It is of course not easy but IMHO, it is more important to clarify the semantics as the best performance does not provide any advantage if it is difficult for users to understand the underlying behavior.

> AFAICS, this is the only sensible option. We cannot return a reasonable value of the target type with a non-blocking get, can we? We could return a dash::future, though (I think this is what @fuchsto suggested earlier). Right now, get() is blocking while get(nonconst_value_type *tptr) is non-blocking, as documented.

We can return an undefined value which is valid if we explicitly specify this and which is common practice as we know. However, I agree that a future is better in this case. With a future it becomes semantically clear that the underlying operation may be asynchronous and we provide a handle to test or even wait for the pending operation. That is actually what the user "wants" to have in my opinion.

> We'd need at least local completion in the dtor of GlobAsyncRef before the cache value goes out of scope.

 Now I can see that there may be many many temporaries in a complex program (if the programmer is not smart enough) which flood the network and, simply stated, only the last written value is really of interest (revisiting the earlier mentioned performance problem). With futures this will not be an issue as we transfer the responsibility to the user to complete a pending operation.

devreal:
> Performance is a different story and something we can (and should) solve as we are HPC scientists. :) It is of course not easy but IMHO, it is more important to clarify the semantics as the best performance does not provide any advantage if it is difficult for users to understand the underlying behavior.

IMO we should not offer semantics that are nice to have but will never perform well or have the notion of "fix performance issues later".

> We can return an undefined value which is valid if we explicitly specify this and which is common practice as we know.

That is the worst solution as that will cause so much headache (ranging from "why can't I store value X in my array" "well, X is our undefined value" to "why do I see an undefined value on *some* operations and not on others" "well, because the transfer sometimes finishes faster..."). Returning a future for an async `get` would be the cleanest way. 

> Now I can see that there may be many many temporaries in a complex program (if the programmer is not smart enough) which flood the network and, simply stated, only the last written value is really of interest (revisiting the earlier mentioned performance problem).

Consider something like:

```
dash::Array<int> array(N);
if (dash::myid() == 0) {
  for (int i = 0; i < N; ++i) {
    if (i % 2) {
      array.async[i] = compute(i);
    }
  }
}
array.flush();
```
It's probably one of the simpler (yet maybe hypothetical) use-cases but it's perfectly legal and not supported right now. For large N, the cache will become problematic. Ideally, we'd have something like `dart_put_local_complete` (or just `dart_put_local`) which ensures local completion (e.g., using `MPI_Rput`) such that we can get rid of the cached value. Alternatively, `dart_put` + `dart_flush_local` will do as well but might be more `flush` than actually required. 

Asynchronous put operations then have *eventual consistency*, i.e., the updates are visible to everyone after the flush, including the current unit. Before that, you may or may not see the changes as they are "in flight". Et voila, no cache needed and the semantics are clear. 

> With futures this will not be an issue as we transfer the responsibility to the user to complete a pending operation.

It's exactly the same issue for `put` operations as we have right now. If the future is destroyed before the user calls a `flush` things go downstream from there. Plus, how do you return a future from an assignment? (I guess it's technically possible but will buy you years of programming hell in an instant)

rkowalewski:
> It's probably one of the simpler (yet maybe hypothetical) use-cases but it's perfectly legal and not supported right now.

You are right and this is something which should work as it is kind of basic use case for asynchronous updates. Here we should enable at least local completion. But then we do not really need the internal caching stuff as you mentioned.

> For large N, the cache will become problematic. Ideally, we'd have something like dart_put_local_complete (or just dart_put_local) which ensures local completion (e.g., using MPI_Rput) such that we can get rid of the cached value. Alternatively, dart_put + dart_flush_local will do as well but might be more flush than actually required.

Well, now we are a little bit off-topic and this is way too complex to discuss here. Actually the current implementation is not really a cache it is just some buffer in order to flush it later. I agree with your proposed solution of ensuring local completion since it solves the headache of buffering something on particular `GlobAsyncRef` which does not enable other `GlobAsyncRefs` to read prior local updates. Even in eventual consistency models I have often read about the "read your writes" guarantee while I am not claiming that this is generally true. But most users at least want to have a consistent "local" view.

fmoessbauer:
This PR has become quite a large discussion on the memory consistency of DASH in general. During my research on the CoArray implementation I often read the following which I try to summarize here:
- puts are non-blocking, reads are blocking per default.
 - after a put (write), the memory location is in an undefined state (for all units except the unit that performed the update), hence it is not allowed to read that location until a flush is executed (in fact is is allowed, but might return the old or the new value). Since we support fine granular flushes, the scope of the flush should determine for which units the address is in an defined state again. This is important for p2p synchronization.
- GlobRef should provide a method for blocking put with local and remote completion.
- GlobRef should provide a method for non-blocking reads.
- pure non blocking communication is supported by wrapping a GlobRef in a future.

The [CoArray C++ paper, P 9](http://www.pgas2013.org.uk/sites/default/files/finalpapers/Day2/R4/2_paper6.pdf) from Johnson has some quite interesting thoughts on the tradeof between consistency and performance.

devreal:
Thanks for your summary. Some remarks from my side:

> 1. puts are non-blocking, reads are blocking per default.

That totally makes sense, a non-blocking read is a strange concept (unless wrapped in a future). I'm open to discussing changing the default behavior of `GlobRef` to non-blocking `puts` as we need some form of synchronization to synchronize updates across units anyway (p2p or collective). It would be much easier to get good performance by default.

I'm not yet sure about the other implications though, esp. with regards to the next point below (how likely is it to re-read a value you just wrote into a global array? You'd want to use a local copy for performance reasons anyway...). 

> 2. after a put (write), the memory location is in an undefined state (for all units except the unit that performed the update), 

Unfortunately, currently it is in an undefined state for *all* units, including the issuing unit (due to missing local caches). Is that solved in CoArray in any way?

> 3. hence it is not allowed to read that location until a flush is executed (in fact is is allowed, but might return the old or the new value). Since we support fine granular flushes, the scope of the flush should determine for which units the address is in an defined state again. This is important for p2p synchronization.

Good point, will add that to the `dash::Array::flush` interface. 

> 4. GlobRef should provide a method for blocking put with local and remote completion.

That is the default right now but should be made explicit if the default changes (see 1.)

> 5. GlobRef should provide a method for non-blocking reads.
> 6. pure non blocking communication is supported by wrapping a GlobRef in a future.

Those transfers have to be explicit then, e.g., `array[0].get_async()` returns a future. 

With our current `GlobRef`/`GlobAsyncRef` implementation we are close to that (with different defaults with regards to sync/async operations). What's missing is a non-blocking `get`, for which we'd need to beef up the `dash::Future` implementation I guess.

fmoessbauer:
I already proposed the idea of non-blocking puts and blocking gets as a default to the Munich team.

> Unfortunately, currently it is in an undefined state for all units, including the issuing unit (due to missing local caches). Is that solved in CoArray in any way?

As our Coarray implementation is based on `container.async` / `GlobRefAsync` this is currently not clear. With the local cache in an async object that was the case, however unfortunately only for the same object, not the same memory location. That also fooled me. In Crays solution (I never tested it) the describe this behavior.

> Those transfers have to be explicit then, e.g., array[0].get_async() returns a future.

That is one possible solution. Johnson also proposes (and I vote for it) to build it utilizing the C++ operator semantic ([Coarray C++, P.10](http://www.pgas2013.org.uk/sites/default/files/finalpapers/Day2/R4/2_paper6.pdf)):
```cpp
auto a = array[0];  // returns a GlobRef, until now no dart_get
dash::future df(a); // wrap it in a future
dash::future df2;
df2 = array[0];       // same as construction above
```
To avoid local caches, we could (as already mentioned) use puts with local completion. Then the caching is offloaded to MPI. Alternatively we could do the caching in DART on a per-globmemory scope. Then this would be transparent to DASH. I do not recommend to implement the caching in DASH, as GlobRefs can be constructed upon DART global pointers, hence the semantics would be inconsistent again, if a user constructs a `GlobRef` using a `dart_gptr`.

devreal:
I have added an overload for `dash::Future<GlobRef<T>>` and removed the blocking `get` and cast operators from `GlobAsyncRef`, making `GlobAsyncRef` write-only. This is meant to avoid possible confusion about consistency/buffering in `container.async`, hopefully making it clear that an async `put` has to be flushed before it is visible in the container. Asynchronous transfers `get`s can be done using the Future.

dash/include/dash/GlobAsyncRef.h
```diff
@@ -320,33 +321,56 @@ std::ostream & operator<<(
...
+      &_handle);
   }
 
   /**
    * Create a Future from a \ref GlobAsyncRef instance.
    */
-  Future(dash::GlobAsyncRef<T>& aref) {
-    dart_storage_t ds = dart_storage<T>(1);
-    dart_get_handle(&_value, aref.dart_gptr(), ds.nelem, ds.dtype, &_handle);
+  Future(dash::GlobAsyncRef<T>& aref, size_t count = 1)
```
fmoessbauer: Adding support for bulk transfer should be a very deliberate choice, as it breaks with many of the DASH concepts. Is there a use-case where this brings any benefit over `dash::copy` / `copy_async`?

devreal: My reasoning was that I have to dynamically allocate the value buffer anyway (to make the Future movable) so I thought it would be a neat addition. But I agree that it does not really fit into the DASH concept.

PR #374, opened 2017-04-28
Fixes and Improvements for multi-node test runs

devreal:
I tried to run our test suite on multiple nodes on our cluster and came across some bugs that don't seem to manifest as errors if running on one node only. In particular, this PR contains fixes to the handling of `MPI_Request` objects and an issue in our shared memory optimization. 

Additionally, this PR adds `dash::generate_with_index` to enable parallel generation of container values in cases where the generated value depends on the index of the element. This function is applied to two tests in which the serial generation was really slow on 8 nodes.


dart-impl/mpi/src/dart_communication.c
```diff
@@ -1276,40 +1278,33 @@ dart_ret_t dart_waitall_local(
...
-            if (MPI_Request_free(&mpi_req[r_n]) != MPI_SUCCESS) {
-              DART_LOG_TRACE("dart_waitall_local ! MPI_Request_free failed");
-              ret = DART_ERR_INVAL;
-              break;
-            }
-          } else {
-            DART_LOG_TRACE("dart_waitall_local: cannot free request %zu "
-                           "mpi_sta[%zu] = %d (%s)",
-                           r_n,
+        if (handle[i]->request) {
```
fmoessbauer, line 1285: I am surprised that this even worked in the past...

dart-impl/mpi/src/dart_globmem.c
```diff
@@ -283,7 +283,7 @@ dart_team_memalloc_aligned(
   MPI_Comm_rank(sharedmem_comm, &sharedmem_unitid);
   // re-use previously allocated memory
   if (segment->baseptr == NULL) {
-    segment->baseptr = malloc(sizeof(char *) * team_data->sharedmem_nodesize);
+    segment->baseptr = calloc(team_data->sharedmem_nodesize, sizeof(char *));
```
fmoessbauer, line 286: `calloc` should not be necessary here, but who cares ;)

dash/include/dash/algorithm/Generate.h
```diff
@@ -48,6 +48,55 @@ void generate (
...
+ * \complexity  O(d) + O(nl), with \c d dimensions in the global iterators'
+ *              pattern and \c nl local elements within the global range
+ *
+ * \ingroup     DashAlgorithms
+ */
+template <
+    typename ElementType,
+    class    PatternType,
+    class    UnaryFunction >
+void generate_with_index(
```
fmoessbauer, line 73: Cool, I missed this function quite often.

dash/include/dash/algorithm/Generate.h
```diff
@@ -48,6 +48,55 @@ void generate (
...
+  /// Iterator to the initial position in the sequence
+  GlobIter<ElementType, PatternType> first,
+  /// Iterator to the final position in the sequence
+  GlobIter<ElementType, PatternType> last,
+  /// Generator function
+  UnaryFunction                      gen) {
+  /// Global iterators to local index range:
+  auto index_range  = dash::local_index_range(first, last);
```
fmoessbauer, line 81: Better use the new range based views, as this does not work with multiple local ranges AFAIK. Maybe @fuchsto can assist.

dash/include/dash/algorithm/Generate.h
```diff
@@ -48,6 +48,55 @@ void generate (
...
+  auto lbegin_index = index_range.begin;
+  auto lend_index   = index_range.end;
+
+  if (lbegin_index != lend_index) {
+    // Pattern from global begin iterator:
+    auto & pattern    = first.pattern();
+    auto first_offset = first.pos();
+    // Iterate local index range:
+    for (auto lindex = lbegin_index;
```
fmoessbauer, line 90: Same problem here.

fmoessbauer (commented):
Thanks for the fixes. See my comments inline.

dash/include/dash/algorithm/Generate.h
devreal: Do you have any pointer on how to use them? I'm not sure range based views and I have crossed paths so far...

fuchsto: Hey, I even introduced you personally last time in Munich!
Anyways: yes, I will add documentation.
For usage you can refer to quite an abundance of tests in dash/test/view.

devreal: To be honest, I don't even know where to start looking. I guess I could crawl through a host of undocumented code, still not knowing where to start or how to put the pieces together...

fuchsto (approved):

PR #377, opened 2017-05-01
fixes CI timeout on Circle

fmoessbauer:
This PR fixes the problem of CI timeouts by printing a message every two minutes on the CI host. In the past, often the CI timed out, as the clang build takes quite long without output. It is important to print the messages directly on the host, as the logs from the docker containers are buffered and so often not printed on the host in time.


devreal (approved):
Badly needed, thanks a lot!

PR #383, opened 2017-05-05
Array: Use correct template type for the index of the operator[]s.

bertwesarg:


devreal (approved):

devreal (changes requested):
On second thought, shouldn't it be `index_type` instead of `size_type`?

bertwesarg:
@dhinf told me, that `size_type` is the guaranteed unsigned type of `index_type`, thus we should use the `size_type`, as negative indices should be prohibited.

fmoessbauer (approved):

devreal (approved):

devreal:
@bertwesarg Can you please rebase/merge against development, CI should be happy now.

PR #392, opened 2017-05-09
build: Preload the GTest related cmake cache variables with their env…

bertwesarg:
…ironment content.


fmoessbauer (approved):
Thanks for the fix!

Issue #387, opened 2017-05-09
build: Have an MPI3 cmake check


bertwesarg:
It is currently possible to pass `cmake` but `make` fails because the used MPI implementation does not have MPI 3 support. It would be valuable for possible users, that than already `cmake` fails.

fmoessbauer:
Unfortunately the MPI implementation version (e.g. OpenMPI 2.1.0) does not match the MPI standard version (e.g. 3.0). Hence, we would have to implement the mapping on our own. We could do that propably for mpich, openmpi and intelmpi, but we will never be able to verify all MPI implementations.

But, you are right: We should at least document the requirements in the README. Can you do that?

bertwesarg:
See, I'm comming from the autotools, and the philosophy there is _do not test for version, test for features_. Thus you do not need a version mapping at all, just check if `MPI_OP_NO`, `MPI_Win_flush` and some other relevant functions/decls exists.

fmoessbauer:
Check for `MPI_NO_OP` is implemented and tested.

PR #393, opened 2017-05-09
README: Note that DART needs a MPI 3.0 compliant implementation.

bertwesarg:
See #387. I do not consider that issue fixed with this commit though.


fmoessbauer (approved):
I will have a look at the corresponding CMake checks.

fmoessbauer:
The desired functionality is implemented in PR #396

Issue #389, opened 2017-05-09
Add support for move in dash::Array


fmoessbauer:
Currently the `dash::Array` is [not movable](https://github.com/dash-project/dash/blob/development/dash/include/dash/Array.h#L887). To be consistent with the other containers, this should be implemented.

PR #394, opened 2017-05-09
Feat 389 array move

fmoessbauer:
This PR implements move semantic for `dash::GlobStaticMem` and `dash::Array`, as proposed in #389. The changes are implemented in a way that benefits from RAII. Hence, the move constructors for the containers can be defaulted in the future.

### DIFF
--- a/dash/include/dash/memory/GlobStaticMem.h
+++ b/dash/include/dash/memory/GlobStaticMem.h
@@ -238,7 +238,10 @@ public:
   ~GlobStaticMem()
   {
     DASH_LOG_TRACE_VAR("GlobStaticMem.~GlobStaticMem()", _begptr);
-    _allocator.deallocate(_begptr);
+    // check if has been moved away
+    if(!DART_GPTR_ISNULL(_begptr)){
+      _allocator.deallocate(_begptr);
+    }
     DASH_LOG_TRACE("GlobStaticMem.~GlobStaticMem >");
   }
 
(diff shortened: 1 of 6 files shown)


dash/include/dash/memory/GlobStaticMem.h
```diff
@@ -277,13 +277,14 @@ class GlobStaticMem
...
    */
   self_t & operator=(const self_t & rhs)
     = delete;
 
   /**
    * Move-assignment operator.
+   *
+   * \TODO make move constructor defaultable by using RAII 
    */
   self_t & operator=(self_t && other) {
```
devreal: Maybe I am missing something here: what is in the way of defaulting them? 

Also: the operator has no return statement.

devreal (changes requested):

dash/include/dash/memory/GlobStaticMem.h
fmoessbauer: Arg... these annoying missing return statements. Will fix it. With defaulting I mean writing  `self_t & operator=(self_t && other) = default`. This is only possible if all members support moving them. Pointers which will be deallocated on destruction have to be encapsulated with `std::unique_ptr`, as it is set to `nullptr` after moving it away.

devreal: Agh right, forgot about that.

devreal (approved):

Issue #388, opened 2017-05-09
build: Check that the provided HDF5 has MPI support


bertwesarg:
If the build env does provide an HDF5 lib, than this library does not necessarily have MPI support. But DASH does not perform a check in `cmake`, thus the build may fail at `make` time, which is cumbersome for a user.

fmoessbauer:
That is not correct. If the version of CMake is recent enough, it is checked if the found version is the parallel one. However, for older versions it's not possible to check which version is found. If you have a (portable!) solution for that, please tell us.

Here is the check implemented: [CMakeExt/HDF5.cmake](https://github.com/dash-project/dash/blob/development/CMakeExt/HDF5.cmake#L29)

bertwesarg:
Than increase the required CMake version! I find this rather lazy to rely on CMake to do such basic check. For HDF5, one would check if it `#define H5_HAVE_PARALLEL`, which should be trivial without relying on a specific CMake version.

PR #395, opened 2017-05-09
let cmake check if hdf5 parallel version is installed

fmoessbauer:
Addresses #388


bertwesarg:
This is odd. It detects, that HDF5 does not have parallel support but still enables it:
```
-- Looking for H5_HAVE_PARALLEL
-- Looking for H5_HAVE_PARALLEL - not found
-- HDF5 cmake module provides only serial version
-- HDF5 package not found, try to find libs
-- Found HDF5: /usr/lib64/libhdf5.so
-- HDF5 includes:   /usr/include
-- HDF5 libraries:  /usr/lib64/libhdf5.so
:
-- HDF5                      enabled
```

The build includes than still HDF5:
```
[ 29%] Building C object dart-impl/mpi/CMakeFiles/dart-mpi.dir/src/dart_io_hdf5.c.o
/home/wesarg/scratch/mephisto/dash/dart-impl/mpi/src/dart_io_hdf5.c: In function ‘dart__io__hdf5__prep_mpio’:
/home/wesarg/scratch/mephisto/dash/dart-impl/mpi/src/dart_io_hdf5.c:29:19: warning: implicit declaration of function ‘H5Pset_fapl_mpio’ [-Wimplicit-function-declaration]
   herr_t status = H5Pset_fapl_mpio(plist_id, comm, MPI_INFO_NULL);
                   ^~~~~~~~~~~~~~~~
```

bertwesarg:
Maybe unrelated, but even before this change, I get strange warnings from CMake, when building against an MPI-enabled HDF5:

```
-- Trying to find HDF5 using cmake package
-- Found HDF5: /home/wesarg/opt/hdf5-1.8.17-openmpi/lib/libhdf5.so (found version "1.8.17")  
-- Found HDF5: /home/wesarg/opt/hdf5-1.8.17-openmpi/lib/libhdf5.so  
-- HDF5 includes:   /home/wesarg/opt/hdf5-1.8.17-openmpi/include
-- HDF5 libraries:  /home/wesarg/opt/hdf5-1.8.17-openmpi/lib/libhdf5.so
:
-- HDF5                      enabled
:
WARNING: Target "ex.02.matrix_global_access.mpi" requests linking to directory "/home/wesarg/opt/hdf5-1.8.17-openmpi/lib".  Targets may link only to libraries.  CMake is dropping the item.
WARNING: Target "bench.03.gups.mpi" requests linking to directory "/home/wesarg/opt/hdf5-1.8.17-openmpi/lib".  Targets may link only to libraries.  CMake is dropping the item.
WARNING: Target "ex.03.globref.mpi" requests linking to directory "/home/wesarg/opt/hdf5-1.8.17-openmpi/lib".  Targets may link only to libraries.  CMake is dropping the item.
```

This is with CMake 3.7.2.

bertwesarg (approved):
Tested.

CMakeExt/HDF5.cmake
```diff
@@ -61,10 +53,23 @@ find_package_handle_standard_args(
...
 )
 
+
 # set flags
 if(HDF5_FOUND)
-#  set (HDF5_LINKER_FLAGS "-lhdf5_hl -lhdf5 -ldl -lm -lz")
-   set (HDF5_LINKER_FLAGS $ENV{HDF5_LIB} ${SZIP_LIB} -lz)
+  CHECK_SYMBOL_EXISTS(
+    H5_HAVE_PARALLEL
+    "${HDF5_INCLUDE_DIRS}/H5pubconf.h"
```
bertwesarg: I think we can solve this absolute path think with something like:

```
set(CMAKE_REQUIRED_INCLUDES ${HDF5_INCLUDE_DIRS})
```

At least @devreal is doing this in #396.

bertwesarg (changes requested):

CMakeExt/HDF5.cmake
fmoessbauer: I'll try that in the evening. The errors are strange, but it works nevertheless. Seems like a problem with static linking as the directory is chosen instead of the library (xx.a).

fmoessbauer:
@bertwesarg Now the CMakePushCheckState script is used. On our dev machine this now detects the two versions properly (tested with both ohpc packages).

CMakeExt/HDF5.cmake
```diff
@@ -59,18 +54,33 @@ find_package_handle_standard_args(
...
+  if(NOT HAVE_H5_PARALLEL)
+    message(STATUS "HDF5 provides only serial version")
+    set(HDF5_FOUND OFF CACHE BOOL "HDF5_FOUND" FORCE)
+    unset(HDF5_LIBRARIES)
+    unset(HDF5_INCLUDE_DIRS)
+  else()
+#   set (HDF5_LINKER_FLAGS "-lhdf5_hl -lhdf5 -ldl -lm -lz")
+    set (HDF5_LINKER_FLAGS $ENV{HDF5_LIB} ${SZIP_LIB} -lz)
+  endif()
+  cmake_pop_check_state()
```
bertwesarg: I find it clearer, that the pop is directly after the `CHECK_SYMBOL_EXISTS`. Also the style is not consistent, either ALL_APPER_CASE or all_lower_case.

bertwesarg (approved):
thas

PR #397, opened 2017-05-10
[WIP] Fix CI test hangs

devreal:


devreal:
Update: It's a mess. Attaching to running containers is broken with the Docker version used by CircleCI (1.10), see https://discuss.circleci.com/t/docker-1-10-0-is-available-beta/2100/44

For the records, the error when trying to attach to a running container (in order to attach gdb): 

```
ubuntu@box323:~$ sudo docker exec -it romantic_blackwell /bin/bash 
open /proc/self/oom_score_adj: no such file or directory
```

Starting a new container works of course but is useless for debugging without proper X forwarding (for xterm + gdb). Time to get CI on systems we control...

PR #398, opened 2017-05-10
Honor output range distribution in dash::transform

### LINKED ISSUE #386: Strange behavior of dash::transform with local input range
Right now, `dash::transform` can be called with a local input range. The user is responsible for blocking the output range, i.e., he has to call `dash::transform` for every target unit individually. For an example, see https://github.com/dash-project/dash/blob/development/dash/test/algorithm/TransformTest.cc#L81. 

This is really not intuitive. I propose changing this to require `len(local_input_range) == len(global_output_range)` and do the blocking to units in `dash::transform` using non-blocking accumulate calls. 

The use case (global element-wise reduction into a global container) is as follows (contributed by @bertwesarg): 

```
std::vector<double> local_v(100);    
std::fill(local_v.begin(), local_v.end(), (double)dash::myid());    
dash::Array<double> global_v(local_v.size());    
std::fill(global_v.lbegin(), global_v.lend(), 0.0);    
dash::transform<double>(        
  local_v.begin(), local_v.end(),        
  global_v.begin(),        
  global_v.begin(),        
  dash::plus<double>()    
);
```

devreal:
1. Implement global-to-global `dash::copy`, needed for
2. `dash::transform` that honors the distribution of the output range for local input ranges, which
3. now also works with std iterators, .e.g, coming from `std::vector`

Fixes #386

### DIFF
--- a/dash/include/dash/algorithm/Transform.h
+++ b/dash/include/dash/algorithm/Transform.h
@@ -4,6 +4,7 @@
 #include <dash/GlobRef.h>
 #include <dash/GlobAsyncRef.h>
 
+#include <dash/algorithm/Copy.h>
 #include <dash/algorithm/LocalRange.h>
 #include <dash/algorithm/Operation.h>
 #include <dash/algorithm/Accumulate.h>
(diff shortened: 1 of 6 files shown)


dash/include/dash/algorithm/Transform.h
```diff
@@ -271,40 +296,65 @@ GlobOutputIt transform(
...
   // Send accumulate message:
-  trace.enter_state("transform_blocking");
-  dash::internal::transform_blocking_impl(
+  auto &team    = pattern.team();
+  size_t towrite = num_local_elements;
+  auto out_it    = out_first;
+  auto in_it     = in_first;
+  while (towrite > 0) {
+    auto   lpos  = out_it.lpos();
+    size_t lsize = pattern.local_size(lpos.unit);
```
bertwesarg, line 333: Doesn't the `local_size` covers the total number of elements in the that unit? What if the unit has multiple local blocks?

dash/test/algorithm/TransformTest.cc
```diff
@@ -221,3 +223,41 @@ TEST_F(TransformTest, MatrixGlobalPlusGlobalBlocking)
...
+  global_v.barrier();
+
+  ASSERT_EQ_U(it, global_v.end());
+
+//  size_t idx = 0;
+
+  dash::for_each_with_index(global_v.begin() + 1, global_v.end(),
+    [](value_t val, size_t idx){
+      ASSERT_EQ_U(val, (dash::size() - 1) * 1000 + (idx - 1));
+      ++idx;
```
bertwesarg, line 259: Remove `++idx`.

dash/include/dash/algorithm/Transform.h
```diff
@@ -271,40 +296,65 @@ GlobOutputIt transform(
...
 
-  // Resolve local range from global range:
-  // Number of elements in local range:
-  size_t num_local_elements     = std::distance(in_first, in_last);
   // Global iterator to dart_gptr_t:
   dart_gptr_t dest_gptr         = out_first.dart_gptr();
   // Send accumulate message:
-  trace.enter_state("transform_blocking");
-  dash::internal::transform_blocking_impl(
+  auto &team    = pattern.team();
```
bertwesarg, line 327: Unused?

bertwesarg (commented):
Not yet tested.

dash/include/dash/algorithm/Transform.h
fmoessbauer: That is what I already mentioned multiple times. This approach only works for a single  continuous index range. Hence it would be better to use the new range based views interface by @fuchsto . There should be something like `dash::chunks(dash::local(...))`. Unfortunately I do not have detailed knowledge in this API either.

fuchsto: @fmoessbauer @bertwesarg
Yes, almost, it's
 `dash::blocks(dash::local(view_or_range_expr))`
or
`dash::local(dash::blocks(view_or_range_expr))`.
Containers are valid range expressions, iterators can be converted to ranges using
`dash::make_range(it_begin, it_end)`.

The term `chunks` was intended for any ad-hoc user-specified decomposition like `dash::chunks(dash::stride(10, array))`, blocks depend on the pattern. But actually there is no reason to differentiate between those.

fuchsto:
@devreal I merged your changes into PR #410 to fix this on a "substantial" level using views / ranges,  your unit test are especially useful for this.

devreal:
Perfect, thanks a lot!

PR #400, opened 2017-05-11
updated openmpi to 2.1.1. Temporary use only mpich and openmpi2 in CI…

fmoessbauer:
… (instead of openmpi 1)

After merging this PR, the build settings on hub.docker.com have to be set to branch development again. Propably a re-run of the CI is necessary.


devreal (approved):
Thanks for taking of this.

PR #402, opened 2017-05-11
DART: construct correct sharedmem mapping table

devreal:
Use current team's communicator instead of DART_COMM_WORLD.

Thanks to @HuanZhou2 for the discovery.


fmoessbauer (approved):
That's something that could be mentioned in the changelog as well.

devreal:
I think it's a bug I introduced while restructuring the code :(

PR #403, opened 2017-05-16
Check OpenMPI version

devreal:
... and only disable shared memory windows for versions <1.10.7 or <2.1.1. 

OpenMPI 2.1.1 came with a fix for the alignment of MPI shared memory windows, see https://www.mail-archive.com/announce@lists.open-mpi.org//msg00099.html. OpenMPI 1.10.7 likely will come with the same fix.


dart-impl/mpi/CMakeLists.txt
```diff
@@ -6,8 +6,6 @@ set(DASH_DART_IMPL_MPI_LIBRARY dart-mpi)
 
 set(DASH_DART_BASE_LIBRARY dart-base)
 
-include(${CMAKE_SOURCE_DIR}/CMakeExt/MPI.cmake)
-
```
fuchsto: Great, but why did you remove the include here?

fuchsto (changes requested):

dart-impl/mpi/CMakeLists.txt
devreal: Because it is already called in CMakeList in the top level directory: https://github.com/dash-project/dash/blob/development/CMakeLists.txt#L127

I think DASH also requires information on MPI, right?

fmoessbauer: DART should work as standalone module and hence cmake exports it as its own package. So I guess this include is mandatory here. DASH should not directly require information on MPI, but the MPI implementation can be queried using the dash config interface. This is used in some unit tests which do not work with some MPI backends.

bertwesarg:
Just to remove a common misnomer: It is called _Open MPI_!

fmoessbauer:
By the way: We should only require MPI if `DART-MPI` is used. Currently DASH cannot be build without MPI (shmem build): 

```
CMake Error at CMakeExt/MessageColor.cmake:25 (_message):
  Detected MPI library does not support MPI-3.
```

CMakeExt/MPI.cmake
```diff
@@ -30,9 +30,18 @@ if (MPI_INCLUDE_PATH AND MPI_LIBRARY)
   elseif ("${MPI_INCLUDE_PATH}" MATCHES "openmpi")
```
bertwesarg: You could improve the Open MPI detection, by replacing this string matching with a compile check which tests for the `OPEN_MPI` or the version macros.

devreal: Thought about that as well, yes. Sometimes I regret touching things in the first place though :D

Will look into your comments.

devreal:
Current build fails because the tests  environment uses MPI directly. Unless someone wants to fix that I will revert to calling `MPI.cmake` from the root `CMakeList.txt`.

fmoessbauer:
Yes, that is my fault. Will have a look at it. The problem is that we need MPI for gathering the test results when DASH is not running. There is already a check for MPI support, but DART sets the `MPI_IMPL_ID`. This should only be set if MPI is actually loaded in this compilation unit.

**Edit:** And the tests only require a minimal MPI feature set, so we should allow that for dash. @devreal : We should rename `MPI_IMPL_ID` to `DART_MPI_IMPL_ID` and add a new `MPI_AVAILABLE` which is only true if `mpi.h` is in the path and the lib is loaded.

CMakeExt/MPI.cmake
devreal: Implemented. Unfortunately, Cray MPI claims to be MPICH so we still need to check the include path.

fmoessbauer:
- Moved `MPI.cmake` to top level again as this sets up the MPI environement globally (`set(... CACHE ...)`).
- Separated MPI compatiblity for DASH and DART by replacing the fatal error with a state stored in variable `MPI_IS_DART_COMPATIBLE`. The compatiblilty is then only enforced in DART-MPI. So if using SHMEM (which is theoretically possible), and older MPI version can be used for features like HDF5 io or for gathering the test results.
- Whenever the macro `MPI_IMPL_ID` is set, MPI is available.

dart-impl/CMakeLists.txt
```diff
@@ -17,9 +17,11 @@ add_subdirectory(base)
 
 # DART implementation of every enabled DART variant:
 if (";${DART_IMPLEMENTATIONS_LIST};" MATCHES ";mpi;")
+  if(NOT MPI_IS_DART_COMPATIBLE)
+    message(FATAL_ERROR "Detected MPI library does not support MPI-3.")
+  endif()
```
devreal, line 34: Why not move this check to `dart-impl/mpi/CmakeLists.txt`?

devreal:
@fmoessbauer Thanks a lot for the changes!

dart-impl/CMakeLists.txt
fmoessbauer: would also be possible. Currently the build system of DASH and DART is quite tightly coupled, unfortunately.

fmoessbauer (approved):

bertwesarg:
`try_run` is bad when cross compiling.

devreal:
What's a better alternative then?

bertwesarg:
I know only one: multiple try_compile, until one succeeds. Getting the actual version is a little harder, but you already had this test once as an try_compile for the Open MPI case.

devreal:
Mhh, according to the documentation at https://cmake.org/cmake/help/v2.8.12/cmake.html#command:try_run `try_run` will compile the executable but won't execute it if `CMAKE_CROSSCOMPILE` is set. So this won't work. 

I will try to detect the MPI implementation through multiple `check_symbol_exists` and verify Open MPI version through a `try_compile`. An epic can of worms...

devreal:
@fuchsto: Bump. 

Upon request, Cray has added the pre-compiler macro `CRAY_MPICH_VERSION` in MPT 7.6.0. AFAICS, only IBM MPI flies under the radar. Maybe the Munich people want to poke IBM for that?

devreal:
Because this PR is over 9 months old and I need this feature in another place I will go ahead and merge unless someone voices opposition by EOB *today*.

PR #405, opened 2017-05-16
Build DART/DASH static libs with PIC

fuchsto:
When linking DART/DASH libraries in a shared library (as required for pydash bindings), static libraries must be build with `-fPIC` (Position Independent Code).

Using PIC per default increases library filesize, but it shouldn't hurt to enable it by default.

Also, with @fmoessbauer 's changes in `dash::Array`, the build broke when logging was enabled as no stream conversion for `std::unique_ptr` is defined. Fixed by removing a log message.


dash/include/dash/Array.h
```diff
@@ -1324,7 +1324,7 @@ class Array
     m_team->unregister_deallocator(
       this, std::bind(&Array::deallocate, this));
     // Actual destruction of the array instance:
-    DASH_LOG_TRACE_VAR("Array.deallocate()", m_globmem);
+    // DASH_LOG_TRACE_VAR("Array.deallocate()", m_globmem);
```
devreal: Please remove the commented out code.

dart-impl/base/CMakeLists.txt
```diff
@@ -120,6 +120,11 @@ set_target_properties(
   COMPILE_FLAGS ${ADDITIONAL_COMPILE_FLAGS}
 )
 
+set_target_properties(
+  ${DASH_DART_BASE_LIBRARY}
+  PROPERTIES POSITION_INDEPENDENT_CODE TRUE
+)
+
```
devreal, line 129: Fine with me, couldn't see a performance impact significant to stick out of the noise in my experiments. That might be different on 32bit systems, though (in case anyone still uses them for real).

devreal (changes requested):

dart-impl/base/CMakeLists.txt
fuchsto: Hmyes. I'd prefer a more canonical way myself to toggle this setting, like `BUILD_SHARED_LIBS`, but I'm not aware of any. Still, will wrap it in a build option `BUILD_PIC_LIBS` and disable it by default. It's sensible to assume that the common use case is to link DASH with an executable.

devreal: Agreed. Will pydash be integrated into the main build system? If so, we might just trigger PIC builds if pydash is being built.

devreal (approved):

PR #406, opened 2017-05-18
[WIP] Feat dart shmem

fmoessbauer:
First minimal (at least compiling) version of DART-shmem. The changes are mostly related to the new unit id concept and typed storage.

Don't be disturbed by the ugly code layout: It was not pretty before, and most of my changes are done using `sed`.

What currently blocks linking is the locality interface by @fuchsto . Is it possible to disable this in cmake?

### DIFF
--- a/dart-impl/shmem/src/dart_groups_impl.c
+++ b/dart-impl/shmem/src/dart_groups_impl.c
@@ -4,6 +4,12 @@
 #include <dash/dart/if/dart_team_group.h>
 #include <dash/dart/shmem/dart_groups_impl.h>
 
+static struct dart_group_struct* allocate_group()
+{
+  struct dart_group_struct* group = malloc(sizeof(struct dart_group_struct));
+  return group;
+};
+
 dart_ret_t dart_group_sizeof(size_t *size)
 {
   *size=sizeof(dart_group_t);
@@ -32,13 +38,15 @@ dart_ret_t dart_group_destroy(dart_group_t *group)
 dart_ret_t dart_group_clone(const dart_group_t g, dart_group_t *gout)
 {
   int i;
+  struct dart_group_struct* res = allocate_group();
   
-  (*gout)->nmem = g->nmem;
+  res->nmem = g->nmem;
   for (i = 0; i < MAXSIZE_GROUP; i++)
     {
-      ((*gout)->g2l)[i] = (g->g2l)[i];
-      ((*gout)->l2g)[i] = (g->l2g)[i];
+      (res->g2l)[i] = (g->g2l)[i];
+      (res->l2g)[i] = (g->l2g)[i];
     }
+  *gout = res;
   return DART_OK;
 }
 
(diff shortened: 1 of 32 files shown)


fmoessbauer:
Currently, it builds (if commenting out all locality stuff ...), but the command line arguments do not get to ` dart_init`. This should be checked, as this might also be a bug in the dart-mpi build.

`/dash-test-shmem --dart-id=0 --dart-size=1` (better than using dartrun, as there it's hard to attach a debugger).

fuchsto:
@fmoessbauer The locality stuff is about to be separated out entirely into dyloc, so removing locality functions from DART is perfectly fine.

dart-impl/shmem/src/dart_groups_impl.c
```diff
@@ -14,36 +14,36 @@ dart_ret_t dart_group_create(dart_group_t *group)
...
   return DART_OK;
 }
 
-dart_ret_t dart_group_clone(const dart_group_t *g, dart_group_t *gout)
+dart_ret_t dart_group_clone(const dart_group_t g, dart_group_t *gout)
 {
   int i;
   
-  gout->nmem = g->nmem;
+  (*gout)->nmem = g->nmem;
```
devreal: In the MPI implementation, `gout` is actually allocated: 

```
 * \param[out] gout Pointer to the target group object (will be allocated).
```

dart-impl/shmem/src/dart_groups_impl.c
```diff
@@ -65,108 +65,115 @@ void group_rebuild(dart_group_t *g)
...
+dart_ret_t dart_group_union(const dart_group_t g1,
+                            const dart_group_t g2,
                             dart_group_t *gout)
 {
   int i;
   
   for (i = 0; i < MAXSIZE_GROUP; i++)
     {
-      (gout->g2l)[i] = -1;
+      ((*gout)->g2l)[i] = -1;
```
devreal: `gout` should be allocated, see above.

dart-impl/shmem/src/dart_malloc.c
```diff
@@ -7,11 +7,28 @@
...
+    case DART_TYPE_INT: sizeof(int); break;
+    case DART_TYPE_UINT: sizeof(unsigned int); break;
+    case DART_TYPE_LONG: sizeof(long); break;
+    case DART_TYPE_ULONG: sizeof(unsigned long); break;
+    case DART_TYPE_LONGLONG: sizeof(long long); break;
+    case DART_TYPE_FLOAT: sizeof(float); break;
+    case DART_TYPE_DOUBLE: sizeof(double); break;
+    default: -1;
+  }
+}
```
devreal: Does this compile? I don't see any return statement

dart-impl/shmem/src/dart_mempool.c
```diff
@@ -21,27 +21,28 @@ void dart_mempool_init(dart_mempoolptr pool)
...
-			       dart_unit_t myid,
+			       dart_team_unit_t myid,
 			       size_t localsz)
 {
   if( !pool ) return DART_ERR_INVAL;
 
   size_t totalsz = teamsize * localsz;
   int   attach_key;
   void *attach_addr;
+  dart_team_unit_t root = {0};
```
devreal: Better: `DART_TEAM_UNIT_ID(id)`

devreal (dismissed):
Thanks for taking care of this. Some suggestions inline

devreal:
While we are at it: I assume the shmem implementation uses relative offsets in global pointers as well, right? Is there any use for `gptr.addr_or_offset.addr`? Aside from being quite verbose, using the `addr` (in the sense of an absolute memory address) field breaks the common assumption that the first allocated element starts at `offset_or_addr.offset == 0`. There is no way to tell whether a `gptr` uses absolute or relative addresses.

fmoessbauer:
Currently the shm only works for one unit, and there it fails in ` dart_put_blocking`  in the `memcopy`.

PR #407, opened 2017-05-18
fixed argc and argv not being set in tests

fmoessbauer:
The problems in PR #406 related to missing argc and argv are due to a erroneous declarations of the global variable `TESTENV`. Hence, each translation unit got it own global state.


devreal (approved):
Good catch!

PR #419, opened 2017-07-10
Improve DART handle handling

devreal:
I looked at `dart_test_local` today and, well, had to do something about it...

This PR: 
- Makes sure handles are properly `free`'d if test succeeds
- Reduces the amount of debug output ("hey, I'm gonna call `free` now, wait for it...". Srsly?)
- Uses `alloca` for small temporary allocation to avoid calling `malloc` in `dart_waitall*` and `dart_testall_local`
- Makes sure a flush is only issued if required, allowing flush-free `dart_wait` on handles acquired from `dart_get_handle`
- Gets rid of `MPI_STATUS` objects, which are not needed


fmoessbauer (approved):

devreal:
@dhinf I changed the interface accordingly, thanks for pointing out this flaw. The corresponding functions had not been used so far.

dhinf:
Thanks. I uses dart_wait in the branch feat-halo. That's why i noticed it.

dhinf (approved):

PR #420, opened 2017-07-10
[WIP] Add full test case for variadic delayed Matrix allocation

devreal:
@fuchsto: Please have a look at this. Your test `MatrixTest.DelayedAlloc` succeeds while the new test `MatrixTest.VariadicDelayedAlloc` fails while checking the validity of the pattern arguments, which are the same as in the first test. So either the first test case is invalid or the check in `PatternArguments` is overly strict. I cannot decide that :)


devreal:
Hmpf, the error is not triggered in CI. Here is  what I see on my laptop:

```
[= -1 LOG =]               TestBase.h : 251 | ===> Running test case MatrixTest.VariadicDelayedAlloc ...
[=  1 LOG =]               TestBase.h : 254 | -==- DASH initialized with 4 units
[    1 ERROR ] [  9679 ] PatternArguments.h       :280  | dash::exception::InvalidArgument             | [ Unit 1 ] Pattern arguments invalid: Mixed distribution types
[  ERROR   ] [UNIT 1] in [=  1 LOG =]               TestBase.h : 264 | -==- Test case finished at unit 1
[=  1 LOG =]               TestBase.h : 267 | -==- Finalize DASH at unit 1
[= -1 LOG =]               TestBase.h : 272 | <=== Finished test case MatrixTest.VariadicDelayedAlloc with 4 units

```

ddiefenthaler:
I can report the same error, for me it even triggers in the first test `MatrixTest.DelayedAlloc`.
It occurs in a 32bit VM, only when run with 4 or 6 units.

I tried to debug it a little bit and it seems to fail in `check_tile_constraints()` of [PatternArguments.h#L265](https://github.com/dash-project/dash/blob/development/dash/include/dash/pattern/internal/PatternArguments.h#L265).
The distribution passed down seems to be [NONE, TILE(5), TILE(3)].

devreal:
The main issue has been fixed in #444 (thanks to @fmoessbauer). What is left in this PR is splitting the test for variadic matrix allocation into a separate test case.

PR #422, opened 2017-07-14
Allow DART to use regular MPI Windows

devreal:
Because dynamic windows have grown to be a constant pain (both performance- and debugging-wise), this PR introduces the CMake flag `ENABLE_DYNAMIC_WINDOWS`, which controls whether dynamic windows are used for collective memory allocation or not. If disabled, DART uses `MPI_Win_allocate` instead of `MPI_Win_attach`. Dynamic windows will still be in use: 
1) for registered memory, and  
2) if shared memory windows are enabled

This also comes with a host of changes in the DART communication functions (in part simplifying them).


devreal:
For the records: one CI run failed with an error from NastyMPI, which seemed to be a random hickup. However, I could not reproduce it locally and after triggering a rebuild things look fine again.

rkowalewski (approved):
As already described in an E-Mail it is not the best solution to wrap this in a compile-time flag. However, we need certainly an alternation between these two approaches as our benchmarks have shown.

PR #423, opened 2017-07-17
Fixes to [Local]MatrixRef

devreal:
- Use `size_type` in `operator[]` interface (still converted to `index_type` in the back though)
- `LocalMatrixRef`: directly return `T&` on lowest level in `operator[]` (similar to `MatrixRef`)
- Some minor improvements.

### DIFF
--- a/dash/include/dash/matrix/internal/LocalMatrixRef-inl.h
+++ b/dash/include/dash/matrix/internal/LocalMatrixRef-inl.h
@@ -12,7 +12,7 @@ template <class T_>
 LocalMatrixRef<T, NumDim, CUR, PatternT>
 ::LocalMatrixRef(
   const LocalMatrixRef<T_, NumDim, CUR+1, PatternT> & previous,
-  index_type coord)
+  size_type coord)
   : _refview(previous._refview)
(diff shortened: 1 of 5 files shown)


dash/include/dash/matrix/internal/LocalMatrixRef-inl.h
```diff
@@ -359,15 +375,65 @@ ::operator[](
...
 template<typename T, dim_t NumDim, dim_t CUR, class PatternT>
+template<dim_t __NumViewDim>
+typename std::enable_if<(__NumViewDim == 0), const T&>::type
+LocalMatrixRef<T, NumDim, CUR, PatternT>
+::operator[](
+  size_type pos) const
+{
+  auto  l_coords   = _refview._coord;
+  auto& l_viewspec = _refview._l_viewspec;
+  auto& pattern    = _refview._mat->_pattern;
```
fmoessbauer: Why not const? `l_viewspec` and `pattern` are not written.

fmoessbauer (approved):
Looks good! If possible, add the const as stated in my comment.

Issue #426, opened 2017-07-28
std::sort() for dash::Array fails


mmue:
Compiling a small test program for sorting a dash::Array with std::sort fails.

Test program:

```c++
#include <iostream>
#include <random>
#include <libdash.h>

using namespace std;

int main( int argc, char* argv[] ) {
  dash::init( &argc, &argv );

  dash::Array<int> arr(100);

  if ( dash::myid() == 0 ) {
    for ( auto it= arr.begin(); it != arr.end(); it++ ) {
      *it = std::rand() % 20 + 1;
    }
  }

  dash::barrier();

  std::sort(arr.begin(), arr.end());

  dash::barrier();

  if ( dash::myid() == 0 ) {
    for( auto el: arr ) {
      cout << (int)el << " ";
    }
    cout << endl;
  }

  dash::finalize();
}
```

Error message:
```bash
$ make
mpicxx -std=c++11 -o test -I/home/mmu/dash-tmp/dash-0.3.0/include test.cc -L/home/mmu/dash-tmp/dash-0.3.0/lib -ldash-mpi -ldart-mpi -ldart-base
In file included from /usr/include/c++/5/bits/char_traits.h:39:0,
                 from /usr/include/c++/5/ios:40,
                 from /usr/include/c++/5/ostream:38,
                 from /usr/include/c++/5/iostream:39,
                 from test.cc:1:
/usr/include/c++/5/bits/stl_algobase.h: In instantiation of ‘void std::iter_swap(_ForwardIterator1, _ForwardIterator2) [with _ForwardIterator1 = dash::GlobIter<int, dash::BlockPattern<1>, dash::GlobStaticMem<int, dash::allocator::SymmetricAllocator<int> >, dash::GlobPtr<int, dash::GlobStaticMem<int, dash::allocator::SymmetricAllocator<int> > >, dash::GlobRef<int> >; _ForwardIterator2 = dash::GlobIter<int, dash::BlockPattern<1>, dash::GlobStaticMem<int, dash::allocator::SymmetricAllocator<int> >, dash::GlobPtr<int, dash::GlobStaticMem<int, dash::allocator::SymmetricAllocator<int> > >, dash::GlobRef<int> >]’:
[...]

PR #427, opened 2017-07-30
Fix ADL of swap to support STL algorithms on dash iterators

fmoessbauer:
This PR fixes some ADL problems on `dash::GlobRef<T>` which prevented STL algorithms on dash iterators to compile. Also a new testsuite is added to explicitly test these use-cases. See issue #426, thanks to @mmue for reporting.

The following snippet (taken from `GlobIterTest.cpp`) demonstrates the new ADL functionality as well as the move-semantics of `dash::GlobRef`

```cpp
auto refbeg = *(arr.begin());
auto refend = *(arr.end()-1);
{
  using std::swap;
  using std::iter_swap;
  refbeg.swap(refend);
  swap(refbeg, refend);
  swap(*(arr.begin()), *(arr.end()-1));
  iter_swap(arr.begin(), arr.end()-1);
}
// without ADL
{
  // (1) uses move semantics, hence references are swapped
  // but not values of references
  std::swap(refbeg, refend);
  // (2) not possible as it would use move semantics on temporaries -> compiler error
  std::swap(*(arr.begin()), *(arr.end()-1));
  // (3) works as internally unqualified swap is used, values are swapped
  std::iter_swap(arr.begin(), arr.end()-1);
}
```


dash/test/iterator/GlobIterTest.h
```diff
@@ -0,0 +1,21 @@
+#ifndef DASH__TEST__GLOB_ITER_REF_TEST_H_
+#define DASH__TEST__GLOB_ITER_REF_TEST_H_
+
+#include "../TestBase.h"
+
+/**
+ * Test fixture for non-blocking operations using \c dash::GlobAsyncRef.
```
devreal: The comment is not quite correct ;)

dash/test/iterator/GlobIterTest.cc
```diff
@@ -0,0 +1,91 @@
...
+  dash::Array<int> arr(dash::size());
+  arr.local[0] = static_cast<int>(dash::myid());
+  arr.barrier();
+
+  if(dash::myid() == 0){
+    std::iter_swap(arr.begin(), arr.end()-1);
+  }
+  arr.barrier();
+  int newbegval = arr[0];
+  int newendval = arr[static_cast<int>(dash::size()-1)];
```
devreal: cast is not required

dash/test/iterator/GlobIterTest.cc
```diff
@@ -0,0 +1,91 @@
...
+
+  if(dash::myid() == 0){
+    auto refbeg = *(arr.begin());
+    auto refend = *(arr.end()-1);
+    refbeg.swap(refend);
+    //swap(*(arr.begin()), *(arr.end()-1));
+  }
+  arr.barrier();
+  int newbegval = arr[0];
+  int newendval = arr[static_cast<int>(dash::size()-1)];
```
devreal: cast is not required

dash/test/iterator/GlobIterTest.cc
```diff
@@ -0,0 +1,91 @@
+
+#include "GlobIterTest.h"
+
+#include <dash/Array.h>
+#include <algorithm>
+
+TEST_F(GlobIterTest, CompilerADLTest){
+  dash::Array<int> arr(dash::size());
+  arr.local[0] = static_cast<int>(dash::myid());
```
devreal: cast is not required

devreal (changes requested):
Looks good to me, thanks a lot for fixing this! But why call it `GlobIterTest`? Maybe `StdAlgoTest` is better suitable.

devreal (approved):

Issue #418, opened 2017-07-06
MatrixRef::lbegin() fails to compile


devreal:
Consider the following code snippet: 

```
dash::NArray<int, 2> matrix(dash::size(), N);
int *my_row = matrix[dash::myid()].lbegin();
```
which gives the following compiler error:
```
/home/joseph/opt/dash-0.3.0//include/dash/matrix/internal/MatrixRef-inl.h: In instantiation of ‘T* dash::MatrixRef<T, NumDimensions, CUR, PatternT>::lbegin() [with ElementT = int; int NumDimensions = 2; int NumViewDim = 1; PatternT = dash::BlockPattern<2, (dash::MemArrange)1u, long int>]’:
isx.cc:277:63:   required from here
/home/joseph/opt/dash-0.3.0//include/dash/matrix/internal/MatrixRef-inl.h:269:28: error: cannot convert ‘dash::LocalMatrixRef<int, 2, 2, dash::BlockPattern<2, (dash::MemArrange)1u, long int> >::iterator {aka dash::GlobViewIter<int, dash::BlockPattern<2, (dash::MemArrange)1u, long int>, dash::GlobStaticMem<int, dash::allocator::SymmetricAllocator<int> >, dash::GlobPtr<int, dash::GlobStaticMem<int, dash::allocator::SymmetricAllocator<int> > >, dash::GlobRef<int> >}’ to ‘int*’ in return
   return sub_local().begin();
```

As a workaround, I can probably use `matrix.lbegin()` for now.

ddiefenthaler:
I'm surprised this one is left unanswered so long.

I tried to look into this yesterday and wanted to report my findings.
In short this `TODO` has to be done: [MatrixRef.h#L47-L49](https://github.com/dash-project/dash/blob/development/dash/include/dash/matrix/MatrixRef.h#L47-L49)

To check this use case in a unit test it is possible to extend the existing case `MatrixTest.MatrixLBegin`:
```diff
   EXPECT_EQ_U(myid, static_cast<int>(*(matrix.lbegin())));
+  EXPECT_EQ_U(myid, static_cast<int>(*(matrix[myid].lbegin())));
   EXPECT_EQ_U(myid, static_cast<int>(*(matrix.local.block(0).begin())));
   EXPECT_EQ_U(myid, static_cast<int>(*(matrix.local.begin())));
```

First of all the type of `lbegin()` in `MatrixRef` can't be right.
[...]

PR #428, opened 2017-08-03
[WIP]Fix MatrixRef::lbegin/lend

devreal:
This PR fixes two issues with `MatrixRef` `lbegin`/`lend`: 

1. Fixed compiler error 
2. Fixed `lend()` in `LocalMatrixRef` (and thus `MatrixRef`) to not return a `nullptr` but a pointer past the end of the range (`GlobViewIter::local()` returns `nullptr` if pointing beyond the current unit)

### DIFF
--- a/build.nasty.sh
+++ b/build.nasty.sh
@@ -61,11 +61,11 @@ fi
 # Configure with default release build settings:
 mkdir -p $BUILD_DIR
 rm -Rf $BUILD_DIR/*
-(cd $BUILD_DIR && cmake -DCMAKE_BUILD_TYPE=Release \
+(cd $BUILD_DIR && cmake -DCMAKE_BUILD_TYPE=Debug \
                         -DENVIRONMENT_TYPE=default \
                         -DINSTALL_PREFIX=$HOME/opt/dash-0.3.0-nasty \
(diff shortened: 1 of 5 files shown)


devreal:
The offsets are still broken and I cannot seem to figure out which offsets to use. Will look at it again tomorrow... Any help is appreciated, though.

devreal:
@fuchsto I'm giving up here as it is not clear to me anymore which are global and local block coordinates and indexes. Also, I suspect that there are fixes lingering around in your branches. I'd appreciate a timely merge or a quick fix as the broken `lbegin()` on blocks is a road block for me right now.

fuchsto:
@devreal Sorry, I don't get it: what's the original issue? Is the use case covered in the unit test you added?
I found a bug related to #418 and will fix it in a separate PR, not sure if it helps your use case.

devreal:
> Sorry, I don't get it: what's the original issue?

This PR was my attempt to fix `Matrix::lbegin` a month after I reported the issue in #418 but I got lost in the usage of `CartesianIndexSpace` (does it hold local or global indexes? pretty confusing for me) My original use case should be covered by the test case I added here.

Issue #430, opened 2017-08-09
Feature Request: Provide DASH compiler wrapper


devreal:
Building and linking DASH applications can be challenging for novice users as they have to provide the correct paths and libraries (in correct order) to the compiler and linker. On top of that, the libraries to link may vary based on the platform and configuration. Having simple `dashCC`/`dashcxx`/`dashc++` wrapper (shell scripts) would make lives easier.

This is just to keep track of it, I haven't put any effort into it, yet. Maybe someone already has something along that way lingering around?

devreal:
Closed by #431

PR #431, opened 2017-08-10
Provide DASH compiler wrapper

devreal:
See #430. 

The wrapper shell script is created at CMake-time as `dash-<DART_IMPL>cxx` and symlinks `dash-<DART_IMPL>c++` and `dash-<DART_IMPL>CC` are installed with it, i.e., `dash-mpicxx`, `dash-mpic++`, and `dash-mpiCC`.

The script uses the compiler determined by CMake and takes care of setting the right include paths (based on `CMAKE_INSTALL_PREFIX`). It also passes on libraries for PAPI, libNUMA, MKL, hwloc, hdf5, PLASMA, and SCALAPACK, as determined by CMake. Unfortunately, I couldn't test all of them.

Apart from that, the script also sets the C++ standard flag (user-provided standard flags should override this). It does not pass any other compiler-flags, i.e., optimizations, warnings, LTO (the latter could be useful but the flags currently used by DASH won't fit).  

There are two special options at the moment: 
- `[-]-dash:verbose`: print the compiler command to be invoked
- `[-]-dash:nocppflags` suppress passing DASH-related macros to the compiler. 

Other options can be added easily.

Tested on our Linux cluster and the Cray XC40. If possible, please give it a try on your system and let me know if something breaks.


ddiefenthaler:
Works for me :+1:, Thanks!

For little projects this works fine, but what about larger ones (multiple source files)?
Would it be possible to add an option that creates a CMake-file for that?
The build-files by @fmoessbauer in [dash-apps/dummy-app](https://github.com/dash-project/dash-apps/tree/master/dummy-app) could be used as a default for that.

devreal:
Can't you use `dash-mpicxx` as the compiler in your CMake project?

ddiefenthaler:
If this works properly it is a very clean solution.
But cmake is still a riddle for me, so I'm not sure if it works that simple.
I remember @fmoessbauer saying something about passing the same flags to the app as they were used in the build process of `DASH`. In his cmake-file I see this line ([line 34](https://github.com/dash-project/dash-apps/blob/1cf64913f8bc2a41a0f65e786e0bcab3522e83bd/dummy-app/CMakeLists.txt#L34)): 
```set( CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} ${DASH_CXX_FLAGS} ${CXX_FLAGS}" )```

Are the same parameters used in the wrapper?
I see the `@CMAKE_CXX_FLAGS@` in the wrapper, but I do not know where they come from.
Nor do I know whether it is really important to use exactly the same flags...

**Comment by devreal** [comment]
> I remember @fmoessbauer saying something about passing the same flags to the app as they were used in the build process of DASH. In his cmake-file I see this line (line 34):
`set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${DASH_CXX_FLAGS} ${CXX_FLAGS}" )`

I don't think this is a good idea in general. IMO, the only flags we should pass on are macro definitions, include paths and libraries, and the language standard. The user should be free to select his own optimization and debug levels, among other things. Hence, the `${CMAKE_CXX_FLAGS}` are filtered int he wrapper script.

Maybe someone has a CMake application project at hand and wants to try out the wrapper?

**Review by ddiefenthaler — CHANGES_REQUESTED** [review]
I still like the idea to provide a compiler-wrapper that helps compile little projects without the need of a big build system.

But as I learned more about cmake the last days,
I think that most of the stuff the script is doing at runtime could be moved to buildtime or even could be done when cmake configures its Makefiles.

For example the parsing of the cppflags may also be useful when using plain cmake without the wrapper.
So perhaps it should be collected in seperate variables already during the CMakeLists of dash.

It should possible to assemble the ldflags during buildtime using `add_custom_command` or `add_custom_target`.

Also `CMAKE_EXPORT_COMPILE_COMMANDS` may be worth a look.

**Comment by devreal** [comment]
> I still like the idea to provide a compiler-wrapper that helps compile little projects without the need of a big build system.

That's only half the point. The point of this really is that we need a way to adapt the pass things like  libraries and compiler flags required by DASH to the application compile step and to adapt to changes in the DASH installation without having to adapt the application's build system.

> 
> But as I learned more about cmake the last days,
> I think that most of the stuff the script is doing at runtime could be moved to buildtime or even could be done when cmake configures its Makefiles.
> 
> For example the parsing of the cppflags may also be useful when using plain cmake without the wrapper.
> So perhaps it should be collected in seperate variables already during the CMakeLists of dash.

Good point, I will look into that. 

> 
> It should possible to assemble the ldflags during buildtime using add_custom_command or add_custom_target.
> 
> Also CMAKE_EXPORT_COMPILE_COMMANDS may be worth a look.

I'm not sure how `add_custom_target` or `add_custom_command` would help here. Any suggestions are welcome.

**Comment by devreal** [comment]
@ddiefenthaler please have a look at the latest changes, most of flag processing has now moved to the CMake file.

**Review by fmoessbauer — APPROVED** [review]
For me this PR looks good. Especially for new DASH developers this simplifies compiling a lot. We should make a note in the readme as well as in the readthedocs documentation.

**Review by ddiefenthaler — APPROVED** [review]
Thanks for the changes, in my opinion it is cleaner now.

**Comment by devreal** [comment]
> We should make a note in the readme as well as in the readthedocs documentation.

I added an entry to the the Changelog. Do you mean to also mention it in the doxygen documentation? I'm not sure where is the right place to put it there.

**Comment by fmoessbauer** [comment]
I thought of putting it in the `Running DASH Applications` section in `README.md`. But we can also merge this PR first and then update the Readme.

**Comment by devreal** [comment]
I'll write a short section on building DASH applications, that has been missing anyways

**Comment by ddiefenthaler** [comment]
>> It should possible to assemble the ldflags during buildtime using add_custom_command or add_custom_target.
>> 
>> Also CMAKE_EXPORT_COMPILE_COMMANDS may be worth a look.
>
>I'm not sure how `add_custom_target` or `add_custom_command` would help here. Any suggestions are welcome.

The `PARSE_CPPFLAGS()` should be possible to execute at compiletime this way.
For my understanding it is possible to call any abitrary command that way.
But it is probably not worth the effort as it would require multiple additional files.
Also it would be harder to read and it would be hard to extend it with options that way.

But on the regard of the `CXXFLAGS`. When I looked at the file produced by cmake I noticed some additional stuff:
- I noticed that there are no `-I` include dir flags. I wonder if this breaks with third party libraries, if they are not in the default include path? I'm asking as I'm guessing they are also needed to produce the syntastic config files from PR #445.
- Also there are flags like `-mtune=native -march=native` included, which I guess would be useful when building for a different environment. I'm not sure if this is really relevant for dash, but I could imagine that building for `supermic` might need those flags.
- Currently it filters for `-DDASH`, but I think I also saw some `-DDART`-flags
- For the `Debug` build mode it may be worth to provide an option to keep the warning flags. I propose an option like `[-]-dash:inheritwarnings` instead of having to retype all the warning flags.

**Comment by devreal** [comment]
> The PARSE_CPPFLAGS() should be possible to execute at compiletime this way.

Part of that has to be done at runtime based on the parameters passed to the script. We might want to add parameters in the future to enable additional flags, hence I pass everything down to the script.

> 
>     I noticed that there are no -I include dir flags. I wonder if this breaks with third party libraries, if they are not in the default include path? I'm asking as I'm guessing they are also needed to produce the syntastic config files from PR #445.

Good catch, should be fixed now.

>     Also there are flags like -mtune=native -march=native included, which I guess would be useful when building for a different environment. I'm not sure if this is really relevant for dash, but I could imagine that building for supermic might need those flags.

Optimization flags are not passed on on purpose because we should really leave it up to the user to decide the optimization flags he wants to use. 

>     Currently it filters for -DDASH, but I think I also saw some -DDART-flags

`-DDART` flags are not relevant for the DASH headers (if they are used somewhere it is a bug). Hence, we can drop them. 

>     For the Debug build mode it may be worth to provide an option to keep the warning flags. I propose an option like [-]-dash:inheritwarnings instead of having to retype all the warning flags.

See my argument above, the user should be responsible for choosing the warnings he wants to see. The simplest case `-Wall` or `-Wextra` is really easy to use, we are just a bit picky about the warnings we deliberately ignore :)

**Comment by ddiefenthaler** [comment]
**TL;DR**

From my understanding this PR should be ready to get merged. All the stuff I mentioned is of minor relevance.

>> The PARSE_CPPFLAGS() should be possible to execute at compiletime this way.
>
>Part of that has to be done at runtime based on the parameters passed to the script. We might want to >add parameters in the future to enable additional flags, hence I pass everything down to the script.

Yeah, that is why I meant it is not worth the effort. At first I just wanted to mention it, as it may be worth a thought.

>> I noticed that there are no -I include dir flags. I wonder if this breaks with third party libraries, if they are not in the default include path? I'm asking as I'm guessing they are also needed to produce the syntastic config files from PR #445.
>
>Good catch, should be fixed now.

Thanks for the change

>> Also there are flags like -mtune=native -march=native included, which I guess would be useful when building for a different environment. I'm not sure if this is really relevant for dash, but I could imagine that building for supermic might need those flags.
>
>Optimization flags are not passed on on purpose because we should really leave it up to the user to decide the optimization flags he wants to use. 

Ok, I was thinking about cross-compiling here as the `-march`-flag some kind of looks like it. Up to now I always ignored cross-compiling so I have no clue which flags would go with it. That is also why I asked if it is really relevant for dash, as I don't know if it is really necessary to cross-compile dash-applications.

>> Currently it filters for -DDASH, but I think I also saw some -DDART-flags
>
>`-DDART` flags are not relevant for the DASH headers (if they are used somewhere it is a bug). Hence, we can drop them. 

I was thinking about inlined dart functions here.

>> For the Debug build mode it may be worth to provide an option to keep the warning flags. I propose an option like [-]-dash:inheritwarnings instead of having to retype all the warning flags.
>
>See my argument above, the user should be responsible for choosing the warnings he wants to see. The simplest case `-Wall` or `-Wextra` is really easy to use, we are just a bit picky about the warnings we deliberately ignore :)

Ok, I'm aware of what you've writte before, I just didn't know the meaning of all the additional warning-flags. I assumed they were even on top of `-Wall`/`-Wextra`.
Knowing that now I suppose that it would be nice to hide these additional warnings from within DASH,
as otherwise users that compile with `-Wall`/`-Wextra` stumble across our well-known and ignored warnings. I assume there is no easy way to achieve this if at all possible...

**Comment by devreal** [comment]
> From my understanding this PR should be ready to get merged. All the stuff I mentioned is of minor relevance.

Thanks :)

>  I was thinking about inlined dart functions here.

Good point. I just checked that there are no externally defined macros expected in the DART headers. This should stay that way as the interface we expose is DASH and the user should not have to care about DART at all.

> Knowing that now I suppose that it would be nice to hide these additional warnings from within DASH,
as otherwise users that compile with -Wall/-Wextra stumble across our well-known and ignored warnings. I assume there is no easy way to achieve this if at all possible...

Eventually we should have DASH `-Wall`-clean to not confuse users with warnings. There have been some efforts in the past but I'm not sure what the current state is.

**Comment by fmoessbauer** [comment]
> Eventually we should have DASH -Wall-clean to not confuse users with warnings. There have been some efforts in the past but I'm not sure what the current state is.

I definitely vote for that and at least some time ago I fixed all warnings. However as nobody compiles in developer mode, there are many new warnings. Especially the locality part is full of such code.

File: dash/CMakeLists.txt [hunk 1]
```diff
@@ -255,6 +258,7 @@ foreach (dart_variant ${DART_IMPLEMENTATIONS_LIST})
   set (DART_LIBRARY "dart-${dart_variant}")
   set (DASH_LIBRARY "dash-${dart_variant}")
   set (VARIANT_ADDITIONAL_COMPILE_FLAGS ${ADDITIONAL_COMPILE_FLAGS_STR})
+  set (VARIANT_ADDITIONAL_COMPILE_FLAGS ${CXX_OMP_FLAG})
```
ddiefenthaler, line 261: @devreal I just noticed that there is a little problem with this line.
It drops the `ADDITINAL_COMPILE_FLAGS` by overwriting the variable.

Those flags include:
```
-DDASH
-DDASH_ENABLE_DEFAULT_INDEX_TYPE_LONG
-DDASH_ENABLE_THREADSUPPORT
-DDASH_ENABLE_UNIFIED_MEMORY_MODEL
-DDASH_ENABLE_TEST_LOGGING
-DDASH_ENABLE_PAPI
-DDASH_ENABLE_HWLOC
-DDASH_ENABLE_NUMA
-DDASH_ENABLE_HDF5
-DDASH_ENV_HOST_SYSTEM_ID='default'
```
I will provide a hotfix for this.

devreal: Gnarf, you're right. Please provide the fix in a PR. And thanks for taking care of it.

PR #433, opened 2017-08-15
Move dart_storage to DASH and fix size type

### LINKED ISSUE #432: dash::Array - not enough space allocated/accessible
Hey DASH Team!
I noticed that dash::Array has a problem with large array sizes!
It either leads to an error when trying to allocate or it seems to successfully allocate and on element access you get an segfault!

Since i got the problem in winnow (one of my cowichan implementations) I didn't write a dedicated test program for time reasons.
I give my best to describe it thoroughly and wrote some scripts to execute and see the problem at first hand... 
The location of the files is in [this](https://github.com/dash-project/dash-apps/tree/bug-ArrayAllocation/cowichan/winnow) dedicated branch.

The code in question is in "winnow.h".
important note:
When executing one of the scripts you should get some output to stdout.
All these outputs are debug outputs for just this problem case.
That's why they aren't in any DEBUG clauses!
And for gods sake don't execute winnow with this large input without the "--is_bench" flag!
The IO will take forever ^^ (but first you will have to provide the right input anyway...)
Okay back to the problem:
If you execute "crash_withError.sh" an Array of the size 225000000 (elements! one element should be 12 bytes big) is needed and should fail to allocate at [here](https://github.com/dash-project/dash-apps/blob/9e99bb80425ef1a4f9440ca2de404d46b162b111/cowichan/winnow/winnow.h#L293). At least it does crash for me at project03.
And that's the message i get:
`Fatal error in MPI_Alloc_mem: Invalid argument, error stack:
MPI_Alloc_mem(129): MPI_Alloc_mem(size=-1594967296, MPI_INFO_NULL, baseptr=0x7ffec70879b8) failed
MPI_Alloc_mem(87).: Invalid value for size, must be non-negative but is -1594967296`
Curious that he gets negative values at some point...! That's not what it gets from me.

Next case, next script: "crash_withoutError.sh". Despite the space needed for the same Array is even bigger (1600000000 elements) no error is thrown [here](https://github.com/dash-project/dash-apps/blob/9e99bb80425ef1a4f9440ca2de404d46b162b111/cowichan/winnow/winnow.h#L293).
But the [memcpy](https://github.com/dash-project/dash-apps/blob/9e99bb80425ef1a4f9440ca2de404d46b162b111/cowichan/winnow/winnow.h#L497) fails.
The reason is that the access to the elements of "toSort" is limited somehow... there seems to be less space than requested!
For example i can access [toSort[437700264]](https://github.com/dash-project/dash-apps/blob/9e99bb80425ef1a4f9440ca2de404d46b162b111/cowichan/winnow/winnow.h#L485) but not [toSort[437700265]](https://github.com/dash-project/dash-apps/blob/bug-ArrayAllocation/cowichan/winnow/winnow.h#L488).
for readability:
1600000000 (number of elements that should be accessible) >
437700265.
In "noCrash.sh" the input (or the resulting allocation...) seems to be small enough that the error doesn't occur. 
In "noCrash2.sh" the same input as in "crash_withError.sh" is used, but with 2 units -> the allocation per unit is smaller!

I need it to work with one unit as well, because I want to measure the speedup.
Hopefully i didn't forget anything... please feel free to ask!
(Note: CSR-Pattern doesn't support size_t as indextype )

devreal:
Fixes #432

### DIFF
--- a/dart-if/include/dash/dart/if/dart_types.h
+++ b/dart-if/include/dash/dart/if/dart_types.h
@@ -123,18 +123,6 @@ typedef enum
 #  error "Cannot determine DART type for size_t!"
 #endif
 
-
-/** \cond DART_HIDDEN_SYMBOLS */
-/**
- * \todo This is not part of the DART interface
- * and should move to dash::internal eventually.
- */
-typedef struct {
-    dart_datatype_t dtype;
-    int             nelem;
-} dart_storage_t;
-/** \endcond */
-
 /**
  * Data type for storing a unit ID
  * \ingroup DartTypes
--- a/dash/include/dash/GlobAsyncRef.h
+++ b/dash/include/dash/GlobAsyncRef.h
@@ -214,7 +214,7 @@ public:
       if (_is_local) {
         _value = *_lptr;
       } else {
-        dart_storage_t ds = dash::dart_storage<T>(1);
+        dash::dart_storage<T> ds(1);
         DASH_ASSERT_RETURNS(
           dart_get_blocking(
             static_cast<void *>(&_value), _gptr, ds.nelem, ds.dtype),
@@ -297,7 +297,7 @@ public:
     if (_is_local) {
       *tptr = *_lptr;
     } else {
-      dart_storage_t ds = dash::dart_storage<T>(1);
+      dash::dart_storage<T> ds(1);
       dart_get(static_cast<void *>(tptr), _gptr, ds.nelem, ds.dtype);
     }
   }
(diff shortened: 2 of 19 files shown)


ddiefenthaler (approved):
Looks good :+1:

rkowalewski:
I have not dived too deep into the details of #432 so I cannot really understand how it fixes this issue. Anyway, it is a good change to move this into DASH.

rkowalewski (approved):
Looks good to me either.

BenjaProg:
That fix works but i am still struggling with the other [problem/fix.](https://github.com/dash-project/dash/pull/434)

Issue #437, opened 2017-09-14
error: expression must have a constant value w/ Intel 18


jeffhammond:
I tried to compile DASH with Intel 18, which was released a few days ago.  I see some discussion of this topic in https://github.com/dash-project/dash/issues/231.  ICC is associated with the system default of GCC 5.4.1, which should be new enough to support C++11 features.
```
cd $HOME/Work/PGAS/DASH/github/build/dash && /opt/intel/compilers_and_libraries_2018.0.128/linux/mpi/intel64/bin/mpiicpc    -I$HOME/Work/PGAS/DASH/github/dash/include -I$HOME/Work/PGAS/DASH/github/dash/src -I$HOME/Work/PGAS/DASH/github/dart-if/include -I$HOME/Work/PGAS/DASH/github/build/gtest/src/GTestExternal/googletest/include  -DDART_HAVE_SYNC_BUILTINS -pthread -DDART_HAVE_PTHREADS -O3 -DNDEBUG  -std=c++17 -qopenmp -O3 -DDASH_RELEASE -DDASH_ENABLE_ASSERTIONS -DDART_ENABLE_ASSERTIONS -D_XOPEN_SOURCE=700    -qopenmp -DDASH -DDASH_ENABLE_DEFAULT_INDEX_TYPE_LONG -DDASH_ENABLE_THREADSUPPORT -DDASH_ENABLE_UNIFIED_MEMORY_MODEL -DDASH_ENABLE_TEST_LOGGING -DDASH_ENABLE_PAPI -DDASH_ENABLE_HWLOC -DDASH_ENABLE_BLAS -DDASH_ENV_HOST_SYSTEM_ID='default' -Wno-sign-compare -DMPI_IMPL_ID='intelmpi' -Wno-unused -Wno-sign-compare -o CMakeFiles/dash-test-mpi.dir/test/meta/ConstexprTest.cc.o -c $HOME/Work/PGAS/DASH/github/dash/test/meta/ConstexprTest.cc
$HOME/Work/PGAS/DASH/github/dash/test/meta/ConstexprTest.cc(98): error: expression must have a constant value
    constexpr auto arr_rev = dash::ce::reverse(arr);
                             ^
$HOME/Work/PGAS/DASH/github/dash/include/dash/util/ArrayExpr.h(290): note: access to uninitialized object
            : values);
              ^
$HOME/Work/PGAS/DASH/github/dash/include/dash/util/ArrayExpr.h(287): note: called from:
[...]

PR #438, opened 2017-09-17
[Fix] dash::ce::reverse Implementation

rkowalewski:
Addresses #437

### DIFF
--- a/dash/include/dash/util/IndexSequence.h
+++ b/dash/include/dash/util/IndexSequence.h
@@ -10,12 +10,31 @@
 namespace dash {
 namespace ce {
 
+
+/**
+ * Represents a compile-time sequence container
+ */
 template <
   std::size_t... Is >
 struct index_sequence
 { };
 
-
+/**
+ * Generates a compile-sequence integer (size_t) sequence in ascending order
+ *
+ * Example:
+ *
+ * template<std::size_t... Is>
+ * void print(dash::ce::index_sequence<Is...>&& s)
+ * {
+ *   for (auto i : { Is... }) std::cout << i << " ";
+ *   std::cout << "\n";
+ * }
+ *
+ * print(dash::ce::make_rev_index_sequence<5>());
+ * //output: 0 1 2 3 4
+ *
+ */
 template <
   std::size_t    N,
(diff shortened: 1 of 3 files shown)


dash/include/dash/util/IndexSequence.h
```diff
@@ -28,6 +29,20 @@ struct make_index_sequence<0, Is...>
   : index_sequence<Is...>
 { };
 
+template <
+  std::size_t    N,
+  std::size_t... Is >
+struct rev_index_sequence
+  : rev_index_sequence<N-1, Is..., N-1>
```
devreal: Shouldn't that be `rev_index_sequence<N-1, Is..., N>`? (So that `rev_index_sequence<3, 2, 1>` boils down to `index_sequence<1, 2, 3>`?) But I'm just guessing here from the documentation, though...

rkowalewski: No. That is why I call it rev_index_sequence (reverse index sequence). ;)

rkowalewski: No. That is why I call it rev_index_sequence (reverse index sequence). ;)

devreal: Ahh, I didn't fully grasp the meaning of `index_sequence` at first (I'll blame my lack of sleep). Some documentation might be useful though ;)

Also: `dash::detail::reverse` should be named `dash::detail::reverse_impl` for consistency. For the same reason, you might want to rename `rev_index_sequence` to `make_rev_index_sequence` but I have no strong feelings about that.

devreal (changes requested):

dash/include/dash/util/IndexSequence.h
rkowalewski: Good point. will do that. :)

rkowalewski:
@devreal I made some rework after your review comments and extended documentation.

rkowalewski:
I removed duplicated functionality (my fault) as well.

devreal (approved):

devreal:
Thanks a lot, good work! :)

Issue #439, opened 2017-09-18
dash::acumulate doesn't work for doubles


dhinf:
dash::accumulate uses size_t as element type for dash::Array and "auto result = 0".
Both convert the double values to an int.

Fix is on the way.

ddiefenthaler:
Fixed by PR #440

PR #440, opened 2017-09-18
now Accumulate works for double too

dhinf:
fixes issue: #439


dash/include/dash/algorithm/Accumulate.h
```diff
@@ -41,9 +41,9 @@ ValueType accumulate(
   auto l_first     = index_range.begin;
   auto l_last      = index_range.end;
   auto l_result    = std::accumulate(l_first, l_last, init);
-  auto result      = 0;
+  ValueType result = 0;
```
devreal: That should be `ValueType result = init` to support non-basic types.

dash/include/dash/algorithm/Accumulate.h
```diff
@@ -91,9 +91,9 @@ ValueType accumulate(
   auto l_first     = index_range.begin;
   auto l_last      = index_range.end;
   auto l_result    = std::accumulate(l_first, l_last, init, binary_op);
-  auto result      = 0;
+  ValueType result = 0;
```
devreal: That should be `ValueType result = init` to support multiplication as `binary_op` and non-basic types.

devreal (changes requested):

dash/include/dash/algorithm/Accumulate.h
rkowalewski: With @devreals changes included, we may revert to `auto result = init`? Just a guess.

rkowalewski (changes requested):

devreal:
A more general comment: in this implementation, `init` is not just used for initialization (once) but as the neutral element. Thus, the results of `std::accumulate(arr.begin(), arr.end(), 5)` and `dash::accumulate(arr.begin(), arr.end(), 5)` will differ. 

I suggest to modify the local accumulate to `std::accumulate(l_first+1, l_last, l_first, binary_op)` and only use `init` once to initialize the global result.

devreal:
I had some time during lunch break so I fixed it myself. I also had to remove the `double` test case before comparing `double` for equality is never a good idea. Added a test case using a `struct` instead.

What's missing:
1. Execution policies to support broadcasting the result to all units
2. Using `dart_reduce` for efficient handling of basic data types.

devreal:
Gnarf, I cannot reproduce the failed CI on my local machine. It seems to fail only for the combination of MPICH+NastyMPI, not sure what is going on there...

rkowalewski:
@devreal I will dive into this today.

devreal:
See also #216 which already mentioned some of the things fixed in this PR.

devreal (approved):

rkowalewski (approved):

PR #442, opened 2017-09-30
updated libs in docker-ci containers, added openmpi3 container

fmoessbauer:
This PR updates the libraries used in out docker-ci containers. A new container using OpenMPI 3.0.0 is added as well. This container is currently not used in CI but can be used for testing against OpenMPI 3.

**Note:** Compiling with PAPI and `-pedantic` will raise an error as there is still an unresolved issue regarding `noexcept` specifier. If necessary, switch back to PAPI 5.4.x.

## ci:openmpi (now deprecated)
- Papi: 5.4.1 -> 5.5.1
- OpenMPI: 1.10.5 -> 1.10.7
- HDF5: 1.10.0 -> 1.10.1

## ci:openmpi2

- Papi: 5.4.1 -> 5.5.1
- OpenMPI: 2.1.1 -> 2.1.2
- HDF5: 1.10.0 -> 1.10.1

## ci:openmpi3 (experimental)

- Papi: 5.5.1
- OpenMPI: 3.0.0
- HDF5: 1.10.1

## ci:mpich

- Papi: 5.4.1 -> 5.5.1
- mpich: 3.2
- HDF5: 1.10.0 -> 1.10.1

### DIFF
--- a/dash/scripts/docker-testing/mpich/dockerfile
+++ b/dash/scripts/docker-testing/mpich/dockerfile
@@ -18,10 +18,9 @@ WORKDIR       /tmp
 ENV           CC=gcc
 ENV           CXX=g++
 
-# Install PAPI from source
-# Compiler error when compiling with -pedantic for papi > 5.4.1
-# ADD         http://icl.utk.edu/projects/papi/downloads/papi-5.5.1.tar.gz papi.tgz
-ADD           http://icl.cs.utk.edu/projects/papi/downloads/papi-5.4.1.tar.gz papi.tgz
+# Compiler error when compiling with -pedantic for papi >= 5.5
+ ADD         http://icl.utk.edu/projects/papi/downloads/papi-5.5.1.tar.gz papi.tgz
+#ADD           http://icl.cs.utk.edu/projects/papi/downloads/papi-5.4.3.tar.gz papi.tgz
 RUN           tar -xf papi.tgz
 RUN           cd papi*/src/                     \
               && ./configure --prefix=/opt/papi \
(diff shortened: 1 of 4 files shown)


dash/scripts/docker-testing/mpich/dockerfile
```diff
@@ -18,10 +18,9 @@ WORKDIR       /tmp
 ENV           CC=gcc
 ENV           CXX=g++
 
-# Install PAPI from source
-# Compiler error when compiling with -pedantic for papi > 5.4.1
-# ADD         http://icl.utk.edu/projects/papi/downloads/papi-5.5.1.tar.gz papi.tgz
-ADD           http://icl.cs.utk.edu/projects/papi/downloads/papi-5.4.1.tar.gz papi.tgz
+# Compiler error when compiling with -pedantic for papi >= 5.5
+ ADD         http://icl.utk.edu/projects/papi/downloads/papi-5.5.1.tar.gz papi.tgz
+#ADD           http://icl.cs.utk.edu/projects/papi/downloads/papi-5.4.3.tar.gz papi.tgz
```
devreal, line 23: Are we not building with -pedantic anymore?

devreal (approved):
Looks good to me. Are the tests for Open MPI 3.0.0 activated automatically?

dash/scripts/docker-testing/mpich/dockerfile
fmoessbauer: As the flag is not listed in the `compile_commands.json`, someone must have removed it. However we could compile with `-Wpedantic`, so only warnings are generated. Probably we should discuss this, but this exceeds this PR. Anyway, I'm not a fan of using pedantic as this often clashes with third-party libraries. And at least for PAPI we are in a dead-end.

The OpenMPI 3.0.0 tests are currently not enabled in CI, as this would require a third CI container (currently mpich, openmpi2). As we only have 4 containers, this would allow us only to execute one CI job at a time (instead of 2).

Our tests seem to work also with OpenMPI3, except some errors in the HDF5 API.

rkowalewski (approved):
Thanks for this. Looks good to me.

PR #444, opened 2017-10-01
fixed pattern in MatrixTest.DelayedAlloc ...

fmoessbauer:
… to be valid for any number of units. Before this change, the test failed if run with 4 units.


devreal (approved):
See also #420 for this (this PR might fix it)

fmoessbauer:
@devreal Ok, I did not notice that there is already some work going on. For the meantime I merge this PR. When #420 PR is ready, we can replace my changes by yours.

PR #445, opened 2017-10-03
Seperate build tools directory

ddiefenthaler:
### Update:
As this pull request changes the build scripts quite a bit here is a summarized changelog:

- Build scripts moved to a new folder `build-tools` with `build.sh` in the root remaining as symlink.
- Changed the build configurations from seperate independant scripts to configuration files which get included in a main build script. To build with configuration `dev` for example the call changes from
`./build.dev.sh` to `./build.sh dev`. This call also works from the root of the repository.
- This also works for the `my.build.<type>.sh` scripts. If there is a `my.build.dev.sh` it is used on top of the `build.dev.sh` configuration when called with `./build.sh dev`.
- Added more command line options, see also `./build.sh --help`:
  - `--purge`: purge the build-directory, this does not happen by default anymore.
     To restore the previous default put the line `PURGE=true` in the `my.build.default.sh` file.
     Create it if it does not exists. Make sure that it is executable. You could also use `--edit default`.
  - `--force`: as before, skip the confirmation step.
  - `--nobuild`: stop after the confirmation without building
  - `--install`: call an additional install command instead of just mentioning the command.
  - `--list`: list available configurations
  - `--edit <type>`: create a new configuration/edits a configuration if it already exists.
- Updated the `README` for the new call syntax of the build script.
- Removed the `syntastic`-files. Currently it is not planned to add them back again.
  Please leave a review requesting them, if you need them created dynamically.

### Original post:

As discussed earlier today/yesterday:
- Moved build-scripts to a seperate directory,
  additionally the build-scripts now work independant from the current working directory,
  as long as the scripts remain in the `build-tools` folder
- Removed the syntastic-files as they are to static
  It should be possible to generate them with cmake.
  I haven't used them yet, but for me it looks like the variable [`CXXFLAGS_WRAP`](https://github.com/dash-project/dash/pull/431/files#diff-40505a078802c8427ec7dab9abd8a29cR20) from PR #431 already collects most of the values that would go in those files.

Please check if the build script still work in different environments, as it uses the non-POSIX `readlink`.


build-tools/build.analyze.sh
```diff
@@ -6,7 +6,18 @@
...
 # export CCC_CXX=clang++3.8
 
-BUILD_DIR=./build.analyze
+SCRIPT=$0
+while [ -h $SCRIPT ]
+do
+  BASE=$(dirname $SCRIPT)
+  SCRIPT=$(readlink $SCRIPT)
+  cd $BASE
+done
```
devreal: What is this while loop for? (please add comment)

To get the absolute path of the current script, you can also do something like `base=$(cd $(dirname $SCRIPT) && pwd)`

build-tools/build.nasty.sh
```diff
@@ -1,6 +1,17 @@
...
+SCRIPT=$0
+while [ -h $SCRIPT ]
+do
+  BASE=$(dirname $SCRIPT)
+  SCRIPT=$(readlink $SCRIPT)
+  cd $BASE
+done
+
+cd $(dirname $SCRIPT)/..
+DASHDIR=$(pwd -P)
```
devreal: `DASHDIR` --> `ROOTDIR`

devreal (changes requested):
Thanks for taking the effort!

We could put the syntastic files as a template into a folder (like `supp`) so that people can use them if they want.

build-tools/build.analyze.sh
ddiefenthaler: Sorry for the lack of documentation, will add it.

It is for the symlink in the rootdir, when calling `dirname` the path is relative to the symlink and not to the build-tools dir.

ddiefenthaler:
Looking at the documentation of syntastic ([Configuration-Files#location](https://github.com/vim-syntastic/syntastic/wiki/Configuration-Files#location)),
it seems that those files have to go in the root directory to work out of the box.

I would wait for this until your PR #431 is merged and use the code in it to create them on the fly.
They should probably go into `.gitignore` then as well...

Looking at the amout of build scripts we have and the duplication of the code resulting from that,
it probably would also be useful to extend the parameters that `build.sh` takes to reduce that.
For the ease of use they would be optional.
I will look into this tomorrow.

build-tools/build.analyze.sh
devreal: It looks like the `--canonicalize` flag to `readlink` would do just that? (assuming that the while loop is meant to traverse the symlink chain down to the real file)

ddiefenthaler: Yes, but unfortunately this option is not available on all systems,
namely probably all BSD derivates.

But your right I'll put the loop in a function, with an appropriate name.

ddiefenthaler:
With the last commit I unified the different build scripts.
The main script file now just calls the `cmake && make`. The different configurations are in seperate script files which get sourced.

To use them just pass the name of the configuration as parameter. For example:
`./build.sh debug` for `build.debug.sh`. If `my.build.debug.sh` exists it is used additinally.
`./build.sh foo` will look for `build.foo.sh` and `my.build.foo.sh` and use them if they exist,
To add a new configuration just copy an existing one and adjust it to your needs.

Please check if all available scripts still work,
@fmoessbauer especially your scripts `cov` and `analyze` as I have no environment set up to use them.

Also I removed the `cp compile_commands.json` from the script as I can't see why the file would be needed in the root directory of DASH. It can still be found in the respecting build directory.

devreal:
One thing I noticed: the `build` script tells the user to call `make install` in the corresponding `build` directory. Why not have a `install.sh` script that does this? This might be minor but is more consistent imo.

ddiefenthaler:
Agreed, but I would aim at a function within the build script.
A seperate script would probably need quite a bunch of duplicate code again.
Currently I'm thinking of something like:
```
Clean build directory? [Y/n/q]
Reconfigure with cmake? [Y/n/q]
Build? [Y/n/q]
Install? [Y/n/q]
```
While at it I'll probably add options to directly perform those steps. Also a silent option may be worth a thought.

I'll try to realize it this afternoon.
Also as PR #431 is merged now I'll look into the syntastic-files. Even though I would like some advice from somebody who actually uses it, as I'm not sure if I get the filenames or all the needed options right.

ddiefenthaler:
Sorry that it took a little bit longer.

The build script now asks three questions,
```
   Purge build directory before building? [y/N]
   Build with these settings? [Y/n]
   Done. Install DASH now? [y/N]
```

There are also matching parameters for them.
On top of that there are options to list and edit build types.
See `--help` for more information.

Here are some examples:
- To get the same behaviour as the old scripts call it with `./build.sh -pf`.
- To stop before building, but still ask befor purging call `./build.sh -n`.
- To list all available build type: `./build.sh -l`.
- To edit a new build type `example`: `./build.sh -e example`.
The created file is a duplicate of `default` before editing

@devreal Your comments should be addressed for now.

On the topic of syntastic, are those files still needed by anybody?
If nobody uses them I would like to circumvent writing an useless script.

fuchsto:
I know I'm the one who introduced the "confirm" stuff, but all this "Are you sure?" is getting out of hands.
When I run `build.sh`, I want to build.
The reason why I added the confirmation in the first place was to give an opportunity to review the build settings, just like in childhood days with `configure` and `make`.
Leave the original confirmation but please replace "Purge build directory?" and "Install DASH?" by command line arguments like `clean` and `install`.

fuchsto (changes requested):
I know I'm the one who introduced the "confirm" stuff, but all this "Are you sure?" is getting out of hands.
When I run build.sh, I want to build.
The reason why I added the confirmation in the first place was to give an opportunity to review the build settings, just like in childhood days with configure and make.
Leave the original confirmation but please replace "Purge build directory?" and "Install DASH?" by command line arguments like clean and install.

ddiefenthaler:
There are already command line arguments `--purge` and `--install` that do exactly that.
Thanks for the feedback, that is an easy change.

ddiefenthaler:
I updated the first post of this pull request, so that there is a summarized changelog directly at the top.

@fuchsto The additional confirmations are removed again. The already mentioned command line options still work the same.

devreal:
Ahh, sorry for the long time, this PR totally flew under my radar... 

I tested it and it works well for me. I took the freedom to fix some slight inconsistencies. Otherwise great work!

devreal (approved):

ddiefenthaler:
Thanks, I also forgot about this PR as it is not that important.

I guess the inconsistencies were just inherited from the original build scripts. Thanks for fixing it.

devreal:
Should this PR still be merged or is it dead?

ddiefenthaler:
This PR got (offline) negative feedback.
To some degree I can see the point that a build script shouldn't be too complex.

Yet those reasons aren't documented here. I'm not really able to do so either as I do not really see the problem. But it is probably useful to keep the build helpers simple and close this PR.
The clean-up of the root of the repository has to be done in an other PR then, if still needed.

Issue #376, opened 2017-04-28
Synchronization at the end of DASH algorithms


devreal:
While working on #374 I realized that some algorithms in DASH synchronize units at the end (e.g., `dash::for_each`) and some don't (e.g., `dash::generate`). This should be unified. 

@fuchsto I assume this is covered by Execution Policies (#104)? If so, this is just to track progress on this issue.

fuchsto:
Yes, exactly, but it's sensible to track this as a separate issue.

PR #446, opened 2017-10-04
[WIP] Implement dash::sort()

mmue:


dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,217 @@
...
+{
+  auto index_range  = dash::local_index_range(first, last);
+  auto lbegin_index = index_range.begin;
+  auto lend_index   = index_range.end;
+  auto & pattern     = first.pattern();
+  auto lrange_begin = (first + pattern.global(lbegin_index)).local();
+  auto lrange_end   = lrange_begin + lend_index;
+
+  auto n = dash::distance(first, last);
+  auto p = dash::size();
```
devreal, line 35: Please use more descriptive variable names.

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,217 @@
...
+
+  ElementType first_possible_key = * dash::min_element (first, last);
+  ElementType last_possible_key  = * dash::max_element (first, last);
+  ElementType diff = (abs(last_possible_key - first_possible_key) / p) * 0.1;
```
devreal, line 39: `diff` does not seem to be the right name here. Where does the  factor `0.1` come from?

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,217 @@
...
+
+  dash::barrier();
+
+  if (p == 1) {
+    return std::sort(first, last);
```
devreal, line 44: Do this before the setup above.

devreal (commented):
Is this PR ready for review? (I've added some comments inline before realizing that no review has been requested so far)

fmoessbauer:
@mmue presented this solution yesterday to the Munich group. Currently this algorithm is not working correctly (MPI errors, probably due to bugs in DASH's CSR pattern), but we are investigating this.
Anyways there are various points which are not optimal regarding both the C++ and DASH side of the implementation. However this should be easily fixable and as a whole this algorithm looks promising.

dash/include/dash/algorithm/Sort.h
fmoessbauer: And use team relative sizes / ids:
```c++
p    = first.team().size();
myid = first.team().myid();
```

devreal:
Thanks for the clarification. It looks like this algorithm is only suitable for interger/floating point values but not derived data types. I guess we'd need another algorithm for derived data types.

dash/include/dash/algorithm/Sort.h
fmoessbauer: And also check if the whole range resides on one unit:

```cpp
if(first.is_local() && (last-1).is_local()){
  return std::sort(first.local(), (last-1).local()+1);
}
// + tell or check on the other units that they can return without sorting
```
the -1 trick is necessary as the global last iterator might point to the next unit, but `sort()` takes a past-end iterator, hence this iterator does not have to be dereferenced.

By the way: if the local part of a unit is empty, do nothing, but do not return directly as the algorithm is some kind of collective operation and hence should end with a barrier. This holds as long as issues #376 and #104 are implemented.

fmoessbauer:
@devreal : the algorithm is suitable for all datatypes which can be mapped to integers / natural numbers. This is a requirement of (almost) all efficient distributed sorting algorithms. Floating point numbers could be problematic as they are not unique. But this could probably be solved using a threshold.

fmoessbauer:
@mmue while trying to get the MPI error (on shared memory systems I was not able to get the error, even if MPI is used for communication), I noticed that the algorithm does not sort properly. Just add these lines of code [here](https://github.com/mmue/dash/blob/3e9ecf421962e9758481fe53d538228eba538c07/dash/examples/bench.13.sort/main.cpp#L147).

```cpp
  // PROCEDURE STEP 6 ----------------------------------------------------
  // "Validate Output"
  //

  // validate local part
  bool validated = true;
  for(decltype(key_array)::index_type i=0; i<key_array.lsize()-1; ++i){
    if(key_array.local[i] > key_array.local[i+1]){
      validated &= false;
      break;
    }
  }
  if(!validated){
    std::cout << "Validation failed on unit "
              << static_cast<int>(dash::myid())
              << std::endl;
  }
// TODO: Check the boundaries
```

fmoessbauer:
Closed in favor of #467.

Issue #390, opened 2017-05-09
RFC: flush-free async put/get using stream operators


devreal:
During a short discussion on our interface(s) for asynchronous transfer today, we agreed that an interface using `dash::Future` is desireable (and on the way in #371) and that we also want to have a nice C++'ish interface for transfers for which the user provides the buffer. Right now this is done using `dash::Array::async.get(T*)`, which is rather C-style. 

My proposal: use the stream operators on `dash::Array::async` to kick off a transfer and any `flush`/`barrier` on the container to wait for completion. Example:

```
int val;
dash::Array<int> array(N);
// fill array...

// kick-off transfer
array.async[0] >> val;
// do some computation

// wait for transfer to complete
// array.async.flush() or
array.flush();
// val can be used now
std::cout << val;
```

The streaming operators seem to be a good choice because the `istream` and `ostream` stream buffers are known concepts to users. In our case though, the variables cannot be re-used before the `flush()` operation but that is a matter of documentation, imo. In contrast to `dash::Future`, the overhead should be smaller as it does not require creating intermediate objects.

It's just an idea I wanted to put out for discussion (maybe it's floating around somewhere already?) and I am looking forward to your comments.

dhinf:
Although i like the current syntax a tiny bit more, I would like to use the proposed one in future. We get rid of the pointer and i think it is more obvious to the user what happens.

fuchsto:
Prettyfullness aside, using the stream operator for something that's not a stream is ... I'm sweating.

Seriously though, Is the operator precedence of `>>`, `<<` okay?
And you wrote it yourself: 

> The streaming operators seem to be a good choice because the istream and ostream stream buffers
> are known concepts to users.

Would be devious then if we would provide the operator ... but not stream semantics ... *pets white persian cat*

> In our case though, the variables cannot be re-used before the flush() operation but that is a matter
> of documentation

*insane laughter, raising steel fist prosthesis, persian cat runs away*

devreal:
*Sigh*. Cynicism make discussions so much more fun. 

I regard these operators as a nice way of capturing the address of the target or source variable without having to poke around C-style pointers. But well, I agree that it's not exactly stream semantics... 

Maybe someone has a better idea. Until then, we will have to cope with ugly pointers (and no, I do not regard something like `async.get(T&)` as any better, I think it's even worse than clear pointer syntax).

fuchsto:
No no! Sorry, I wasn't being cynical. Not intentionally, at least.
[...]

PR #447, opened 2017-10-13
Fix GlobAsyncRef

### LINKED ISSUE #365: GlobAsyncRef flush via container.flush() does not work

devreal:
This PR is (yet another) attempt to get `GlobAsyncRef` into a stable state. As discussed at the F2F, value semantics are removed and no consistency guarantees are provided, i.e., a read does not necessarily yield a previously written value unless there has been a flush in between. 

Asynchronous writes use a temporary value and a handle, which is used to ensure local completion upon destruction of the `GlobAsyncRef` object. Reads are never asynchronous as that would require speculatively starting a transfer even if the `GlobAsyncRef` would be used for writing only. With value semantics removed, this is now explicit by calling `get()`. 

The following use-case is covered: 

```

for (size_t i = 0; i < dash::size(); ++i) {
  arr.async[i] = i;
}
arr.flush(); // flush asynchronous transfers to all units
assert(arr[dash::myid()] == dash::myid()); // succeeds
assert(arr.async[dash::myid()].get() == dash::myid()); // succeeds
```

What does not work: 
```

for (size_t i = 0; i < dash::size(); ++i) {
  arr.async[i] = i;
}
assert(arr[dash::myid()] == dash::myid()); // might fail
assert(arr.async[dash::myid()].get() == dash::myid()); // might fail
```

The PR also fixes the semantics of `dash::Matrix::flush*` and `dash::Array::flush*` and removes `push` and `fetch`, which have not been used.

Related: #357 #358 #409
Fixes #365 
Closes #371 #390


devreal:
Update: I also removed the comparison operators to avoid hidden synchronous `get()` operations upon comparison. I also removed the reference-based comparison operators to avoid the following pitfall:

```
dash::Array<int> arr(...);
dash::fill(arr.begin(), arr.end(), 0);
if (arr.async[0] == arr.async[1]) { // always false 
  // ...
}
```

I noticed that `GlobRef` has the same behavior. IMO, reference-based comparison should be explicit on global references, i.e., `arr.async[0].dart_gptr() == arr.async[1].dart_gptr()`. Should we adapt the behavior of `GlobRef` as well?

dash/include/dash/GlobAsyncRef.h
```diff
@@ -315,203 +222,72 @@ class GlobAsyncRef
...
-        dash::dart_storage<T> ds(1);
-        DASH_ASSERT_RETURNS(
-          dart_put(
-            _gptr, static_cast<const void *>(&_value), ds.nelem, ds.dtype),
-          DART_OK
-        );
-      }
+   * but the value referenced by \c new_value can be re-used immediately.
+   */
+  void set(const_value_type& new_value) {
```
fmoessbauer, line 244: Does this renaming break anything? Should we provide an alias `put()`?

dash/include/dash/Matrix.h
```diff
@@ -417,30 +417,32 @@ class Matrix
...
+   * on the container's underlying global memory.
    *
    * \see  DashContainerConcept
    */
   inline void                 flush();
 
   /**
-   * Complete all outstanding non-blocking operations executed by the
-   * local unit on the narray's underlying global memory.
+   * Complete all outstanding non-blocking operations to the specified unit
```
fmoessbauer, line 428: Really cool feature. This fixes / answers #416.

fmoessbauer (approved):
This PR looks really good. The interface changes are all reasonable and required. Thanks for putting time into that.

Just one note: The move construction of `GlobAsyncRef` is defaulted, but there is no move-assignment operator. IMO we can default this as well. However I have not thought about possible impact, especially if the target `GlobAsyncRef` should be flushed before the assignment.

Regarding your question in the comments: Yes, I vote for changing the behavior of `GlobRef` as well.

And last but not least, we should check that all dash-apps still compile.

dash/include/dash/Matrix.h
devreal: Not entirely, unfortunately. We still need a beefed up `dash::Future` for bulk wait to not leak memory through the handles. It's on my list :)

dash/include/dash/GlobAsyncRef.h
devreal: I will check. IMO, `put()` is closer to a C PGAS API.

PR #450, opened 2017-10-17
GlobRef: Remove reference-based comparison and postfix operators

devreal:
Changes to `GlobRef`:

1. Remove reference-based comparison operator to perform value-based comparison and allow a test like `arr[0] == arr[1]` to succeed. Add test. See also #447 
2. ~Remove postfix increment/decrement operators because their semantic cannot be correctly implemented, i.e., `(arr[0]++)++` will not perform two increments and return the result of the first increment. It's better to error-out than to yield incorrect results.~ After thinking about this again I realized that it's perfectly reasonable to have these operators on global references as they return a value and not a reference and `(arr[0]++)++` is illegal anyway. Re-added the operators.


devreal:
Fix for the failing test case is contained  in #447

fuchsto:
The increment-operator issue is a really good catch. The same situation with `operator+=`, and any `operator@=` for that matter. For asyncs, these should return a future on the operation. I will meditate on that.

devreal:
> The same situation with operator+=, and any operator@= for that matter.

Not sure I can follow: `operator+=` should return the result of the addition-assignment so returning a reference should work as expected. The catch really only is the postfix increment since that is supposed to return the previous state, which cannot be done by returning a reference.

fmoessbauer (approved):

PR #434, opened 2017-08-15
Chunk up puts/gets to cover large transfers

devreal:
Chunk up puts/gets in `dart_put`, `dart_put_blocking`, `dart_get`, and `dart_get_blocking`.

This change is required because MPI does not expect users to copy more than 2GB of data.

This PR is incomplete as `dart_put_handle` and `dart_get_handle` are not covered. There is a host of changes to the way handles are handled in DART waiting for review in #419 so this would only create unnecessary conflicts. I will adjust them accordingly as soon as #419 is merged.

This PR does not include a test case because such large allocations won't work on small systems and thus render the test suite unusable.

@BenjaProg Your test case should be covered, IIRC the problem for you was in `dart_get_blocking`. Please test whether this works now.

Related to #432.

PR #410, opened 2017-05-20
[WIP] Fixing and improving algorithms with view expressions

fuchsto:
Addresses #346, among others.

A diff that demonstrates the benefit of view expressions pretty well:
<https://github.com/dash-project/dash/pull/410/files#diff-c4c2bc3b549c09735c61251a8f8d8ecfR1149>

With view expressions, `dash::copy(lbegin, lend, gbegin)` is now supported and also allows non-contiguous global destination ranges (blocked, strided, anything a pattern might define).
Its full implementation:

~~~c++
template <class ValueType, class GlobOutputIt>
GlobOutputIt copy(
  const ValueType  * in_first,
  const ValueType  * in_last,
  GlobOutputIt       out_first) {
  // Return value, initialize with begin of output range, indicating no
  // values have been copied:
  GlobOutputIt out_last   = out_first;
  // Number of elements to copy in total:
  auto num_elements       = std::distance(in_first, in_last);
  // Global iterator pointing at hypothetical end of output range:
  GlobOutputIt out_h_last = out_first + num_elements;
  // Create range from global output iterators:
  auto out_range  = dash::make_range(out_first, out_h_last);
  // Be happy:
  auto out_blocks = dash::blocks(out_range);
  auto in_copy_it = in_first;
  for (auto block : out_blocks) {
    out_last   = static_cast<GlobOutputIt>(
                   dash::internal::copy_block(
                     in_copy_it,
                     in_copy_it + block.size(),
                     block.begin()));
    in_copy_it = in_first + dash::distance(out_first, out_last);
  }
  return out_last;
}
~~~

Also added example `ex.02.array-views` for illustration:

https://github.com/dash-project/dash/blob/feat-192-nviews/dash/examples/ex.02.array-views/main.cpp

~~~c++
  // Note that global range is strided:
  dash::Array<float> a(array_size, dash::BLOCKCYCLIC(3));
  initialize_array(a);

  auto copy_num_elem       = a.size() / 2;
  auto copy_dest_begin_idx = a.size() / 4;
  auto copy_dest_end_idx   = copy_dest_begin_idx + copy_num_elem;

  std::vector<float> buf(copy_num_elem);
  std::iota(buf.begin(), buf.end(), 0.9999);

  a.barrier();

  if (dash::myid() == 0) {
    print("array: " << range_str(a));
    print("copy target index range: "
          << "[" << copy_dest_begin_idx << "," << copy_dest_end_idx << ")");

[...]

PR #451, opened 2017-10-17
Improvements to dash::Future

### LINKED ISSUE #416: Bulk wait for async copy
In the [hpccg](https://github.com/dash-project/dash-apps/blob/hpccg-coarray/hpccg/src_dash/exchange_externals.cpp#L69) benchmark the following is used to wait for a bulk of `dash::copy_async` transfers:

```c++
for(int i=0; i<num_neighbors; ++i) {
    int n_send = send_length[i];
    int neighbor = neighbors[i];
    int offset_at_neighbor = A->offset[i];
    dash::copy_async(send_buffer, send_buffer + n_send,
      narray.row(neighbor).begin() + A->offset[i]);
    send_buffer += n_send;
  }

  // wait for the transfer to complete
  narray.flush_all();
```

Is this a valid way to wait for the transfers? Otherwise I would have to use (e.g.) a `std::vector<VeryComplicatedType>` to store the futures. But as indicated, then the type of the vector might be very complicated, as `auto` cannot be used.

devreal:
This PR extends `dash::Future` to allow for a meaningful `test()` function and proper tear-down. The latter is required to avoid memory leaks if `dash::copy_async` uses handles and the user does not `wait()` on it. 

Fixes #416 
See also #447


fmoessbauer (commented):
Your changes look good so far, however I have some concerns regarding the performance of multiple relatively small async copy calls. Are there already benchmark results comparing before and after? If not, we could easily use the DASH implementation of Cafbench therefor.

devreal:
> Are there already benchmark results comparing before and after? If not, we could easily use the DASH implementation of Cafbench therefor.

I don't think we have a proper benchmark for it but it would be easy to create one. I might give it a try next week. 

However, I don't immediately see how we could improve performance without sacrificing features. The current implementation of `dash::Future` is very basic and fails a simple use-case like `while (fut.test()) do_something_else();`, not to mention the potential for memory leaks. 

I am open to ideas on how to save the dynamic allocation of the `std::vector` holding the handles (which probably is the critical addition in this PR).

fuchsto:
Did not have a look at the implementation yet but the PR is a good catch.
More important than benchmarks for now: It's a new feature -> add unit tests!

devreal:
@fuchsto Good point. Did that, made me change the implementation even more:

1. Made `dash::Future` non-copyable, similar to `std::future`. We might want to add `dash::SharedFuture` at some point but that's a different story.
2. Consolidated `dash::internal::copy_impl` and `dash::internal::copy_async_impl` into a single implementation to i) reduce duplicated code and ii) overlap the transfer to/from multiple remote units. There is still a metric ton of duplicated code in `Copy.h` but I didn't bother touching that as I expect it to change with #410 anyway. 
3. Added `dart_testall` and `dart_test`, which upon success also ensure remote completion if required.
4. All copy implementations now use handles: `dash::copy` waits for their (remote) completion while `dash::copy_async` puts them into a future. 
5. Handling of large transfers is now left to DART (see #434), slightly simplifying the code in `copy_impl`.

fmoessbauer (approved):
This PR looks good and IMO we can merge it. As the implementation of copy will change anyways I have no problem with using handles there. In the future we can improve that by providing optimized copy for tiny transfers.

The changes related to `dash::Future` are really good ones.

PR #452, opened 2017-10-20
Feat halo

dhinf:
get HaloMatrixWrapper out of experimental
- adapted tests and examples
- removed old Halo.h
- removed GlobStencilIter + tests for GlobStencilIter (Tobias please have look at it)


dart-impl/mpi/src/dart_communication.c
```diff
@@ -550,6 +550,317 @@ dart_ret_t dart_compare_and_swap(
...
+    return DART_ERR_INVAL;
+  }
+
+  handle->num_reqs++;
+  *handleptr = handle;
+
+  DART_LOG_TRACE("dart_get_strided_handle > handle(%p) dest:%d win:%"PRIu64,
+                 (void*) handle, handle->dest, (unsigned long)win);
+
+  MPI_Type_free(&mpi_strided_type);
```
devreal: Please make sure the stride is free'd in all cases (including errors).

dart-if/include/dash/dart/if/dart_communication.h
```diff
@@ -473,46 +516,46 @@ typedef struct dart_handle_struct * dart_handle_t;
...
- * \param nelem  The number of elements of type \c dtype to transfer.
- * \param dtype  The data type of the values in buffer \c dest.
- * \param[out] handle Pointer to DART handle to instantiate for later use with \c dart_wait, \c dart_wait_all etc.
- *
- * \return \c DART_OK on success, any other of \ref dart_ret_t otherwise.
- *
- * \threadsafe
- * \ingroup DartCommunication
- */
+void            * dest,
```
devreal: Please undo the unindent (probably done by your editor)

devreal (changes requested):
Only looked at the DART part for now, please see my two cents inline. I will look at the rest later

dart-if/include/dash/dart/if/dart_communication.h
```diff
@@ -473,11 +516,11 @@ typedef struct dart_handle_struct * dart_handle_t;
  * \ingroup DartCommunication
  */
 dart_ret_t dart_get_handle(
-  void            * dest,
-  dart_gptr_t       gptr,
-  size_t            nelem,
-  dart_datatype_t   dtype,
-  dart_handle_t   * handle) DART_NOTHROW;
+void            * dest,
```
devreal: Some more whitespaces accidentally removed :)

dash/include/dash/halo/Halo.h
```diff
@@ -0,0 +1,1062 @@
...
+}; // StencilSpec
+
+
+
+/**
+ * Specifies the values global boundary Halos
+ */
+enum class Cycle : uint8_t {
+  /// No global boundary Halos
+  NONE,
```
devreal, line 74: Are we sure there is no name conflict on `dash::NONE`? Can we move this to `dash::Stencil::NONE`?

dash/include/dash/halo/Halo.h
```diff
@@ -0,0 +1,1062 @@
...
+      index = coords[i] + index * 3;
+
+    return index;
+  }
+
+  static CoordsT coords(const region_index_t index) {
+    CoordsT coords{};
+    region_index_t  index_tmp = static_cast<long>(index);
+    for (auto i = (NumDimensions - 1); i >= 1; --i) {
+      auto res  = std::div(index_tmp, 3);
```
devreal: Where does the constant `3` come from?

dash/include/dash/halo/Halo.h
```diff
@@ -0,0 +1,1062 @@
...
+      while (nextRegion(stencil, stencil_combination)){
+        setRegionSpec(stencil_combination);
+      }
+    }
+  }
+
+
+  template<typename... ARGS>
+  HaloSpec(const ARGS&... args) {
+    std::vector<HaloRegionSpecT> tmp{args...};
```
devreal: Better use `std::array` here

dash/include/dash/halo/Halo.h
```diff
@@ -0,0 +1,1062 @@
...
+  }
+
+  const SpecsT& haloSpecs() const{
+    return _specs;
+  }
+
+private:
+  void setRegionSpec(const StencilT& stencil) {
+        auto index = HaloRegionSpecT::index(stencil);
+        auto max = stencil.max();
```
devreal: Indentation!

dash/include/dash/halo/Halo.h
```diff
@@ -0,0 +1,1062 @@
...
+  using pointer         = PointerT;
+  using const_pointer   = const pointer;
+
+public:
+  /**
+   * Constructor, creates a block boundary iterator on multiple boundary
+   * regions.
+   */
+  RegionIter(GlobMemT& globmem, const PatternT& pattern, const ViewSpecT& _region_view,
+             index_type pos, index_type size)
```
devreal: `index_type size` -> `size_type size`

dash/include/dash/halo/Halo.h
```diff
@@ -0,0 +1,1062 @@
...
+      : _globmem(globmem), _pattern(pattern), _region_view(_region_view), _idx(pos), _size(size),
+        _max_idx(_size - 1), _myid(dash::myid()), _lbegin(globmem.lbegin()) {}
+
+  /**
+   * Copy constructor.
+   */
+  RegionIter(const SelfT& other) = default;
```
devreal: What about move assignment/construction?

dash/include/dash/halo/Halo.h
```diff
@@ -0,0 +1,1062 @@
...
+
+public:
+  /**
+   * Constructor, creates a block boundary iterator on multiple boundary
+   * regions.
+   */
+  RegionIter(GlobMemT& globmem, const PatternT& pattern, const ViewSpecT& _region_view,
+             index_type pos, index_type size)
+      : _globmem(globmem), _pattern(pattern), _region_view(_region_view), _idx(pos), _size(size),
+        _max_idx(_size - 1), _myid(dash::myid()), _lbegin(globmem.lbegin()) {}
```
devreal: The iterator should be team-aware, use `pattern.team.myid()` instead of `dash::myid()`

dash/include/dash/halo/Halo.h
```diff
@@ -0,0 +1,1062 @@
...
+  }
+
+  SelfT operator+(index_type n) const {
+    SelfT res{*this};
+    res += n;
+
+    return res;
+  }
+
+  index_type operator+(const SelfT& other) const { return _idx + other._idx; }
```
devreal: Formatting! 

Plus: why does this `operator+` return `index_type`?

Same holds true for the below operator.

dash/include/dash/halo/Halo.h
```diff
@@ -0,0 +1,1062 @@
...
+
+  const ViewSpecT _region_view;
+  /// Iterator's position relative to the block border's iteration space.
+  index_type _idx{0};
+  /// Number of elements in the block border's iteration space.
+  index_type _size{0};
+  /// Maximum iterator position in the block border's iteration space.
+  index_type _max_idx{0};
+  /// Unit id of the active unit
+  dart_unit_t _myid;
```
devreal: Use `dash::team_unit_t` here

dash/include/dash/halo/Halo.h
```diff
@@ -0,0 +1,1062 @@
...
+
+public:
+  using iterator        = RegionIter<ElementT, PatternT>;
+  using const_iterator  = const iterator;
+
+public:
+  Region(const HaloRegionSpecT& region_spec, const ViewSpecT& region, GlobMemT& globmem,
+         const PatternT& pattern, const BorderT& border)
+      : _region_spec(region_spec), _region(region), _border(border),
+        _border_region(std::any_of(border.begin(), border.end(), [](bool border_dim){ return border_dim == true;})),
```
devreal: line length >80 chars?

dash/include/dash/halo/Halo.h
```diff
@@ -0,0 +1,1062 @@
...
+  BorderT               _border;
+  bool                  _border_region;
+  iterator              _beg;
+  iterator              _end;
+}; // Region
+
+template <typename ElementT, typename PatternT>
+class HaloBlock {
+private:
+  static const dim_t NumDimensions = PatternT::ndim();
```
devreal: Can this be `constexpr`?

dash/include/dash/halo/Halo.h
```diff
@@ -0,0 +1,1062 @@
...
+
+  const ViewSpecT& view_safe() const { return _view_safe; }
+
+  const ViewSpecT& view_inner() const { return _view_inner; }
+
+  const std::vector<ViewSpecT>& boundary_elements() const { return _boundary_elements; }
+
+  size_type halo_size() const {
+    return std::accumulate(_halo_regions.begin(), _halo_regions.end(), 0,
+                           [](size_type sum, region_t region) { return sum + region.size(); });
```
devreal: Pass `region` by reference?

dash/include/dash/halo/HaloMatrixWrapper.h
```diff
@@ -0,0 +1,307 @@
...
+
+  struct Data {
+    const HaloRegionT&             region;
+    std::function<void(HaloData&)> get_halos;
+    HaloData                       halo_data;
+  };
+
+  void updateHaloIntern(Data& data, bool async) {
+    auto rel_dim = data.region.regionSpec().relevantDim() - 1;
+    if (data.region.borderRegion() && data.region.borderDim(rel_dim) && _cycle_spec[rel_dim] == Cycle::FIXED)
```
devreal: line length

devreal (changes requested):
Good work! A bunch of comments inline. 

In general: the code is basically undocumented, esp. the parts dealing with dimensions, offsets, and indexes could use some documentation, e.g., `HaloBlock::HaloBlock` among others.

dash/include/dash/halo/Halo.h
dhinf: it's an enum class -> you can call it with dash::Cycle::NONE only

dash/include/dash/halo/Halo.h
dhinf: it's i fixed constant -> i have dimension^3 regions(including the center)
e.g. 1D 3 regions: north, center, south
       2D 9 regions: north west, north, north east, west, center, east,south west, south, south east
etc.

dash/include/dash/halo/Halo.h
devreal: Ahh, didn't see that (and wasn't aware of enum classes to be honest ^^)

dash/include/dash/halo/Halo.h
devreal: Mhh, but in 1D 1^3 is 1 ;)

In any case, I would suggest using a `constexpr` variable to give the value `3` a meaningful name.

dhinf: meant 3^dimension. sorry

dhinf: constexpr THREE = 3 :D

dhinf: you are right. i will find a meaningful name

dash/include/dash/halo/Halo.h
dhinf: 120

devreal: Please stick to 80 :)

dhinf: but we don't have IBM punched cards anymore and instead nice wide screens :)

dash/include/dash/halo/Halo.h
```diff
@@ -0,0 +1,1104 @@
...
+namespace dash {
+
+// TODO move to util/FunctionalExpr.h
+template <typename BaseT, typename ExpT>
+constexpr BaseT pow( BaseT base, ExpT exp ) {
+  static_assert( std::is_integral<BaseT>::value, "Base must be an integer." );
+  static_assert( std::is_integral<ExpT>::value && std::is_unsigned<ExpT>::value,
+                 "Exponent must be an unsigned integer." );
+
+  return ( exp == 0 ? 1 : base * pow( base, exp - 1 ) );
```
devreal: `return ( exp == 0 ? 1 : (exp > 0) ? base * pow( base, exp - 1 ) : base / pow(base, exp + 1) );` would get rid of the requirement that `ExpT` be an unsigned int, which it wouldn't be if I called `pow(2, 5)`.

dash/include/dash/halo/Halo.h
```diff
@@ -0,0 +1,1104 @@
...
+   */
+  reference operator*() const { return operator[]( _idx ); }
+
+  /**
+   * Subscript operator, returns global reference to element at given
+   * global index.
+   *
+   * \see DashGlobalIteratorConcept
+   */
+  reference operator[]( index_type idx ) const {
```
devreal: I'm a bit puzzled by the semantics of `operator[]` on an iterator: My understanding is that the iterator points to an element in a container. This implementation ignores the current position completely. If it's not really needed, I'd suggest to remove it.

dash/include/dash/halo/Halo.h
```diff
@@ -0,0 +1,1104 @@
...
+  using halo_reg_spec_t = HaloSpec<NumDimensions>;
+  using HaloRegionSpecT = HaloRegionSpec<NumDimensions>;
+  using region_t        = Region<ElementT, PatternT, NumDimensions>;
+  using regions_t       = std::vector<region_t>;
+  using halo_extent_t   = typename halo_reg_spec_t::extent_t;
+  using HaloExtsMaxT =
+    std::array<std::pair<halo_extent_t, halo_extent_t>, NumDimensions>;
+  using size_type     = typename PatternT::size_type;
+  using CoordsT       = std::array<index_type, NumDimensions>;
+  using RegionCoordsT = RegionCoords<NumDimensions>;
```
devreal: Type naming is inconsistent: `SelfT` vs `size_type` vs `region_t`. I know we're sloppy throughout DASH with this but let's try to be consistent within one class at least  (helps readability) :)

dash/include/dash/halo/Halo.h
```diff
@@ -0,0 +1,1104 @@
...
+
+  const ViewSpecT& view_safe() const { return _view_safe; }
+
+  const ViewSpecT& view_inner() const { return _view_inner; }
+
+  const std::vector<ViewSpecT>& boundary_elements() const {
+    return _boundary_elements;
+  }
+
+  size_type halo_size() const {
```
devreal: it seems that the size can be computed once during setup and cached.

dash/include/dash/halo/Halo.h
```diff
@@ -0,0 +1,1104 @@
...
+    _halobuffer.resize( haloblock.halo_size() );
+    auto* offset = _halobuffer.data();
+
+    for( const auto& region : haloblock.halo_regions() ) {
+      _halo_offsets[region.index()] = offset;
+      offset += region.size();
+    }
+  }
+
+  ElementT* haloPos( region_index_t index ) { return _halo_offsets[index]; }
```
devreal: Also: you're using both underscores and CamelCase in the public interface, I think we should consolidate that to underscores.

devreal (changes requested):
Thanks for the fixes. I have some more comments now that I looked over it again. 

Are you planning to add documentation for this PR?

dash/include/dash/halo/Halo.h
dhinf: base / pow(base, exp + 1) ? -> that's not 1/ base^exp

dash/include/dash/halo/Halo.h
dhinf: We are using the subscript operator in other iterators the same way. The standard library uses this operator for the random access iterator different -> http://en.cppreference.com/w/cpp/concept/RandomAccessIterator

dash/include/dash/halo/Halo.h
dhinf: true

dash/include/dash/halo/Halo.h
devreal: `*(i + n)` seems to be reasonable if you think in pointer arithmetics. If we use the subscript operator in other places the same way those places need fixing as well. That's all but intuitive!

dash/include/dash/halo/Halo.h
devreal: Gnarf, you're right! How about that: 

```
return ( exp == 0 ? 1 : (exp > 0) ? base * pow( base, exp - 1 ) : 1 / (base*pow(base, -(exp + 1))) );
```

dhinf:
```c++
return ( exp == 0 ? 1 : (exp > 0) ? base * pow( base, exp - 1 ) : pow(base, (exp + 1)) / base );
```
also works
but the result is the same: 0 -> return type is an integer

devreal: Correct, so you can actually short cut it. This should be caught when moving it to `dash::ce`

dash/include/dash/halo/Halo.h
```diff
@@ -281,137 +277,141 @@ std::ostream& operator<<( std::ostream&                        os,
...
-  constexpr HaloSpec( const SpecsT& specs ) : _specs( specs ) {}
+private:
+  using Self_t          = HaloSpec<NumDimensions>;
+  using RegionCoords_t  = RegionCoords<NumDimensions>;
+  using RegionSpec_t    = RegionSpec<NumDimensions>;
+  using Specs_t         = std::array<RegionSpec_t, RegionCoords_t::MaxIndex>;
+  using Stencil_t       = Stencil<NumDimensions>;
+  using region_index_t  = typename RegionCoords_t::region_index_t;
+  using region_size_t   = typename RegionCoords_t::region_index_t;
+  using region_extent_t = typename RegionSpec_t::region_extent_t;
```
devreal: The types are declared private but exposed to the user through member functions. In that case, they should be public (I know this is messy as well all around DASH). Please also check other places for this.

dash/include/dash/halo/Halo.h
```diff
@@ -281,137 +277,141 @@ std::ostream& operator<<( std::ostream&                        os,
...
-  using SelfT = RegionIter<ElementT, PatternT, PointerT, ReferenceT>;
-  using GlobMemT =
+class RegionIter {
+private:
+  using Self_t = RegionIter<ElementT, PatternT, PointerT, ReferenceT>;
+  using GlobMem_t =
     GlobStaticMem<ElementT, dash::allocator::SymmetricAllocator<ElementT>>;
-  using ViewSpecT = typename PatternT::viewspec_type;
+  using ViewSpec_t      = typename PatternT::viewspec_type;
+  using pattern_index_t = typename PatternT::index_type;
```
devreal: why `pattern_index_t` and not just `index_t`? Is there another (conflicting) index type?

dash/include/dash/halo/Halo.h
```diff
@@ -898,205 +901,203 @@ class HaloBlock {
...
-  const regions_t& boundary_regions() const { return _boundary_regions; }
+  const RegionVector_t& boundary_regions() const { return _boundary_regions; }
 
   // const halo_regs_t& halo_regions() const { return _halo_regions; }
 
-  const ViewSpecT& view() const { return _view; }
+  const ViewSpec_t& view() const { return _view; }
 
-  const ViewSpecT& view_safe() const { return _view_safe; }
+  const ViewSpec_t& view_safe() const { return _view_safe; }
```
devreal: Just out of curiosity: you mentioned that you are looking for a different name for `safe`, any luck so far?

devreal (changes requested):
Looks good, just one more thing regarding the types :)

dash/include/dash/halo/Halo.h
dhinf: The thing is, that if i use index_t, i also have to use size_t. Then it might be not clear whether it is std::size_t or an other. it's an ugly name clash. i also use region_size_t.

devreal: Gnarf, how about `size_type` and `index_type`?

dhinf: Personally i hate type as ending. Typically _t is used to indicate a type. Especially longer names don't profit from type.

dash/include/dash/halo/Halo.h
dhinf: view_foolproof? :)

dhinf: but serious: how view_reliable sound?

devreal: Right now, I can't think of the exact meaning of the safe view anymore (missing documentation, y'a know ;)). I vaguely remember that this was the view on the halo cells (?) Why not call it that?

dhinf: is inner + boundary to guarantee neighbours -> thats why it's called view_guaranteed now

devreal (approved):

PR #453, opened 2017-10-24
Create TeamSpec in PatternArguments

devreal:
The variadic pattern arguments parser should create a team-spec that matches the distspec provided by the user, otherwise the pattern is flawed.


devreal:
OK, I was too fast on this:

```
dash::Matrix<int, 2> mat(size*NX, size*NY, dash::BLOCKED, dash::BLOCKCYCLIC(5));
```
now results in
```
$ aprun -n 4 ./pattern_variants
[    1 ERROR ] [ 45244 ] TeamSpec.h               :170  | dash::exception::InvalidArgument             | [ Unit 1 ] TeamSpec(DistributionSpec, Team) only allows one distributed dimension 
```

@fuchsto How should a teamspec be created to make this work?

ddiefenthaler (commented):
Looks good,

the error does not arrive from this PR, for at least not directly.

You can test this with this call:
`dash::Matrix<int, 2> mat(size*NX, size*NY, dash::BLOCKED, dash::BLOCKCYCLIC(5), dash::Team::All());`
The same error arises.
But it requires your change with the `std::forward`.

I also noticed that `TeamSpec` would need some rework a while ago. But as @fuchsto mentioned a dependency to the topology of the system, I wasn't sure how to accomplish this. I only reworked the function `balance_extents`.

devreal:
Update: The parser now creates a TeamSpec in the following cases:

1) The DistributionSpec contains only distributions that are not `NONE`; 
2) The DistributionSpec contains only one distribution is not `NONE`.

For 1), we create a balanced teamspec. For 2), a teamspec is created based on the distspec.
This covers all two-dimensional patterns. For three-dimensional patterns, the parser errors out on things like `BLOCKED, BLOCKED, NONE` because I have no clue what a correct TeamSpec looks like. @fuchsto Can we automate this?

dash/test/pattern/VarArgsPatternTest.cc
```diff
@@ -0,0 +1,48 @@
...
+{
+  size_t size = dash::size();
+  //  Matrix uses TilePattern by default
+  {
+    dash::Matrix<int, 2> mat(size*NX, size*NY);
+    ASSERT_GT_U(mat.local_size(), 0);
+  }
+  {
+    dash::Matrix<int, 2> mat(size*NX, size*NY,
+                             dash::BLOCKED, dash::BLOCKCYCLIC(5));
```
ddiefenthaler: Is the use of `5` instead of `NY` here on purpose?
I see the test break if NY is too low.

dash/test/pattern/VarArgsPatternTest.cc
```diff
@@ -0,0 +1,48 @@
...
+    ASSERT_GT_U(mat.local_size(), 0);
+  }
+  {
+    dash::Matrix<int, 2> mat(size*NX, size*NY,
+                             dash::BLOCKCYCLIC(3), dash::BLOCKCYCLIC(5));
```
ddiefenthaler: Here too.
Some goes for `3` and `NX`.

dash/test/pattern/VarArgsPatternTest.cc
```diff
@@ -0,0 +1,48 @@
...
+    ASSERT_GT_U(mat.local_size(), 0);
+  }
+
+  //  NArray uses BlockPattern by default
+  {
+    dash::NArray<int, 2> mat(size*NX, size*NY,
+                             dash::BLOCKCYCLIC(3), dash::BLOCKCYCLIC(5));
```
ddiefenthaler: Here too.

dash/test/pattern/VarArgsPatternTest.cc
```diff
@@ -0,0 +1,48 @@
...
+    ASSERT_GT_U(mat.local_size(), 0);
+  }
+
+  //  shift tile pattern
+  {
+    dash::Matrix<int, 2, dash::default_index_t,
+                 dash::ShiftTilePattern<2> > mat(size*NX, size*NY,
+                                                 dash::TILE(3), dash::TILE(5));
```
ddiefenthaler: Here too.

ddiefenthaler (changes requested):

ddiefenthaler (approved):
Thanks!

Looks good for me.
But the word of @fuchsto is here probably worth more than mine.

PR #456, opened 2017-11-03
fix googletest include to work with ninja build system

fmoessbauer:
With this fix, DASH can be build using the ninja build system.
Just add `-G Ninja` to the `cmake` call and then call `ninja` in the build-directory instead of `make`.

For scratch builds there is not a significant differenc in build time compared to make, however incremental builds are way faster.


devreal:
Sounds interesting, just tested it and it builds fine (once I installed the correct package, `ninja-build` instead of `ninja`). However, I get the following warning:

```
CMake Warning (dev):
  Policy CMP0058 is not set: Ninja requires custom command byproducts to be
  explicit.  Run "cmake --help-policy CMP0058" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.

  This project specifies custom command DEPENDS on files in the build tree
  that are not specified as the OUTPUT or BYPRODUCTS of any
  add_custom_command or add_custom_target:

   dash/dash-mpi_CXX_cotire.cmake

  For compatibility with versions of CMake that did not have the BYPRODUCTS
  option, CMake is generating phony rules for such files to convince 'ninja'
  to build.

  Project authors should add the missing BYPRODUCTS or OUTPUT options to the
  custom commands that produce these files.
```

Can we fix the cotire target to suppress this warning?

fmoessbauer:
Yep, an additional policy had to be set, for details have a look here: https://github.com/sakra/cotire/issues/81

devreal (approved):
Thanks a lot, works for me :)

ddiefenthaler (changes requested):
I just wanted to leave the same review on the `cotire`...

But for me it does not seem to be resolved as I still get a linker error on `dash-test-mpi` if `ENABLE_COMPTIME_RED` is set to `ON`:
```
FillTest.cc:(.text.startup+0x58): undefined reference to `testing::internal::MakeAndRegisterTestInfo(char const*, char const*, char const*, char const*, void const*, void (*)(), void (*)(), testing::internal::TestFactoryBase*)'
dash/CMakeFiles/dash-test-mpi.dir/test/algorithm/GenerateTest.cc.o: In function `_GLOBAL__sub_I__ZN30GenerateTest_TestGenerate_Test10test_info_E':
GenerateTest.cc:(.text.startup+0x58): undefined reference to `testing::internal::MakeAndRegisterTestInfo(char const*, char const*, char const*, char const*, void const*, void (*)(), void (*)(), testing::internal::TestFactoryBase*)'
GenerateTest.cc:(.text.startup+0x93): undefined reference to `testing::internal::MakeAndRegisterTestInfo(char const*, char const*, char const*, char const*, void const*, void (*)(), void (*)(), testing::internal::TestFactoryBase*)'
dash/CMakeFiles/dash-test-mpi.dir/test/bits/UniversalMemberTest.cc.o: In function `_GLOBAL__sub_I__ZN36UniversalMemberTest_TestHelpers_Test10test_info_E':
UniversalMemberTest.cc:(.text.startup+0x58): undefined reference to `testing::internal::MakeAndRegisterTestInfo(char const*, char const*, char const*, char const*, void const*, void (*)(), void (*)(), testing::internal::TestFactoryBase*)'
UniversalMemberTest.cc:(.text.startup+0x93): undefined reference to `testing::internal::MakeAndRegisterTestInfo(char const*, char const*, char const*, char const*, void const*, void (*)(), void (*)(), testing::internal::TestFactoryBase*)'
dash/CMakeFiles/dash-test-mpi.dir/test/bits/UniversalMemberTest.cc.o:UniversalMemberTest.cc:(.text.startup+0xce): more undefined references to `testing::internal::MakeAndRegisterTestInfo(char const*, char const*, char const*, char const*, void const*, void (*)(), void (*)(), testing::internal::TestFactoryBase*)' follow
```

Also `BUILD_BYPRODUCTS` seems to only exist since cmake version 3.2.
Are there further steps required for it to work in older versions? The warning to the policy also stated something on the topic of compatibility to older versions of cmake.

fmoessbauer:
@ddiefenthaler : Thanks for your review. I added a guard, so that the `BUILD_BYPRODUCTS` is only set in recent CMake versions. If someone uses an older version, he cannot build with ninja. However that is not a show-stopper as this did not work before this PR either.

I tested the current version on SuperMUC and at my local machine and cannot reproduce your linker error.

CMakeLists.txt
```diff
@@ -23,6 +23,9 @@ endif(POLICY CMP0004)
 if(POLICY CMP0016)
   cmake_policy(SET CMP0016 OLD)
 endif(POLICY CMP0016)
+if(POLICY CMP0058) 
+  cmake_policy(SET CMP0058 NEW) 
+endif()
```
ddiefenthaler, line 28: I don't really know if we want this policy set.
As all it does is to suppress the warning of missing dependencies.
Maybe leave it unset in an developer/debug-build?

On the note of that, I tried to set it to `old` and ended up in an endless loop of cmake reconfigure.
This may also happen with an older version of `cmake` instead of an error.

ddiefenthaler (commented):
I finally figured out why the first build on my system was broken.
It was caused by  `gtest`-header files in the system includes.
As my system didn't include the library to the headers
or as the cmake pulled a different version of `gtest` the linker failed.

This happens as `cotire` is run before `gtest` is pulled.
On the second run the correct version of `gtest` is used and the linker works.
After updating the version in the system includes now the first build works too.

Nevertheless this is a missing dependency as it can break the build as seen.
I don't know if there is an easy fix for this. See also the comment inline on the policy.
If there is no way of setting dependencies to `cotire` this is probably more a issue of ´cotire´.

CMakeLists.txt
fmoessbauer: This policy has no effect in most cases, we still get warnings if dependencies could not be resolved: See here: https://cmake.org/cmake/help/v3.3/policy/CMP0058.html

> This policy has no effect on generators other than Ninja.

So, yes we want to set this policy as the users should only see actual warnings and not fake ones.

The gtest issue is not really related to ninja, but to cotire and a messy system setup. If you like, open an issue at the [cotire repo](https://github.com/sakra/cotire/issues). **Update:** I guess there is already an open issue regarding this problem: https://github.com/sakra/cotire/issues/133

IMO we can merge this PR as it does not even affect the unix make build-system. The ninja-build also should work for most people and if not they can switch back to make. If it makes you sleep better, add a message that ninja builds are experimental.

ddiefenthaler: I don't know if it is really a fake warning, as it is not displayed if `cotire` isn't included.

ddiefenthaler (approved):
@fmoessbauer Thanks for your opinion.
I agree that this PR should be ready to merge.

PR #457, opened 2017-11-03
DASH Coarray

fmoessbauer:
This PR adds CAF 2008 features to DASH. Here are the relevant parts of the interface, copied from the documentation. Reviews are welcome!

**TODO:** There are still some progress problems in Coevent and Comutex which only occur on CI. This is almost certainly due to pinning multiple units to the same CPU core. I'm working on a fix.

# Coarray

The interface of the Coarray is similar to the CAF Coarray. The round bracket operator selects the unit, the square bracket operator is for the offset (within the unit). The all local extents (except the last) of the Coarray have to be specified at compile time, similar to native C arrays.

If no unit is selected using the round-bracket-operator, the access is performed on the local part.

```cpp
dash::Coarray<int> i;          // scalar Coarray
dash::Coarray<int[10][20]> x;  // 2D-Coarray
dash::Coarray<int[][20]> y(n); // one open dim,
                               // set at runtime in ctor
// access syntax
i(unit) = value; // global access
i       = value; // local access
x(unit)[idx1][idx2] = value; // global access
x[idx1][idx2]       = value; // local access
```
Scalar Coarrays can be used like a variable in arithmetic expressions:

```cpp
dash::Coarray<int> x;
x += 1;
x = x(1) + 1;
x = 1 + x(1);
```

# Coevent

Coevent can be used for point-to-point synchronization. Events can be posted to any image. Waiting on non-local events is not supported. The interface of the `dash::Coevent` is similar to a scalar Coarray.

```cpp
Coevent events;
events(2).post();
if(this_image() == 2){
 events.wait();
}
```

# Comutex

Comutex is used to ensure mutual exclusion on a certain image. The interface is similar to `dash::Coarray` but does not allow local accesses. Hence it does not fulfill the DASH Container Concept.

```cpp
Coarray<int> arr;
Comutex      comx;
{
 // lock unit i 
 std::lock_guard<dash::Mutex> lg(comx(i));
 // exclusively access data on unti i
 arr(i) = 42;
}
```

# Additional Features

Functions of the `dash::Coarray` extension to provide an interface similar to CAF 2008. 

## Synchronisation

- `dash::coarray::this_image()` Shortcut to query the global unit ID of the calling unit.
- `dash::coarray::num_images ()` Return the number of units in the global team.

- `dash::coarray::sync_all ()` blocks until all units reach this statement.
- `dash::coarray::sync_images (...)` 

## Collective Operations

- `dash::coarray::cobroadcast(...)` Broadcasts the value on master to all other members of this coarray.
- `dash::coarray::coreduce(...)` Performs a broadside reduction of the Coarray images


dash/include/dash/Coarray.h
```diff
@@ -0,0 +1,771 @@
...
+
+namespace detail {
+
+template <
+  class ArrayT,
+  class ValueT,
+  std::size_t Rank,
+  std::size_t... Is >
+constexpr std::array<ValueT, Rank>
+__get_type_extents_as_array_impl(
```
devreal: I suggest to remove the `__get_` prefix from the type traits, it's hidden in `namespace detail` anyway

dash/include/dash/Coarray.h
```diff
@@ -0,0 +1,771 @@
...
+ * each unit is equal (symmetric allocation).
+ */
+template<
+  typename T,
+  typename IndexType = dash::default_index_t>
+struct make_coarray_symmetric_pattern {
+  using type = BlockPattern<
+                std::rank<
+                  typename dash::remove_atomic<T>::type>::value+1,
+                ROW_MAJOR,
```
devreal: Shouldn't we leave the choice of `ROW_MAJOR` vs `COL_MAJOR` to the user?

dash/include/dash/Coarray.h
```diff
@@ -0,0 +1,771 @@
...
+ * 
+ * x(unit)[idx1][idx2] = value; // global access
+ * x[idx1][idx2]       = value; // local access
+ * \endcode
+ *
+ */
+template<
+  typename T,
+  typename IndexType = dash::default_index_t,
+  class PatternType  = typename coarray::make_coarray_symmetric_pattern<T,IndexType>::type >
```
devreal: Does it make sense to leave the choice of pattern to the user? If they want to be flexible they can use `dash::NArray` directly :)

dash/include/dash/Coarray.h
```diff
@@ -0,0 +1,771 @@
...
+private:
+
+  using self_t          = Coarray<T, IndexType, PatternType>;
+  
+  /**
+   * _element_type has no extent and is wrapped with \cdash::Atomic, if coarray
+   * was declared with \cdash::Atomic<T>
+   */
+  using _element_type   = typename std::remove_all_extents<T>::type;
+  using _underl_type    = typename dash::remove_atomic<_element_type>::type;
```
devreal: `_underl_type` --> `_base_type`?

dash/include/dash/Coarray.h
```diff
@@ -0,0 +1,771 @@
...
+__get_type_extents_as_array() {
+  return __get_type_extents_as_array_impl<ArrayT,ValueT,Rank>(
+      dash::ce::make_index_sequence<Rank>());
+}
+
+template<
+  typename element_type,
+  typename pattern_type,
+  int rank> 
+struct __get_local_type {
```
devreal: `local_type` --> `localref_type` or `local_ref_type`

dash/include/dash/Coarray.h
```diff
@@ -0,0 +1,771 @@
...
+                ROW_MAJOR,
+                IndexType>;
+};
+
+} // namespace coarray
+
+/**
+ * A fortran style coarray.
+ * 
+ * \ingroup DashCoarrayConcept
```
devreal: Not sure but maybe the `\ingroup` statement has to come at the end, documentation is cut off here when building with doxygen.

dash/include/dash/Coarray.h
```diff
@@ -0,0 +1,771 @@
...
+              std::array<size_type, 1> {static_cast<size_type>(dash::size())},
+              coarray::detail::__get_type_extents_as_array<T,
+                size_type, _rank::value>()));
+  }
+
+  constexpr _sspec_type _make_size_spec(const size_type & first_dim) const noexcept {
+    static_assert(
+        std::get<0>(coarray::detail::__get_type_extents_as_array<T,
+          size_type, _rank::value>()) == 0,
+      "Array type is fully specified");
```
devreal: More descriptive error message: `Array type may not be fully specified`

dash/include/dash/Coarray.h
```diff
@@ -0,0 +1,771 @@
...
+  using self_t          = Coarray<T, IndexType, PatternType>;
+  
+  /**
+   * _element_type has no extent and is wrapped with \cdash::Atomic, if coarray
+   * was declared with \cdash::Atomic<T>
+   */
+  using _element_type   = typename std::remove_all_extents<T>::type;
+  using _underl_type    = typename dash::remove_atomic<_element_type>::type;
+  
+  using _rank           = std::integral_constant<int,
```
devreal: `_rank` is the rank of the value type (`_valuetype_rank`). I suggest adding a private member `_rank = _valuetype_rank+1` as this occurs multiple times throughout the code

dash/include/dash/Coarray.h
```diff
@@ -0,0 +1,771 @@
...
+  constexpr size_type size() const noexcept {
+    return _storage.size();
+  }
+
+  /**
+   * For fully specified array types, \c dash::Coarray<T[N...]> is a fixed-size
+   * container. Hence the value equals the value returned by \c max_size .
+   * For containers with one open dimension, the size is determined by dash.
+   * If dash is not initialized, it always returns zero.
+   * \TODO Currently _storage.size() is returned.
```
devreal: Needs fixing?

dash/include/dash/Coarray.h
```diff
@@ -0,0 +1,771 @@
...
+  
+  constexpr bool is_local(index_type gi) const noexcept {
+    return _storage.is_local(gi);
+  }
+  
+  /**
+   * allocate an array which was initialized before dash has been initialized
+   */
+  inline void allocate(){
+    _storage.allocate(_pattern_type(_make_size_spec()));
```
devreal: I cannot pass a team to `allocate`?

devreal: Also, IIRC `allocate` should check whether the CoArray has been allocated already.

dash/include/dash/Coarray.h
```diff
@@ -0,0 +1,771 @@
...
+  template<int __rank = _rank::value>
+  inline view_type<__rank> operator()(const team_unit_t & unit) {
+    return this->operator ()(static_cast<index_type>(unit));
+  }
+  /**
+   * Operator to select remote unit for array types
+   */
+  template<int __rank = _rank::value>
+  typename std::enable_if<(__rank != 0), view_type<__rank>>::type
+  inline operator()(const index_type & local_unit) {
```
devreal: Why is the parameter called `local_unit`? Seems misleading as it selects the unit to access, which is a remote unit.

dash/include/dash/Coarray.h
```diff
@@ -0,0 +1,771 @@
...
+   * Provides access to local array part
+   * \code
+   *   dash::Coarray<int[10][20]> x;
+   *   x[2][3] = 42;
+   * \endcode
+   */
+  template<int __rank = _rank::value>
+  typename std::enable_if<(__rank > 1), local_type>::type
+  inline operator[](const index_type & idx) {
+    return _storage.local[0][idx];
```
devreal: Why the `[0]`? `operator[]` on a CoArray should only dereference the first dimension and leave the remaining dimensions to `LocalMatrixRef`, right?

dash/include/dash/Coarray.h
```diff
@@ -0,0 +1,771 @@
...
+  
+  /**
+   * convert scalar Coarray to a global reference.
+   */
+  template<
+    int __rank = _rank::value,
+    typename = typename std::enable_if<(__rank == 0)>::type>
+  explicit inline operator reference() {
+    return static_cast<reference>(
+            *(_storage.begin()+static_cast<index_type>(dash::myid())));
```
devreal: `return _storage[dash::myid()]` seems more efficient

dash/include/dash/Coarray.h
```diff
@@ -0,0 +1,771 @@
...
+   * \code
+   *   dash::Coarray<int> i;
+   *   i = 42;
+   *   i += 21;
+   * \endcode
+   */
+  template<
+    int __rank = _rank::value,
+    typename = typename std::enable_if<(__rank == 0)>::type>
+  inline value_type & operator +=(const value_type & value) {
```
devreal: Operator `@=` should return reference_type

dash/include/dash/Coarray.h
```diff
@@ -0,0 +1,771 @@
...
+    return *(_storage.lbegin()) / value;
+  }
+  
+  /**
+   * allows fortran like local access of scalars
+   */
+  template<
+    int __rank = _rank::value,
+    typename = typename std::enable_if<(__rank == 0)>::type>
+  inline value_type operator ++() {
```
devreal: Return `reference_type` here as well

dash/include/dash/Coevent.h
```diff
@@ -0,0 +1,153 @@
...
+ * \ingroup DashCoarrayConcept
+ * 
+ *
+ * A fortran style coevent.
+ *
+ * Coevent can be used for point-to-point synchronization. Events can be posted
+ * to any image. Waiting on non-local events is not supported.
+ * 
+ * \note Coevents might deadlock if multiple units are pinned to the same
+ *       cpu-core. This is due to progress problems in MPI.
```
devreal: I'm still not sure why this should be case, CFS should ensure that eventually all units get to do the update even if they are pinned to the same core.

dash/include/dash/Coevent.h
```diff
@@ -0,0 +1,153 @@
...
+
+  /**
+   * initializes the Coevent. If it was already initialized in the Ctor,
+   * the second initialization is skipped.
+   */
+  inline void initialize(Team & team = dash::Team::All()) noexcept {
+    _team = &team;
+    _event_counts.allocate(_team->size());
+    dash::fill(_event_counts.begin(), _event_counts.end(), 0);
+    _is_initialized = true;
```
devreal: Test whether the CoEvent has been initialized before!

devreal: Also, add a barrier at the end as `dash::fill` currently does not block

dash/include/dash/Comutex.h
```diff
@@ -0,0 +1,133 @@
...
+namespace dash {
+
+/**
+ * \ingroup DashCoarrayConcept
+ * 
+ *
+ * A fortran style comutex.
+ *
+ * Comutex is used to ensure mutual exclusion on a certain image.
+ * The interface is similar to \cdash::Coarray but does not allow
```
devreal: `\c` requires a trailing space, please fix throughout the Co* documentation

dash/include/dash/atomic/GlobAtomicRef.h
```diff
@@ -181,7 +181,10 @@ class GlobRef<dash::Atomic<T>>
                        reinterpret_cast<void * const>(&result),
                        dash::dart_punned_datatype<T>::value,
                        DART_OP_NO_OP);
-    dart_flush_local(_gptr);
+    // a flush_local is not sufficient as some versions of
+    // openmpi and MPICH / IntelMPI fetch the old value
+    // due to progress problems
```
devreal: Wait what? That sounds like a bug. Do you have a reproducer/example for this behavior?

dash/include/dash/coarray/Utils.h
```diff
@@ -0,0 +1,212 @@
...
+               image_ids.end(),
+               static_cast<element>(myid)) == image_ids.end())
+  {
+    // I do not participate
+    return;
+  }
+
+  auto root        = global_unit_t{
+                        *(std::min(image_ids.begin(), image_ids.end()))};
+  // 10000 + -MCA_COLL_BASE_TAG_BARRIER of openmpi
```
devreal: Remove reference to Open MPI

dash/include/dash/coarray/Utils.h
```diff
@@ -0,0 +1,212 @@
...
+ * \note If possible use \c sync_all() or \c Coevent for performance reasons.
+ *       \c sync_images() is implemented using two-sided operations based on 
+ *       the implementation of \c MPI_Barrier() in OpenMPI
+ * 
+ * \sa dash::coarray::sync_all()
+ * 
+ * \ingroup DashCoarrayLib
+ */
+template<typename Container>
+inline void sync_images(const Container & image_ids){
```
devreal: It might be more efficient to implement this functionality in DART (using MPI non-blocking communication). I will keep that in mind.

dash/include/dash/coarray/Utils.h
```diff
@@ -0,0 +1,212 @@
...
+  const auto & global_idx    = coarr.pattern().at(global_coords);
+
+  const auto dart_gptr       = (coarr.begin() + global_idx).dart_gptr();
+  const dash::dart_storage<value_type> ds(coarr.local_size());
+
+  // as source and destination location is equal, master must not contribute
+  // in accumulation as otherwise it's value is counted twice
+  if(coarr.team().myid() != master){
+    DASH_ASSERT_RETURNS(
+      dart_accumulate(
```
devreal, line 190: 1) Note that  `dart_accumulate` performs an element-wise atomic update so it is unsafe for anything other than basic types. I suggest limiting `coreduce` to basic types for now.
2) Also, op is limited to predefined operations, which should at least be documented.
3) It should be explicitly documented that the result will be the first value on the receiving unit(s).

dash/test/iterator/GlobRefTest.cc
```diff
@@ -0,0 +1,52 @@
...
+  auto global_offset      = array.pattern().global(0);
+  // Reference first local element in global memory:
+  dash::GlobRef<int> gref = array[global_offset];
+  ASSERT_EQ_U(true, gref.is_local());
+}
+
+typedef struct value_struct {
+  double x;
+  int    y;
+} value_struct_t;
```
devreal: duplicated definition of `value_struct_t` (see above)

devreal (changes requested):
A lot of good work in here :) I have a bunch of comments inline, most of which are minor.

devreal:
I also noticed that there are many empty lines with leading spaces, should be easy to fix in your editor :)

dash/include/dash/Coarray.h
```diff
@@ -0,0 +1,771 @@
...
+  using _underl_type    = typename dash::remove_atomic<_element_type>::type;
+  
+  using _rank           = std::integral_constant<int,
+                            std::rank<T>::value>;
+  
+  using _size_type      = typename std::make_unsigned<IndexType>::type;
+  using _sspec_type     = SizeSpec<_rank::value+1, _size_type>;
+  using _pattern_type   = PatternType;
+  
+  using _storage_type   = Matrix<_element_type, _rank::value+1, IndexType, _pattern_type>;
```
devreal: Just a thought I had: could we use `dash::Array` as underlying storage type for scalar CoArrays?

dash/include/dash/Coarray.h
```diff
@@ -0,0 +1,771 @@
...
+   * \code
+   *   dash::Coarray<int> i;
+   *   i = 42;
+   *   int b = i + 21;
+   * \endcode
+   */
+  template<
+    int __rank = _rank::value,
+    typename = typename std::enable_if<(__rank == 0)>::type>
+  inline value_type operator +(const value_type & value) const {
```
devreal: Isn't that covered through the cast `operator value_type()` already?

dash/include/dash/Coarray.h
fmoessbauer: No, this operator operates only on the local part of the array and hence a native reference is returned.

dash/include/dash/Coarray.h
fmoessbauer: Same as above.

dash/include/dash/Coevent.h
fmoessbauer: This is a heisenbug :( It only appears on CI and I cannot reproduce it locally (even if pinning all units to one core). However what I observed was that the CI only failed hung if the pairing was not unique.

dash/include/dash/Coarray.h
fmoessbauer: Yes, and `dash::myid` has to be replaced by `_storage.team().myid()`

dash/include/dash/Comutex.h
fmoessbauer: Done, I hope I fixed all occurances

dash/include/dash/coarray/Utils.h
fmoessbauer: Well ... It might be and DART is the correct place for such things, however @fuerlinger voted for removing this function as it totally contradicts the team concept. Currently it is only there to make porting of CAF codes easier.

dash/include/dash/coarray/Utils.h
fmoessbauer: Is there already a tag range for DART? I would prefer that as it will clash some time if everyone uses his own tags.

dash/include/dash/coarray/Utils.h
fmoessbauer: 1 and 2 are fixed now. What do you mean with 3? The function performs a broadside reduction:

```
Unit 0 | 1 3 0 2
Unit 1 | 0 2 1 1
----------------
All    | 1 5 1 3 (using coreduce(x,+))
```

dash/include/dash/Coarray.h
fmoessbauer: Theoretically possible, but currently not fully implemented as some types mismatch. For the paper I tested it, so see it as future work ;)

dash/include/dash/Coarray.h
fmoessbauer: Unfortunately not, as C++ does not convert types over two edges. This is required for use cases like that:
```cpp
dash::Coarray<int> x;
// does not work without these operators
int a = x + 5;
// works without
int a = 5 + x;
```

dash/include/dash/atomic/GlobAtomicRef.h
fmoessbauer: This is a bug and if I remember correctly you already reported it to OpenMPI. In fact this is your code and I took it from the reproducer. Unfortunately I did not document the issue.

fmoessbauer:
Hi @devreal thanks for your very detailed review. The new commits address most of your points. The progress problems in `Coevent` probably have to be investigated further.

dash/include/dash/Coarray.h
devreal: Ahh, I got confused by *value*`_type`, which is actually a reference ;)

dash/include/dash/atomic/GlobAtomicRef.h
devreal: Ahh I remember. I reported this and it should be fixed since Open MPI 2.1.1 I think. But that was not related correctness, `MPI_Win_flush_local` didn't trigger the progress engine back then.

dash/include/dash/coarray/Utils.h
devreal: DART doesn't use tags itself (as of now). Feel free to chose your own range for Coarray.

dash/include/dash/coarray/Utils.h
devreal: You're right, I wrote 3) together with a bunch of other comments that I later deleted when I realized that it's an element-wise reduction, that one should have been gone as well.

dash/include/dash/Coarray.h
devreal: Crap, that was bullshit. It's correct the way it is for `operator @=`.

dash/include/dash/Coarray.h
devreal: Actually, `operator ++` should return `value_type &` then, otherwise something like `++(++coarr[i])` won't work correctly.

Also, should these operators be disabled/altered for atomics?

dash/include/dash/coarray/Utils.h
devreal: If this is a CAF/Coarray concept maybe we should leave it in (it doesn't sound like it would scale to the extremes though). I guess you already had sufficient discussions on this :)

dash/include/dash/Coarray.h
devreal: Thanks for adding the comment. I'm still not sure why you have to dereference two dimensions locally?

dash/include/dash/atomic/GlobAtomicRef.h
fmoessbauer: Ok, I reverted this back to the version in development and it seems to work now. Regarding correctness: This only happend if spinning on a local (atomic) value.

fmoessbauer:
@devreal I fixed the atomic operator implementation, however this is dependent on the changes in #458.

devreal (approved):

dash/include/dash/Coevent.h
```diff
@@ -0,0 +1,156 @@
+#ifndef COEVENT_H_INCLUDED
```
bertwesarg, line 1: looks like that should have been named 'DASH__COEVENT_H__INCLUDED'

fmoessbauer: Yep, thanks! Will fix it.

PR #458, opened 2017-11-06
Add support for async atomics

devreal:
While reviewing PR #457 today I once more realized that we are missing support for asynchronous atomics. This PR adds an overload for `GlobAsyncRef<Atomic<T>>`, which provides suitable functionality. 

The PR also removes much of the `const` qualifiers from `GlobAtomicRef` (why was `store()` const?!) and fixes postfix/prefix increment/decrement.


dash/include/dash/atomic/GlobAtomicRef.h
```diff
@@ -320,39 +320,65 @@ class GlobRef<dash::Atomic<T>>
...
+    /// Value to be added to global atomic variable.
+    const T & value)
+  {
+    return fetch_op(dash::multiply<T>(), value);
+  }
+
   /// prefix atomically increment value by one
-  T operator++ () const {
-    return fetch_add(1) + 1;
+  self_t& operator++ () {
```
fmoessbauer: This is a highly problematic change, as the operator is not atomic any more. I thought about this for a while as this also affects the coarray correctness. There are use-cases where returning a reference is necessary
 (`++(++a))`) however even the STL does not support this for [std::atomic](http://en.cppreference.com/w/cpp/atomic/atomic/operator_arith).

If you like we can have a discussion on that, but I strongly recommend to revert this to the current semantics.

dash/include/dash/atomic/Operation.h
```diff
@@ -31,15 +31,15 @@ namespace atomic {
  * Get the value of the shared atomic.
  */
 template<typename T>
-T load(const dash::GlobRef<dash::Atomic<T>> & ref){
+T load(const dash::GlobRef<dash::Atomic<T>>& ref){
```
fmoessbauer, line 34: This is a consequence of the removed `const` in `GlobRef`. There is a documentation entry in GlobRef explaining why the `const` is correct there. There is also a proposal (don't have it at my finger tips currently) which explains this in more detail.

fmoessbauer (changes requested):
Please have a look at my comments inline. Unfortunately I do not agree with the changes regarding the operator semantics as well as the removed `const`.

Probably we should discuss this with the team.

dash/include/dash/atomic/GlobAtomicRef.h
devreal: Thanks a lot for the clarification. I restored the original behavior and documented it for future wanderers wondering about it :)

dash/include/dash/atomic/Operation.h
devreal: I couldn't find that documentation in `GlobRef`. As far as I can see, `GlobRef` doesn't have `const` on member functions altering its value. 

I think we lost const-correctness somewhere on the way. The following example compiles fine where IMO it shouldn't:

```  
  const dash::Array<int>& carr = arr;

  if (dash::myid() == 0) {
    carr[0] = 1;
  }
```

I'd expect `operator[]() const` to return a `const GlobRef<T>` instead of a `GlobRef<const T>` because in the latter the `const` has no effect. Unfortunately, I can't find any documentation...

dash/include/dash/atomic/GlobAtomicRef.h
```diff
@@ -347,38 +348,54 @@ class GlobRef<dash::Atomic<T>>
...
+   */
+  T operator+=(const T & value) {
+    return fetch_add(value) + value;
   }
 
-  /// atomically decrement value by ref
+  /**
+   * atomically decrement value by ref
+   */
   self_t& operator-=(const T & value) {
```
fmoessbauer: `T operator-=`

dash/include/dash/atomic/GlobAtomicRef.h
```diff
@@ -320,38 +321,80 @@ class GlobRef<dash::Atomic<T>>
...
   }
 
-  /// prefix atomically decrement value by one
-  T operator-- () const {
-    return fetch_sub(1) - 1;
+  /**
+   * prefix atomically decrement value by one
+   * Note that this operator does not return a reference but a copy of the value
+   * in order to ensure atomicity. This is consistent with the C++ std::atomic
+   * \c operator++, see
```
fmoessbauer: `operator--`

dash/include/dash/atomic/Operation.h
```diff
@@ -31,15 +31,15 @@ namespace atomic {
...
   return ref.load();
 }
 
 /**
- * Set the value of the atomic reference 
+ * Set the value of the atomic reference
  */
 template<typename T>
-void store(const dash::GlobRef<dash::Atomic<T>> & ref,
+void store(dash::GlobRef<dash::Atomic<T>> ref,
```
fmoessbauer: Cant we use nonconst references here? Or do we better rely on compiler optimization?

dash/include/dash/atomic/Operation.h
fmoessbauer: Ok, that convinces me.

dash/include/dash/atomic/Operation.h
devreal: Unfortunately not, because we then cannot pass an rvalue in, e.g.,

```
dash::Array<dash::Atomic<int>> arr(100);
dash::store(arr[1], 1);
```

dash/include/dash/atomic/GlobAtomicRef.h
```diff
@@ -320,38 +321,80 @@ class GlobRef<dash::Atomic<T>>
...
+  T operator+=(const T & value) {
     return fetch_add(value) + value;
   }
 
-  /// atomically decrement value by ref
-  T operator-=(const T & value) const {
+  /**
+   * atomically decrement value by ref
+   */
+  self_t& operator-=(const T & value) {
```
fmoessbauer: Does this compile? Probably a missing unit test. Should be `T operator`

dash/include/dash/atomic/GlobAtomicAsyncRef.h
```diff
@@ -0,0 +1,432 @@
...
+  /**
+   * Checks whether the globally referenced element is in
+   * the calling unit's local memory.
+   */
+  bool is_local() const {
+    return GlobPtr<T>(_gptr).is_local();
+  }
+
+  /// atomically assigns value
+  GlobRef<atomic_t> operator=(const T & value) {
```
fmoessbauer: Self_t

dash/include/dash/atomic/GlobAtomicRef.h
devreal: Good catch, should be fixed now.

fmoessbauer (approved):

dash/include/dash/atomic/Operation.h
fmoessbauer: Sorry for asking again, but why can't we use universal references? This would work:
```cpp
void store(dash::GlobRef<dash::Atomic<T>> && ref,...);
```

devreal: Ahh, totally forgot about that. Fix it :+1:

fuchsto: @fmoessbauer @devreal 

No!
As soon as a type is partially restricted, it is not a forwarding reference any longer. It's a plain overload for a temporary (aka rvalue).

~~~c++
template <typename T> void func(T && param); // <-- forwarding ref
template <typename T> void func(Foo<T> && param); // <-- rvalue
template <typename T> void func(const T && param); // <-- whisky tango foxtrott? ... but still an rvalue ref
~~~

Also, it's "forwarding reference". Meyers used the term "universal reference" in a great article to formulate the type deduction mechanism, but the terminus technicus has be standardized as "forwarding reference".

See:
- https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers
- https://blog.petrzemek.net/2016/09/17/universal-vs-forwarding-references-in-cpp/

dash/include/dash/atomic/Operation.h
fuchsto: @devreal @fmoessbauer 

No, because there is no such thing as a "const reference", there only is "reference to const".
This is why I implemented `GlobRef` with the transformation to `GlobRef<const T>` when returning a global reference from a constant context. Global references are references, not iterators. Because, well, what exactly could be const about a reference? Its state is defined with initialization and immutable until it becomes invalid. The referenced *value* though ...

fuchsto (changes requested):
See my replies to previous comments:
- https://github.com/dash-project/dash/pull/458#discussion_r149868238
- https://github.com/dash-project/dash/pull/458#discussion_r149867931

dash/include/dash/atomic/Operation.h
devreal: To make this a helpful comment: what do you suggest?

dash/include/dash/atomic/Operation.h
devreal: Obviously, the problem is that the const in GlobRef<const T> is simply lost when accessing the data through an external interface.

So what are you suggesting? To disable member functions altering the value referenced by GlobRef<T> for T=const X?

devreal:
I gave const-correctness a try. Basically, we're imitating the compiler's const rules, make it less intuitive to use, and break `const_cast` on `GlobRef`. But it works, sorta.

rkowalewski (approved):
IMHO this is a good interface for async atomics. The unit tests cover the major challenges very well. I also agree with the changes about const correctness while still not all issues are solved. This may be addressed in another separate issue.

fuchsto:
@devreal Could you elaborate why your last changes make it less intuitive to use now, and how the alternatives would work? I'm fond of reference-const-semantics not to imitate a compiler but because it's elementary to the C++ language. If there is more harm than use from that, it's a different story. But I don't see the problem with sticking to `[const] GlobRef<const T>`. But quite a lot when not doing so.

~~~c++
const Type & cr = value;
Type &  mr = cr; // <-- this won't compile and we agree that it should not

const something;
// won't compile if we stick to canonical reference-const-semantics:
GlobRef<const Type> cr = something.give_const_gref();
GlobRef<Type> mr = cr;
// if we would invent "const references":
const GlobRef<Type> cr = something.give_const_gref();
GlobRef<Type> mr = cr; //  <-- would not complain. You lose constness just by assignment.

devreal:
@fuchsto Sorry, I must have missed your question during the break. 

My main point is that we lose information about `const` in the *interface* of `GlobRef` and substitute it by hand-crafted `static_assert` statements. Take the implementation of `GlobRef::set`: 

```
void set(const_value_type* tptr) const {
   static_assert(std::is_same<value_type, nonconst_value_type>::value,
                 "Cannot modify value through GlobAsyncRef<const T>!");
[...]
}
```

The constness of `GlobRef` becomes irrelevant and we go to great lengths to restore the semantics manually.


Also, `const_cast` has no effect on `GlobRef`, which is not what users expect. Your first example:
```
const T& cref = val;
T& ref = const_cast<T&>(cref);
```
We have no way to apply `const_cast` to convert `GlobRef<const T>` to `GlobRef<T>`. Instead, we have to provide explicit conversion operators that require a `static_cast`: 

```
GlobRef<const T> cref = ...;
GlobRef<T> ref = static_cast<GlobRef<T>>(cref);
```

Having said that, the chances that users actually use the `GlobRef` type explicitely might be smaller than I initially thought and I don't have a good argument against your third example. With a bit of distance I can live with the implementation as it stands now, despite the added complexity of manually enforced const-correctness.

devreal:
Unless there is any opposition by EOB *today* I will go ahead and merge. It's been 2 months after the last change...

PR #465, opened 2017-12-13
Re-enable a test for LoadBalancePattern

pascalj:
The test was disabled in 66cc4a4b due "false" results. It returned the
same size for every node despite having different hwinfo (8x
difference). However, the test showed the correct results according to
the model. To weigh the nodes it uses

        cpu_weight * mem_weight

As both use the cpu_freq and core_count, they cancel out (mem_weight
is `mem_bw / (nthreads * cpu_freq)`). Thus manipulating the core count
while keeping the mem_bw will not alter the nodes' weights relative to
each other.

This change in- and decreases the max_shmem_bw per node.

It would be nice if you have some feedback on two issues:

1. Should the model itself be changed? I see that `mem_weight * cpu_weight` is a very simple model for the performance of each node with the drawbacks that caused this test to get disabled in the first place. However, it's also kind of reasonable that doubling the CPU performance while keeping the memory bandwidth constant won't increase the performance that much.
2. The test fails when executed with 3 or 4 units. This is due to rounding in the calculation for the number of elements on unit 1. I see two ways to solve this: make the calculation more accurate (and more complex) in the test or relax the requirements in the test (i.e. `|expected - actual| <= 1`).


fmoessbauer:
@pascalj The CI has failed as the code does not compile. Please fix it, then we can re-enable the test:

```
../dash/test/pattern/LoadBalancePatternTest.cc: In Elementfunktion »virtual void LoadBalancePatternTest_LocalSizes_Test::TestBody()«:
../dash/test/pattern/LoadBalancePatternTest.cc:72:3: Fehler: »EXPECT_EQ_U« wurde in diesem Gültigkeitsbereich nicht definiert
   EXPECT_EQ_U(cpu_cap_ratio,
   ^
../dash/test/pattern/LoadBalancePatternTest.cc:72:3: Fehler: expected »}« at end of input
```

PR #467, opened 2017-12-13
Another Try: dash::sort

rkowalewski:
Since I need dash::sort for a N-body simulation I gave it another try. Based on very good [previous work](https://github.com/rainwoodman/MP-sort) I greatly modified an existing MPI implementation which uses a partitioned histogram sort. There is still room for optimizations as already suggested by TODOs. Additionally, I will add more unit tests to validate with different containers and patterns.

This implementation has been tested for blocked array distributions and works very well. I will address the major use cases over the next days.


dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,582 @@
...
+   */
+  for (std::size_t idx = 1; idx < l_nlt.size(); ++idx) {
+    auto const transposed_unit = idx - 1;
+    DASH_ASSERT_EQ(
+        g_nlt_all.pattern().local_size(
+            static_cast<dash::team_unit_t>(transposed_unit)),
+        l_nlt.size() - 1, "Invalid Array length");
+    auto const offset = transposed_unit * nunits + myid.id;
+
+    g_nlt_all.async[offset] = l_nlt[idx];
```
devreal: Use `g_nlt_all.async[offset].put(&l_nlt[idx])` to avoid local flush for temporaries (applies to several places throughout)

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,582 @@
...
+      g_nlt_nle.size(), (l_nlt.size() - 1) * 2, "Sizes must match");
+
+  /* reduce among all units */
+  std::fill(g_nlt_nle.lbegin(), g_nlt_nle.lend(), 0);
+
+  g_nlt_nle.team().barrier();
+
+  using glob_atomic_ref_t = dash::GlobRef<dash::Atomic<DifferenceType> >;
+
+  // TODO: Implement GlobAsyncRef<Atomic>, so we can asynchronously accumulate
```
devreal: That is stuck in "review", see #458

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,582 @@
...
+  for (std::size_t idx = 1; idx < l_nlt.size(); ++idx) {
+    // accumulate g_nlt
+    auto const g_idx_nlt      = (idx - 1) * 2;
+    auto       ref_clt        = g_nlt_nle[g_idx_nlt];
+    auto       atomic_ref_clt = glob_atomic_ref_t(ref_clt.dart_gptr());
+    atomic_ref_clt.add(l_nlt[idx]);
```
devreal: This conversion through the DART gptr is really ugly (but of course necessary). Maybe we should come up with something more elegant, like `GlobRef<T>::atomic`.

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,582 @@
...
+  // initial local sort
+  std::sort(lbegin, lend);
+  // Temporary local buffer (sorted);
+  std::vector<value_type> const lcopy(lbegin, lend);
+
+  auto const lmin = *lbegin;
+  auto const lmax = *(lend - 1);
+
+  dash::Shared<dash::Atomic<value_type> > g_min(
+      static_cast<dash::team_unit_t>(0), team);
```
devreal: `dash::team_unit_t(0)` should do as well.

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,582 @@
...
+
+  // exit condition
+  return nonstable_it == p_borders.is_stable.cend();
+}
+}  // namespace detail
+
+template <class GlobRandomIt>
+void sort(GlobRandomIt begin, GlobRandomIt end)
+{
+  auto pattern             = begin.pattern();
```
devreal: Add `static_assert` for integral types.

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,582 @@
...
+  unit = static_cast<team_unit_t>(1);
+
+  for (; unit < last; ++unit) {
+    auto const            prev_u = unit - 1;
+    difference_type const val    = (prev_u == myid)
+                                    ? g_send_count.local[prev_u]
+                                    : g_send_count[prev_u * nunits + myid];
+
+    l_target_displs[unit]                 = val + l_target_displs[prev_u];
+    g_target_displs[unit * nunits + myid] = l_target_displs[unit];
```
devreal: Consider using `async` here.

devreal (changes requested):
Thanks for your efforts! Some comments inline. 

Also: can you please add a high-level description of the algorithm? (or at least a reference?) The code documentation is rather sparse and it would help to have a reference.

dash/test/algorithm/SortTest.cc
```diff
@@ -29,6 +29,8 @@ TEST_F(SortTest, ArraySort)
 
   auto const true_sum = dash::accumulate(array.begin(), array.end(), 0);
 
+  int wait = 0;
+  while(wait);
```
devreal: "OK, I'll wait." Said no CI ever...

rkowalewski: Well that is a very important code snippet as you can imagine. ;)

I'm deleting it. And will also address your change requests. :) Thanks for your review.

devreal:
Thanks for the changes! I'm counting 5 instances of `dash::Array`, three of have dimensions `nunits*nunits` and two with `2*nunits`. Maybe you can consolidate that into two NArrays?

rkowalewski:
As soon as all tests are passing we can do some optimization. This is on my list.

Documentation is on the way as well.

rkowalewski:
First benchmarks are not looking bad. Without discussing too much information here: I simply provide raw data and simple plots. I am inspecting the details, especially regarding scalability with a high number of units.

https://docs.google.com/spreadsheets/d/19avLM8vDvIqz-K6-TP6uGoO8K8xUm_7Z7TP8B3wZ5QQ/edit?usp=sharing

Note that for weak scaling we start with 28 units (1 node), each allocating 475 MB of data. Data is a huge amount of 64 bit signed integers.
In the case of weak scaling, we start with 28 units (1 node), each allocating 250 MB of 64 bit signed integers.
We always scale up to 512 nodes. Data is not complete yet, benchmarks are still running. 

Platform is SuperMUC Phase 2 (Haswell Nodes).

dart-impl/mpi/src/dart_communication.c
```diff
@@ -852,8 +852,11 @@ dart_ret_t dart_get_handle(
 
   *handleptr = handle;
 
-  DART_LOG_TRACE("dart_get_handle > handle(%p) dest:%d",
-                 (void*)(handle), handle->dest);
+  if (NULL != handle) {
+    DART_LOG_TRACE("dart_get_handle > handle(%p) dest:%d",
+                   (void*)(handle), handle->dest);
+  }
```
devreal: That has been fixed in `development`, needs merging

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,1177 @@
...
+ */
+template <class GlobRandomIt>
+void sort(GlobRandomIt begin, GlobRandomIt end);
+
+/**
+ * Sorts the elements in the range, defined by \c [begin, end) in ascending
+ * order. The order of equal elements is not guaranteed to be preserved.
+ *
+ * Elements are sorted by using a user-defined hash function. This hash
+ * function is required to be sortable by operator<.
```
devreal: > This hash function is required to be sortable by operator<.

The return value of the hash function must be sortable by operator<.

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,1177 @@
...
+    std::copy(                                                              \
+        begin, begin + max_elems, std::ostream_iterator<value_t>(os, " ")); \
+    if (nelems > MAX_ELEMS_LOGGING) os << "...";                            \
+    DASH_LOG_TRACE(desc, os.str());                                         \
+  } while (0)
+#else
+#define DASH_SORT_LOG_TRACE_RANGE(desc, begin, end) \
+  do {                                              \
+  } while (0)
+#endif
```
devreal: Can we move `DASH_SORT_LOG_TRACE_RANGE` to `Logging.h`?

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,1177 @@
...
+
+#define IND_DIST(nunits) (nunits * 0)
```
devreal: What is the semantic of these macros? Indizes? If so, better use `IDX_DIST`

Also: `((nunits) * 0)`, similar below

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,1177 @@
...
+  };
+
+  auto comp_upper = [&sortable_hash](
+                        const MappedType& b, const ElementType& a) {
+    return b < sortable_hash(const_cast<ElementType&>(a));
+  };
+
+  auto const n_l_elem = std::distance(lbegin, lend);
+
+  if (n_l_elem == 0) return std::make_pair(n_lt, n_le);
```
devreal: Move this to the top of the function to avoid unnecessary construction/destruction of the objects above.

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,1177 @@
...
+
+  auto const p_left = p_borders.left_partition[last_valid_border_idx];
+
+  std::fill(
+      std::next(std::begin(n_lt), p_left + 2), std::end(n_lt), n_l_elem);
+  std::fill(
+      std::next(std::begin(n_le), p_left + 2), std::end(n_le), n_l_elem);
+
+  DASH_LOG_TRACE("psort__local_histogram >");
+  return std::make_pair(n_lt, n_le);
```
devreal: Avoid copying: `return std::make_pair(std::move(n_lt), std::move(n_le));`

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,1177 @@
...
+
+    std::array<SizeType, NLT_NLE_BLOCK> vals{{l_nlt[idx], l_nle[idx]}};
+    auto const                          g_idx_nlt = (idx - 1) * NLT_NLE_BLOCK;
+
+    dash::transform(
+        std::begin(vals),  // A
+        std::end(vals),
+        it_nlt_nle + g_idx_nlt,   // B
+        it_nlt_nle + g_idx_nlt,   // B = op(B,A)
+        dash::plus<SizeType>());  // op
```
devreal: That is probably a scalability problem. Maybe use `dart_accumulate` and `dart_flush` for now?

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,1177 @@
...
+      });
+
+  // copy local portion
+  index_t lidx      = myid * NLT_NLE_BLOCK;
+  auto    cpy_begin = it_g_nlt_nle + lidx;
+  auto    lbegin    = cpy_begin.local();
+  std::copy(lbegin, lbegin + NLT_NLE_BLOCK, l_nlt_nle_data + lidx);
+
+  // wait for all copies
+  fut_result.wait();
```
devreal: Why construct a future here if you just want to call `dart_waitall_local`? ;)

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,1177 @@
...
+      if (l_nlt_nle[nlt_idx] >= acc_partition_count[peer_idx]) {
+        p_borders.upper_bound[border_idx] = partitions[border_idx];
+      }
+      else {
+        p_borders.lower_bound[border_idx] = partitions[border_idx];
+      }
+    }
+  }
+  for (auto& handle : handles) {
+    DASH_ASSERT_RETURNS(DART_OK, dart_handle_free(&handle));
```
devreal: Not necessary, `dart_waitall_local` will free the handles before returning

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,1177 @@
...
+  auto       dist_begin = l_partition_dist.begin() + IND_DIST(nunits);
+  auto       supp_begin = l_partition_dist.begin() + IND_SUPP(nunits);
+
+  auto const n_my_elements =
+      std::accumulate(dist_begin, dist_begin + nunits, 0);
+
+  // Calculate the deficit
+  auto my_deficit = acc_partition_count[myid + 1] - n_my_elements;
+
+  dash::team_unit_t       unit(0);
```
devreal: Properly scope `unit`

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,1177 @@
...
+
+  auto const last_skipped = p_borders.is_skipped.cend();
+  auto       it_skipped =
+      std::find(p_borders.is_skipped.cbegin(), last_skipped, true);
+
+  auto it_valid = valid_partitions.cbegin();
+
+  std::size_t skipped_idx = 0;
+
+  while (std::find(it_skipped, last_skipped, true) != last_skipped) {
```
devreal, line 473: Is skipping the first found `it_skipped` intentional? If so, please document

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,1177 @@
...
+  auto l_target_displs = &(g_partition_data.local[IND_TARGET_DISP(nunits)]);
+
+  if (0 == myid) {
+    // Unit 0 always writes to target offset 0
+    std::fill(l_target_displs, l_target_displs + nunits, 0);
+  }
+
+  std::vector<SizeType> target_displs(nunits, 0);
+
+  team_unit_t       unit(1);
```
devreal: Proper scoping

devreal (commented):
Updated review inline (about halfway through, more tomorrow)

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,1177 @@
...
+  team_unit_t const last(nunits);
+
+  auto const u_blocksize = g_partition_data.lsize();
```
devreal: Is there a particular reason why `g_partition_data` is an `Array` an not an `NArray`?

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,1177 @@
...
+  std::size_t gsize = nunits * NLT_NLE_BLOCK * 2;
+
+  array_t g_nlt_nle(gsize, dash::BLOCKCYCLIC(NLT_NLE_BLOCK), team);
+
+  array_t g_partition_data(nunits * nunits * 3, dash::BLOCKED, team);
+  std::fill(g_partition_data.lbegin(), g_partition_data.lend(), 0);
+  DASH_ASSERT_GT(
+      g_partition_data.size(),
+      nunits * 2 * 2,
+      "array size is not large enough");
```
devreal: Why is this constraint needed? Should it be documented?

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,1177 @@
...
+ *       dash::Array<int> arr(100);
+ *       dash::generate(arr.begin(), arr.end());
+ *       dash::sort(array.begin(),
+ *                  array.end();
+ * \endcode
+ *
+ * \ingroup  DashAlgorithms
+ */
+template <class GlobRandomIt>
+void sort(GlobRandomIt begin, GlobRandomIt end);
```
devreal, line 53: Clarify documentation: this call is only suitable for arithmetic elements while complex types require the use of a hash function returning an arithmetic type (having `operator<` defined on the type is not sufficient)

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,1177 @@
...
+        all_borders.end(),
+        std::back_inserter(valid_partitions),
+        [&is_skipped](size_t idx) { return is_skipped[idx] == false; });
+  }
+
+  if (valid_partitions.size() == 0) {
+    // Edge case: We may have a team spanning at least 2 units, however the
+    // global range is owned by  only 1 unit
+    team.barrier();
+    return;
```
devreal, line 900: Is the local range already sorted here?

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,1177 @@
...
+  g_min.get().op(dash::min<mapped_type>(), lmin);
+  g_max.get().op(dash::max<mapped_type>(), lmax);
+
+  using array_t = dash::Array<std::size_t>;
+
+  std::size_t gsize = nunits * NLT_NLE_BLOCK * 2;
+
+  array_t g_nlt_nle(gsize, dash::BLOCKCYCLIC(NLT_NLE_BLOCK), team);
+
+  array_t g_partition_data(nunits * nunits * 3, dash::BLOCKED, team);
```
devreal, line 845: This `Array` seems to emulate an `NArray`. Any specific reason? It seems that the index calculation using `IND_DIST` etc could all be done through multiple dimensions of the array, e.g., `NArray<3, size_t> arr(nunits, nunits, 3, BLOCKED, NONE, NONE);`

dash/include/dash/algorithm/Sort.h
```diff
@@ -0,0 +1,1177 @@
...
+    }
+    else {
+      g_partition_data.local[IND_SUPP(nunits) + myid] =
+          g_partition_data.local[IND_DIST(nunits) + unit];
+    }
+  }
+
+  g_partition_data.async.flush();
+
+  team.barrier();
```
devreal: `g_partition_data.barrier()` should do `flush+barrier` ;)

dash/include/dash/pattern/CSRPattern.h
```diff
@@ -1001,7 +1001,8 @@ class CSRPattern<1, Arrangement, IndexType>
   constexpr SizeType local_size(
     team_unit_t unit = UNDEFINED_TEAM_UNIT_ID) const noexcept
   {
-    return _local_size;
+    return unit == UNDEFINED_TEAM_UNIT_ID ?
```
devreal: Already fixed in `development` :)

dash/test/algorithm/SortTest.cc
```diff
@@ -0,0 +1,331 @@
...
+        nullptr>
+static void expect_equals(ValueType const e, ValueType const a)
+{
+  if (std::is_same<decltype(e), double>::value) {
+    EXPECT_DOUBLE_EQ(e, a) << "Unit " << dash::myid().id;
+  }
+  else if (std::is_same<decltype(e), float>::value) {
+    EXPECT_FLOAT_EQ(e, a) << "Unit " << dash::myid().id;
+  }
+}
```
devreal, line 280: Interesting, `EXPECT_EQ_U` should dispatch to the correct equality testing macro, see https://github.com/dash-project/dash/blob/development/dash/test/TestBase.h#L123

devreal (commented):
Some more points, mostly minor :)

dash/include/dash/algorithm/Sort.h
rkowalewski: The reason is simple here: Because we do not support asynchronous operations on the `NArray`. There is no (not yet) async proxy object.

dash/include/dash/algorithm/Sort.h
rkowalewski: Good catch. ;)

dash/include/dash/algorithm/Sort.h
rkowalewski: OK, right. I have simply "copied" it since it is a result of the API changes with futures. Its is modified now.

dash/include/dash/algorithm/Sort.h
rkowalewski: We do not skip the first found `it_skipped` here. If there is a first `it_skipped` it is processed in this while loop.

dash/include/dash/algorithm/Sort.h
rkowalewski: Done

dash/include/dash/algorithm/Sort.h
rkowalewski: Done

dash/include/dash/algorithm/Sort.h
rkowalewski: OK it is changed now. See below, it was originally "copied" from your modifications after API changes. Looks way better now.

dash/include/dash/algorithm/Sort.h
rkowalewski: See below, we do not support async proxies on `NArray`.

dash/include/dash/algorithm/Sort.h
rkowalewski: I think it is a relict of old code. Removed now.

dash/include/dash/algorithm/Sort.h
rkowalewski: Yes, it is. See the std::sort above.

dash/test/algorithm/SortTest.cc
rkowalewski: GoogleTest provides specific macros to compare floats (nearly equal). Our (custom) solution which is different from Google's provided solution here did not work for me at least. That is why I explicitly call the "correct" GoogleTest macro. I will fix that and provide a separate pull request for all unit tests.

rkowalewski: Furthermore we should not rely on internal GoogleTest APIs which may change.

rkowalewski:
@devreal Your review comments should all be addressed now. They include one additional minor bug fix which is the reason why `dash::sort` crashed with very large datasets / problem sizes. I found this after two days of excessive debugging which is why I added additional assert statements in the algorithm.

It further showed me that we should test our algorithms more on "real" distributed HPC platforms and not only in docker containers or local. While unit tests are sufficient for small functional testing complex and disreputable bugs arise with very large problems.

rkowalewski:
Finally we have some very interesting insights with extensive `dash::sort` [benchmarks](https://github.com/dash-project/dash-bench/tree/master/sort-bench). We are very efficient even on shared memory now. The code is also well tested and is ready to merge IMHO.
There are couple of people who want to use it.

devreal (approved):

PR #470, opened 2017-12-20
[WIP] Add support for GoogleBenchmark Benchmarks

fmoessbauer:
This PR includes the following changes

- Add support for custom CMake scripts in examples directory
- Do not "infect" with compiler flags in tests and examples (flags set for one target should only be set for this target and not all following targets)
- Automatically pull GBench if a target requests it


devreal:
Could you please quickly outline what the advantage of using GoogleBenchmark is? 

I am concerned about making GoogleBenchmarks a prerequisit for all examples. There are some machines out there that do not allow pulling from an outside git repo. Any attempt to clone a repository will result in a painfully long timeout... All our tests depend on GoogleTests, so I can easily disable them. With examples, only a subset of examples (`dash/examples/bench.*`) will require GoogleBenchmarks (eventually, I guess) and you might still want to build the remaining examples without manually installing GoogleBenchmarks first. Can we have a CMake flag to disable the benchmarks (that require external libraries)?

devreal:
Another thought on this: why don't we add GoogleTest and GoogleBenchmarks as git submodules to our project? It would be pulled once (and again if the user requests it) and not at every `make`. This also has the benefit that you can actually call make with tests enabled while not being connected to the Internet, i.e., while sitting on the train going through the middle of nowhere. For systems that don't allow connections to the outside, it would be much easier to clone the whole project once and not having to worry about building external dependencies manually.

fmoessbauer:
> Could you please quickly outline what the advantage of using GoogleBenchmark is?

It is sort of industry standard and will allow us to do regressions benchmarks. Later, these benchmarks could be evaluated automatically, which would reveal performance drops. It also has some nice build-in heuristics to check how often the benchmark has to be repeated until the results get stable.

> why don't we add GoogleTest and GoogleBenchmarks as git submodules

I also vote for that. This would also avoid some problems with the CMake external project handler.

fmoessbauer:
After #477 this PR has to be changed to use the new submodule based remote dependencies.

rkowalewski:
In general a very good idea. However, does the sampling of Google Benchmark work even in distributed environments? I cannot see many references about this topic.
The provided solution may have the same issue as described in #481.

devreal:
@fmoessbauer Is this still relevant or can we close this PR?

fmoessbauer:
@devreal : I prefer to keep the branch, but the PR can be closed. If we want to implement it in the future, we have to re-work the cmake stuff as it is probably not compatible with our recent changes and fixes.

Anyways, there are still many unclear points regarding google benchmark in distributed environments.

PR #471, opened 2017-12-20
set C and CXX STD centrally in CompilerFlags

fmoessbauer:
- set C and CXX std centrally in `CompilerFlags.cmake`
- use CMAKE STANDARD option if possible
- require c99, cxx11
- prefere c99, cxx14


dart-impl/base/CMakeLists.txt
```diff
@@ -118,6 +118,8 @@ target_link_libraries(
 set_target_properties(
   ${DASH_DART_BASE_LIBRARY} PROPERTIES
   COMPILE_FLAGS ${ADDITIONAL_COMPILE_FLAGS}
+  C_STANDARD ${DART_C_STD_PREFERED}
+  C_STANDARD_REQUIRED ${DART_C_STD_REQUIRED}
```
devreal, line 122: Why not set `C[XX]_STANDARD` globally in `CMakeExt/CompilerFlags.cmake`?

dhinf: I asked myself the same. Nethertheless the build works with our intel compiler now.

devreal: It didn't work earlier because?

dhinf: Because a function in onesided.h uses void as return type within a constexpr. That's forbidden in C++11 but allowed in C++14. At the Moment only the icpc (2018.1.163) has a problem with that, because the default is not C++14.

devreal: 1) If it's broken, where is the patch that fixes it? 
2) I don't see where that should be in current `development`. Should have been fixed a while ago: https://github.com/dash-project/dash/commit/a09a0434fa2ddf6cec35fc5448d4336e434bf314

fmoessbauer: Well, I prefer to explicitly set the target properties instead of just relying on some vague global state. However if you prefere, it should work to set [`CMAKE_CXX_STANDARD`](https://cmake.org/cmake/help/v3.1/variable/CMAKE_CXX_STANDARD.html#variable:CMAKE_CXX_STANDARD) globally.

@dhinf: It just works because your compiler supports C++14 and hence CMake chooses that. Anyways it should be fixed so that it is C++11 compliant.

devreal: @fmoessbauer Ahh, you're right, it's set as part the target properties. What happens with older CMake versions? Won't they trip if they don't know `C_STANDARD_REQUIRED`?

fmoessbauer: CMake should just ignore it, but well give it a try. Unfortunately I do not have such an ancient version (<3.0). It would make things so much easier if we could just raise minimum cmake to at least 3.0...

dhinf: I don't get it. Why we have to be C++11 compliant?

devreal: @dhinf IIRC, we decided (at the last F2F) to gradually move to C++14 and keep C++11 compliance for now until we know that there are no configurations that break. Until then we should try to support both standards. 

@fmoessbauer I tried with CMake 2.8.11 and found a glitch in the flag for Intel. Should be fixed. The unknown target properties are ignored :+1: 

In general, I'm in support for raising the CMake version requirements. Two out of three systems I quickly checked offer modules for CMake >3.0.0 (the ones at HLRS), the system at Riken only offers 2.8.11 so I'd have to build my own CMake first. This should be a project-wide decision though.

devreal (approved):
I think it looks good. Thanks :+1:

PR #475, opened 2018-01-05
manually check if blas library provides cblas interface

### LINKED ISSUE #474: cblas_* on arch linux not found
**Issued via support@dash-project.org**

I'm trying to build DASH in Arch Linux with the regular cblas package installed.

I get the following error when compiling:

libdash-mpi.a(SUMMA.cc.o): In function `void dash::internal::mmult_local<float>(float const*, float const*, float*, long long, long long, long long, dash::MemArrange)':
SUMMA.cc:(.text+0x32): undefined reference to `cblas_sgemm'
libdash-mpi.a(SUMMA.cc.o): In function `void dash::internal::mmult_local<double>(double const*, double const*, double*, long long, long long, long long, dash::MemArrange)':
SUMMA.cc:(.text+0x78): undefined reference to `cblas_dgemm'
collect2: error: ld returned 1 exit status

What cmake option do I need to edit to fix this?  Example programs that call cblas_dgemm compile successfully with the -lcblas flag.

I'm sure this is very simple, but I'm horrible with cmake.

Ben

fmoessbauer:
If cblas_* interface is not found (symbol does not exist in lib), try to manually find libcblas and substitute blas library with cblas.

Please test this change carefully with various Atlas / BLAS / CBLAS / LAPACK implementations. A minimal working example (dockerfile) for arch-linux is provided.

Fixes #474


devreal (approved):
:+1:

PR #476, opened 2018-01-08
Check presence of std::is_trivially_copyable

pascalj:
Relying on the compiler versions to find out whether std::is_trivially_copyable is present is error-prone. For example clang 5.0 broke because `__GNUC_` is defined.

Instead use cmake to verify that `std::is_trivially_copyable` (and `__has_trivial_copy` as a fallback) is present.

This might be more reasonable than [solving SAT](https://github.com/01org/tbb/blob/4c73c3b5d7f78c40f69e0c04fd4afb9f48add1e6/include/tbb/tbb_config.h) for every new compiler.


dash/include/dash/Types.h
```diff
@@ -218,13 +218,10 @@ template <class T>
...
-    ( defined(__GNUG__) && __GNUC__ >= 5 )
-              // The Cray compiler (as of CCE8.5.6) does not support
-              // std::is_trivially_copyable.
-           && std::is_trivially_copyable<T>::value
-#elif defined(__GNUG__) && __GNUC__ < 5
-           && std::has_trivial_copy_constructor<T>::value
+#ifdef STD_TRIVIALLY_COPYABLE_AVAIL
+              && std::is_trivially_copyable<T>::value
+#elif TRIVIAL_COPY_INTRINSIC_AVAIL
+              && __has_trivial_copy(T)
```
pascalj: NB: this was an alias in older GCC versions.

rkowalewski:
Note that this approach does not work if one relies on `make` as the build tool.

dash/CMakeLists.txt
```diff
@@ -212,6 +212,15 @@ if (ENABLE_SCALAPACK)
   endif()
 endif()
 
+if (HAVE_STD_TRIVIALLY_COPYABLE)
+  set (ADDITIONAL_COMPILE_FLAGS
+       ${ADDITIONAL_COMPILE_FLAGS} -DSTD_TRIVIALLY_COPYABLE_AVAIL)
```
devreal: I suggest to use the `DASH_` prefix for macros, i.e., `DASH_HAVE_STD_TRIVIALLY_COPYABLE`, and similar below.

CMakeExt/StdLib.cmake
```diff
@@ -0,0 +1,24 @@
...
+  int main() { return 0; }
+  " HAVE_STD_TRIVIALLY_COPYABLE)
+
+CHECK_CXX_SOURCE_COMPILES(
+  "
+  #include <type_traits>
+  template <class T> struct is_container_compatible :
+  public std::integral_constant<bool, __has_trivial_copy(T)> {};
+  int main() { return 0; }
+  " HAVE_TRIVIAL_COPY_INTRINSIC)
```
devreal: We could issue a warning if the compiler supports neither

devreal (changes requested):
Looks good, thanks! Just a minor change, see inline.

devreal:
> Note that this approach does not work if one relies on make as the build tool.

@rkowalewski Do you mean the plain makefiles or the Makefiles generated by CMake?

rkowalewski:
I mean our plain Makefiles.

devreal:
I guess there is a metric ton of features that break when using the plain makefiles. I can't even remember the reason why we still have these Makefiles around...

pascalj:
Thanks! Switched to `DASH_HAVE_STD_TRIVIALLY_COPYABLE` and added a warning. The check for the intrinsic will now only execute if `std::is_trivially_copyable` is missing.

devreal (approved):

PR #477, opened 2018-01-09
Include GoogleTest as git submodule

devreal:
Include the Googletest framework as git submodule instead of repeatedly pulling from the repository during make. 

The CMake script checks whether the submodule has been fetched already (and tries to update it if necessary) before setting it to the desired commit. 

Thus, the remote repository is only accessed once, either during (recursive) clone/pull or during the first CMake invocation after a non-recursive clone. This allows for reconfiguration and rebuild without active Internet connection. In case of a version bump (done by adjusting `GTEST_GIT_TAG`) the CMake script should automatically attempt to update the submodule if necessary (unfortunately, this is hard to test since I'd have to wait for `remote` to add commits that have not yet been fetched) 

See also #470


devreal:
Update: I tested the automatic update upon changing the git tag using a commit in Googletest that came in last night after I cloned the submodule and it works as expected.

fmoessbauer (approved):
Looks good. Just one point: We should add a note in the Readme how to pull dash including gtest:
`git clone --recursive <url>`.

devreal:
Good point! I will add that note.

PR #478, opened 2018-01-09
Compatibility for libc++

pascalj:
Two small changes are necessary to make dash compatible with [libc++](https://libcxx.llvm.org/). They're both because of libc++'s different internals.

### DIFF
--- a/dash/include/dash/Pair.h
+++ b/dash/include/dash/Pair.h
@@ -195,17 +195,38 @@ namespace dash {
     x.swap(y);
   }
 
+  namespace internal {
+    /**
+     * Duplicated here to not rely on an STL implementation
+     */
+    template <class T>
+    struct strip
+    {
+      typedef T type;
+    };
+    template <class T>
+    struct strip<std::reference_wrapper<T>>
+    {
+        typedef T& type;
+    };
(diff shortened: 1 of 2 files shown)


dash/include/dash/Pair.h
```diff
@@ -195,16 +195,37 @@ namespace dash {
...
+  struct __make_pair_return_impl<std::reference_wrapper<T>>
+  {
+      typedef T& type;
+  };
+  template <class T>
+  struct __make_pair_return
+  {
+      typedef typename __make_pair_return_impl<typename std::decay<T>::type>::type type;
+  };
+
```
devreal: Instead of the preceding underscores, please put `make_pair_return` (without underscores) in namespace `dash::internal`. Also, the naming is not entirely intuitive. Why not call it `dash::internal::decay_and_strip`? (so it's clear that this is our version of the GNU `std::__decay_and_strip`)

devreal (changes requested):

pascalj:
Yes, that makes sense. I renamed it to `decay_and_strip` and moved it into the `internal` namespace.

devreal (approved):

Issue #481, opened 2018-01-22
CMake cannot find local installation of GoogleTest


rkowalewski:
#477 added GoogleTest as a submodule. However, it does not work if I have a local installation of GoogleTest. The current solution suffers from two issues:

- If I pass both `GEST_LIBRARY_PATH` and `GEST_INCLUDE_PATH`, `GTEST_FOUND` is not set to true which in turn leads to skip tests in the build process.
- I can set both environment variables and CMake seems not to really check if it is a valid GoogleTest library. Shouln't we use something like `find_library` or `find_package` to find the necessary headers of GoogleTest which is the approach in other external libraries?

devreal:
> If I pass both GEST_LIBRARY_PATH and GEST_INCLUDE_PATH, GTEST_FOUND is not set to true which in turn leads to skip tests in the build process.

See #482 

> I can set both environment variables and CMake seems not to really check if it is a valid GoogleTest library.

That should also be fixed. I guess we could also test for valid versions. Patches are welcome :)

> Shouln't we use something like find_library or find_package to find the necessary headers of GoogleTest which is the approach in other external libraries?

Cmake seems to ship with `FindGTest` starting with 3.0.2 (at least that is the earliest version I can find). We would bind ourselves to CMake 3, which in turns means some people on some systems (including me) have to compile CMake first...

rkowalewski:
Fixed by #482

PR #482, opened 2018-01-22
Set GTEST_FOUND on local installation of GoogleTest

devreal:
See #481


rkowalewski (approved):
Looks good and works. Thanks your your changes.

PR #485, opened 2018-02-01
Coarray atomics fix

fmoessbauer:
Two minor fixes to make dash atomics more consistent. Thanks to @rkowalewski for reporting.


rkowalewski (approved):
Looks consistent now. Thanks.

PR #486, opened 2018-02-07
Introducing DASH Iterator Traits

rkowalewski:
We have talked couple of times about introducing DASH iterator traits. I have finally found some time to start implementing it.
The main reason is that as soon as we change the internal template parameter types for `GlobIter`, `GlobViewIter` and their other friends, our DASH algorithms do not compile anymore. This is because the iterator types are hard-coded into the algorithms interface which of course is not really a generic interface. We are on the way to implement additional global iterators (e.g. `GlobListIter`, `HaloIter`, etc.) and all of them should be applicable to our algorithms. If one cannot use the generic variant we always can specialize it for specific iterator traits instead of the concrete iterator type. I struggled with this stuff while integrating memory spaces and expect to need it for additional features such as Alpakka and so on.

The basic interface can be found in the [IteratorTraits.h](https://github.com/dash-project/dash/compare/development...rkowalewski:feat-iterator-traits?expand=1#diff-a5900ca4cfdf9ac684b4ec01d867e706R65). It is simply an extension of `std::iterator_traits` and I not not claim to be complete. This is subject to discussion here.

I modified couple of algorithms to use the iterator traits interface. A very nice example can be found in [Transform.h](https://github.com/dash-project/dash/compare/development...rkowalewski:feat-iterator-traits?expand=1#diff-4b4f4f666df31f6b6e1196aa861ec43bR444). It shows how we can use tag dispatching based on the iterator traits to delegate to the proper implementation. The public interface is not "polluted" with implementation specific types and `static_assert` statements tell the user what variants we currently support. As a positive side effect we not not need the to specify the value type for `dash::transform` anymore (see `TransformTest.cc`).
Tag dispatching in general accords very well with future features like execution policies and enables to hide the implementation details in internal namespaces. If tag dispatching is too much overkill we can always use SFINAE. An example for this can be found in [MinMax.h](https://github.com/dash-project/dash/compare/development...rkowalewski:feat-iterator-traits?expand=1#diff-e80ab11ec934ebb2b008bc1bca927879R156). It uses SFINAE to decide if we call `dash::min_element` on a global or a local (STL) iterator. I am not claiming that this is the best solution here, comments are always welcome.

However, we have an easily extensible iterator traits interface now and I think it greatly simplifies changing internal implementation details.

### DIFF
--- a/dash/include/dash/algorithm/Find.h
+++ b/dash/include/dash/algorithm/Find.h
@@ -24,7 +24,7 @@ GlobIter find(
   GlobIter   first,
   /// Iterator to the final position in the sequence
   GlobIter   last,
-  /// Value which will be assigned to the elements in range [first, last)
+  /// Value which is searched for using operator==
   const ElementType                  & value)
 {
 
--- a/dash/include/dash/algorithm/Find.h
+++ b/dash/include/dash/algorithm/Find.h
@@ -17,17 +17,24 @@ namespace dash {
  * \ingroup     DashAlgorithms
  */
 template<
-  typename ElementType,
-  class    PatternType>
-GlobIter<ElementType, PatternType> find(
+  typename GlobIter,
+  typename ElementType>
+GlobIter find(
   /// Iterator to the initial position in the sequence
-  GlobIter<ElementType, PatternType>   first,
+  GlobIter   first,
   /// Iterator to the final position in the sequence
-  GlobIter<ElementType, PatternType>   last,
+  GlobIter   last,
   /// Value which will be assigned to the elements in range [first, last)
   const ElementType                  & value)
 {
-  using p_index_t = typename PatternType::index_type;
+
+  using iterator_traits = dash::iterator_traits<GlobIter>;
+
+  using p_index_t = typename iterator_traits::index_type;
+
+  static_assert(
+      std::is_same<typename iterator_traits::value_type, ElementType>::value,
+      "element types do not match");
 
   if(first >= last) {
     return last;
@@ -95,29 +102,28 @@ GlobIter<ElementType, PatternType> find(
  *
  * \ingroup     DashAlgorithms
  */
-template<
-  typename ElementType,
-  class    PatternType,
-  class    UnaryPredicate >
-GlobIter<ElementType, PatternType> find_if(
-  /// Iterator to the initial position in the sequence
-  GlobIter<ElementType, PatternType>   first,
-  /// Iterator to the final position in the sequence
-  GlobIter<ElementType, PatternType>   last,
-  /// Predicate which will be applied to the elements in range [first, last)
-  UnaryPredicate                       predicate)
+template <typename GlobIter, typename UnaryPredicate>
+GlobIter find_if(
+    /// Iterator to the initial position in the sequence
+    GlobIter first,
+    /// Iterator to the final position in the sequence
+    GlobIter last,
+    /// Predicate which will be applied to the elements in range [first, last)
+    UnaryPredicate predicate)
 {
-  typedef typename PatternType::index_type index_t;
+  using iterator_traits = dash::iterator_traits<GlobIter>;
 
-  auto & team        = first.pattern().team();
-  auto myid          = team.myid();
+  using index_t = typename iterator_traits::index_type;
+
+  auto& team = first.pattern().team();
+  auto  myid = team.myid();
   /// Global iterators to local range:
-  auto index_range   = dash::local_range(first, last);
-  auto l_first       = index_range.begin;
-  auto l_last        = index_range.end;
+  auto index_range = dash::local_range(first, last);
+  auto l_first     = index_range.begin;
+  auto l_last      = index_range.end;
 
-  auto l_result      = std::find_if(l_first, l_last, predicate);
-  auto l_offset      = std::distance(l_first, l_result);
+  auto l_result = std::find_if(l_first, l_last, predicate);
+  auto l_offset = std::distance(l_first, l_result);
   if (l_result == l_last) {
     l_offset = -1;
   }
@@ -133,11 +139,8 @@ GlobIter<ElementType, PatternType> find_if(
 
   for (auto u = 0; u < team.size(); u++) {
     if (l_results[u] >= 0) {
-      auto g_offset = first.pattern()
-                           .global_index(
-                              u,
-                              { l_results[u] });
-      result = first + g_offset - first.pos();
+      auto g_offset = first.pattern().global_index(u, {l_results[u]});
+      result        = first + g_offset - first.pos();
       break;
     }
   }
@@ -156,17 +159,16 @@ GlobIter<ElementType, PatternType> find_if(
  *
  * \ingroup     DashAlgorithms
  */
-template<
-  typename ElementType,
-  class    PatternType,
-	class    UnaryPredicate>
-GlobIter<ElementType, PatternType> find_if_not(
-  /// Iterator to the initial position in the sequence
-  GlobIter<ElementType, PatternType>   first,
-  /// Iterator to the final position in the sequence
-  GlobIter<ElementType, PatternType>   last,
-  /// Predicate which will be applied to the elements in range [first, last)
-	UnaryPredicate                       predicate)
+template <
+    typename GlobIter,
+    class UnaryPredicate>
+GlobIter find_if_not(
+    /// Iterator to the initial position in the sequence
+    GlobIter first,
+    /// Iterator to the final position in the sequence
+    GlobIter last,
+    /// Predicate which will be applied to the elements in range [first, last)
+    UnaryPredicate predicate)
 {
   return find_if(first, last, std::not1(predicate));
 }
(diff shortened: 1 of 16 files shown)


dash/include/dash/algorithm/Find.h
```diff
@@ -17,17 +17,24 @@ namespace dash {
...
 {
-  using p_index_t = typename PatternType::index_type;
+
+  using iterator_traits = dash::iterator_traits<GlobIter>;
+
+  using p_index_t = typename iterator_traits::index_type;
+
+  static_assert(
+      std::is_same<typename iterator_traits::value_type, ElementType>::value,
+      "element types do not match");
```
devreal: Can we enforce the `ElementType` in the function signature? Currently it would blow right in your face if you mix signed/unsigned integers for example, which is not very user-friendly.

dash/include/dash/algorithm/Find.h
```diff
@@ -17,17 +17,24 @@ namespace dash {
...
+  typename GlobIter,
+  typename ElementType>
+GlobIter find(
   /// Iterator to the initial position in the sequence
-  GlobIter<ElementType, PatternType>   first,
+  GlobIter   first,
   /// Iterator to the final position in the sequence
-  GlobIter<ElementType, PatternType>   last,
+  GlobIter   last,
   /// Value which will be assigned to the elements in range [first, last)
```
devreal: While you're at it, please fix the documentation :)

dash/include/dash/algorithm/Find.h
```diff
@@ -95,29 +102,28 @@ GlobIter<ElementType, PatternType> find(
...
-  auto l_first       = index_range.begin;
-  auto l_last        = index_range.end;
+  auto index_range = dash::local_range(first, last);
+  auto l_first     = index_range.begin;
+  auto l_last      = index_range.end;
 
-  auto l_result      = std::find_if(l_first, l_last, predicate);
-  auto l_offset      = std::distance(l_first, l_result);
+  auto l_result = std::find_if(l_first, l_last, predicate);
+  auto l_offset = std::distance(l_first, l_result);
```
devreal: Only compute distance if necessary, e.g., else branch below.

dash/include/dash/algorithm/ForEach.h
```diff
@@ -28,17 +28,14 @@ namespace dash {
...
-  UnaryFunction                              func)
+    /// Iterator to the initial position in the sequence
+    const GlobInputIt& first,
+    /// Iterator to the final position in the sequence
+    const GlobInputIt& last,
+    /// Function to invoke on every index in the range
+    UnaryFunction func)
 {
   /// Global iterators to local index range:
   auto index_range  = dash::local_index_range(first, last);
```
devreal: Can we statically check whether `GlobInputIt` is a global pointer to avoid ugly compiler errors? (I think it's done in `generate` already)

devreal (changes requested):
Looks good to me, some minor comments inline.

dash/include/dash/algorithm/Find.h
rkowalewski: There are two ways to achieve is. Either we apply SFINAE and use std::enable_if like in MinMax.h which certainly  does not provide user-friendly compiler errors. Or we enforce the user to explicitly specifiy the `ElementType` template parameter like in `dash::transform` before this pull request which is not user-friendly either. I no not know any other possibility right now.
IMHO the `static_assert` is the cleanest way as it tells the user about mismatched element types.

dash/include/dash/algorithm/ForEach.h
rkowalewski: Done

dash/include/dash/algorithm/Find.h
rkowalewski: Done

dash/include/dash/algorithm/Find.h
rkowalewski: Done

dash/include/dash/algorithm/Find.h
devreal: How about this:

```
template<typename GlobIter>
GlobIter find(
  GlobIter begin, 
  GlobIter end, 
  const dash::iterator_traits<GlobIter>::value_type & value)
{
...
}
```

IIRC, the `static_assert` would trigger on something like this:
```
dash::Array<unsigned int> arr;
dash::find(arr.begin(), arr.end(), 0);
```

rkowalewski: Of course, this works. I was thinking kind of too complicated. However, while discussing again about this: I think we should not enforce that the `value_type`  which which we are looking for and the `value_type` of the containers match. All we need is equality comparable. The STL follows exactly this approach. And shouldn't we do it in the same manner?

rkowalewski: It is actually the programmer's fault if there are misleading complicated error warnings. That is what we can see while working with common STL algorithms. Our interface should be STL compatible which is why I actually removed the static_assert as well.

devreal: Works for me.

devreal (approved):

fmoessbauer:
Aside from the aspect that this PR breaks existing code, this looks good so far. And as we still have no 0.3.0 release, I have no problem with that. Good Changes!

Just one minor point: Many algorithm functions take the iterators by pass-by-value semantics. This should be pass-by-reference, if possible.

ddiefenthaler:
@fmoessbauer What exactly do you mean by pass-by-reference? I can't see a reason for the iterators to be call-by-reference, as the algorithm has to copy them nonetheless if it wants to use them.
Otherwise it would result in quite unexpected behavior, as for example after a call to `for_each` the given iterator `first` would then be equal to `last`. See [Possible Implementation](http://en.cppreference.com/w/cpp/algorithm/for_each#Possible_implementation) of std::for_each on cppreference.com.
Also the STL uses call-by-value for iterators.

fmoessbauer:
@ddiefenthaler : I meant call-by-reference, and of course these references have to be const. And not all algorithms have copy the iterators. For just one example have a look here: https://github.com/dash-project/dash/pull/486/files#diff-2a094db83d9a82b562f811172c8d8624R19

It depends on the implementation of the algorithm, but there are various cases where we use call-by-value, where call-by-const-reference would be better.

rkowalewski:
I wouln't agree with that either. Passing iterators by value instead of a const reference is actually more efficient as we can [see](https://godbolt.org/g/qBMfh8). This example has two methods:

1) `bool range_equals(Iter1 first1, Iter1 last1, Iter2 first2)`
2) `bool range_equals_const(Iter1 const & first1, Iter1 const & last1, Iter2 const & first2)`

For clarity I avoided inlining by the compiler. What we can see is that the code of 2) requires more loads from memory compared to 1), both on the callee and in particular on the caller side. Note that I tested only with gcc 5.4.0, however, I assume that more recent version do it even better.
If you remove the `noinline` attribute you can see that the compiler inlines everything anyway and copies are eliminated. I cannot see a great difference then.

The takeway here is that compilers can eliminate copies better compared to references as long as the objects are trivially copyable. And that is the case for both STL and our global iterators which hold only couple of pointers and primitive types.

That is also what you always hear when all these C++ fathers give a talk. Just make your types trivially copyable and the compiler can optimize in most cases.

EDIT: Just to give a hint. The relevant lines on the right hand side are 55 - 88 (calle, i.e., definition of these two funtions) and 315 - 317 (caller)

devreal:
> The takeway here is that compilers can eliminate copies better compared to references as long as the objects are trivially copyable. And that is the case for both STL and our global iterators which hold only couple of pointers and primitive types.

`GlobIter` is currently not trivially copyable but afaics we can make `GlobIter` that way. In that case we should think about prohibiting `GlobIter` (and `GlobListIter`) from being elements in DASH containers (which won't work due to local pointers, in contrast to `GlobRef`). 

Our iterators are most likely bigger than STL iterators (48B for `GlobIter`, 24B for `GlobListIter`) but if the compiler can eliminate the copies the size is a non-issue of course.

rkowalewski:
I think my statement above is not correct in the sense that [copy ellision](http://en.cppreference.com/w/cpp/language/copy_elision) does NOT require trivially copyable types. It just needs to be copyable by a copy constructor. I am just claiming that we can leave it as is.

Moreover, I think we cannot make `GlobIter` trivially copyable since this requires that there are no user-provided copy / move constructors.

rkowalewski:
Credits to @fmoessbauer. Global Iterators are now trivially copyable.

dash/include/dash/iterator/GlobIter.h
```diff
@@ -430,7 +406,7 @@ class GlobIter
    */
   constexpr bool is_local() const
   {
-    return (_myid == lpos().unit);
+    return (dash::Team::All().myid() == lpos().unit);
```
devreal: Caution, that's not the correct team! It should be `_pattern->team().myid()` (`Team::All()` is only used in the default constructor...) 

Same problem below...

devreal (changes requested):

devreal (approved):

devreal:
Gnah, only looked at the one commit...

fmoessbauer:
@rkowalewski: Is it still WIP, or can we merge?

rkowalewski:
Please merge. ;)

PR #492, opened 2018-03-02
Move content of `.gitignore` to `.gitmodules`.

bertwesarg:
Git does not read any full fledged config file from the repository, in
particular any `.gitconfig`. Thus this file is useless, without a
`include.path` statement in any config file Git is reading (i.e.,
`.git/config`). Though it reads `.gitmodules` and as the content of
`.gitconfig` is sub-modules related anyway, it has an effect, if it is in
this file. Do so.

### DIFF
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,2 +1,0 @@
-[submodule "vendor/googletest"]
-    ignore = all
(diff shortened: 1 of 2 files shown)


devreal (approved):
:+1:

PR #493, opened 2018-03-02
Fix support for 64bit integer types on 32bit systems

devreal:
@ddiefenthaler Please test on 32 bit :)


ddiefenthaler (approved):
Thanks,
Compiles for me, and the testuite with 4 process also suceeds for me. Looks good!

PR #494, opened 2018-03-02
Add support for long double in DASH container (incl. Atomic)

### LINKED ISSUE #355: Maximal width of dash::is_atomic_compatible types
Currently the trait `dash::is_atomic_compatible` matches the size of the type against `std::size_t` to determine if atomic access is possible (Defined in [Types.h](https://github.com/dash-project/dash/blob/1978c10151d7fb000ecc3bd6e50099fd3835256e/dash/include/dash/Types.h#L235)).
This was introduced and already shortly discussed in pull request #338.

Is `std::size_t` really the correct type to match against?
I tried to look up the maximal supported width of an atomic operation
and found the extension **Double-wide compare-and-swap** in the wikipedia entry to [Compare-and-swap](https://en.wikipedia.org/wiki/Compare-and-swap#Extensions).
It looks like atomic operation on twice the width of `std::size_t` are possible.

Or is this a limitation of the _One-Sided Communication_ in MPI?
Otherwise `uintmax_t` might be a more appropriate type to check against.

This questions arises as in one test a `dash::Atomic<double>` is used which breaks in my 32 bit VM.

devreal:
Some more details: 
This PR adds support for `long double` in DART and DASH containers. It also imposes restrictions on atomic operations, e.g., `compare_exchange` is not supported on floating point values and arithmetic operations on punned types are likely to return an unexpected result. Hence, only bitwise operations and replace/no-op are supported for punned types.

@ddiefenthaler Please test on 32bit ;)

Fixes #355


ddiefenthaler (approved):
I can confirm that this fixes issue #355 

For me it looks good. :+1:
I would still recommend a second review approving this as well.

devreal:
The failed test yesterday might have been caused by misalignment of the memory handed out by MPI for `long double`. I will investigate this further but I think the implementation on our side should be fine.

fmoessbauer (approved):

PR #495, opened 2018-03-05
Fix move assignment / ctor in Array&Matrix

devreal:
We cannot use the default move assignment because `dash::Array` and `dash::Matrix` require registration of a deallocator.


fmoessbauer:
@devreal: Are you really sure about that? The Container itself does not own any memory, but globmem does. And the globmem should be responsible for deallocation and keeping track of the lifetime. Do I miss something?

And by the way: Don't we have to unregister the obsolete deallocator on move?

devreal:
The problem is an example like this:

```
dash::Array<int> arr;

arr = dash::Array<int>(dash::size());
```

The move assignment operator previously did not register a deallocator (as done by `dash::Array::allocate`), leading to a potential memory leak.

devreal:
To clarify: this is only a problem if the array is defined globally or in the scope of `main`, as in both cases the destructor is not called before `dash::finalize` is called.

fmoessbauer:
Ah, got it. However don't we have to unregister the deallocator on the obsolete object?

Consider the following:
```cpp
dash::Array<int> arr;
{
  dash::Array<int> arr2(dash::size());
  arr = arr2;
}
// arr2 leaves scope, hence deallocator is called?
// here we have to ensure that the deallocator correctly handles a `globmem = nullptr`
```
And are there any unit tests?

rkowalewski:
@fmoessbauer your provided example should hopefully never compile since it uses the deleted copy constructor. If you instead move it this should work, of course.

fmoessbauer:
@rkowalewski Well, that does not compile, as I missed the move. However I am not sure that this works correctly (with move) because the deallocator is called twice.

devreal:
I missed one important corner stone (that was previously broken): if the array is never allocated, no attempts should be made to deregister the deallocator. 

@fmoessbauer Yes, the d'tor only deallocates if it still owns the memory (see https://github.com/dash-project/dash/blob/bug-dash-arraymove/dash/include/dash/Array.h#L1377) and it unregisters the deallocator if it has been previously registered.

fmoessbauer (approved):
Now it looks good to me.

rkowalewski (approved):
fine for me as well.

PR #496, opened 2018-03-05
Fix setting the C++ standard in the DASH compiler wrapper

devreal:
The standard flag got lost in the DASH compiler wrapper somewhere on the way.


rkowalewski:
Very good I had the same problem here on SuperMUC and planned a pull request for this.

fmoessbauer (approved):

PR #497, opened 2018-03-05
Remove erroneous DART onesided test

devreal:
The test is erroneous because it reads outside of the allocated buffer bounds.


rkowalewski:
Am I correct that the fix is just removing a test? What exactly was wrong?

devreal:
Correct. The test is nonsensical as it allocates one block per unit and then tries to copy two blocks using DART. We have agreed that DART is unaware of the global memory layout so this produces invalid memory accesses. There are other tests that test DART copy functionality.

rkowalewski (approved):
I agree. Good catch

PR #498, opened 2018-03-05
Prohibit implicit conversion between different GlobPtr

devreal:
Prohibit implicit conversion between different GlobPtr types and provide explicit conversion operator

Equivalent in C++
```
double *dbuf = new double[N];
int *ibuf = dbuf; // << not allowed
int *ibuf = static_cast<int*>(dbuf); // allowed, you know what you do!
```

See new test case for an example.

Thanks to @tuxamito for reporting this

### DIFF
--- a/dash/include/dash/GlobPtr.h
+++ b/dash/include/dash/GlobPtr.h
@@ -197,9 +197,32 @@ public:
   constexpr GlobPtr(const self_t & other)      = default;
 
   /**
+   * Copy constructor (memory space conversion).
+   */
+  template <class MemSpaceT>
+  constexpr GlobPtr(const GlobPtr<ElementType, MemSpaceT> & other)
+  : _rbegin_gptr(other._rbegin_gptr)
+  , _mem_space(reinterpret_cast<const MemorySpace *>(other._mem_space))
+  , _lsize(other._lsize)
+  , _unit_end(other._unit_end)
+  { }
+
+  /**
    * Assignment operator.
    */
   self_t & operator=(const self_t & rhs)       = default;
+
+  /**
+   * Assignment operator (memory space conversion).
+   */
+  template<class MemSpaceT>
+  self_t & operator=(const GlobPtr<ElementType, MemSpaceT>& other) {
+    _rbegin_gptr = other._rbegin_gptr;
+    _mem_space   = reinterpret_cast<const MemorySpace *>(other._mem_space);
+    _lsize       = other._lsize;
+    _unit_end    = other._unit_end;
+    return *this;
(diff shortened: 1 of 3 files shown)


rkowalewski:
While I have not checked this, I assume it may be a problem for value type dash::Atomic<...>. You may loose all memory semantics and run into lots of hazards.

devreal:
@rkowalewski Can you give an example that could be problematic? I'm happy to add tests for anything that should work :)

devreal:
@rkowalewski On my way home I was reasoning about the cast and found this pointer-cast here fishy: 

https://github.com/dash-project/dash/pull/498/files#diff-2ac68e3401d8de196f46a596bed4b4b5R205

Is that really legal? I haven't dug into the memory spaces but simply casting a pointer to different types that are not related seems dangerous. Maybe memory spaces need a cast operator as well? 

In principle I don't like the fact that the memory space is part of the GlobPtr type. It would seem cleaner to use virtual inheritance here and take the underlying memory space out of the pointer type. But virtual inheritance... ^^

devreal:
Looking at the way `GlobPtr` uses the memory space, the only usage seems to be the query for the `local_size` of the allocation at a specific unit in order to transparently increment the pointer across units. One way to solve this without having the memory space pollute the `GlobPtr` type is to pass a lambda that binds the underlying memory space object and can be used to query the local size, similar to the way `dash::Future` works. That way the memory space stays out of the type and cannot be altered by accident while moving global pointers around. It would also simplify the usage of `GlobPtr` as currently a user has to guess a memory space type just to declare a `GlobPtr` variable.

rkowalewski:
I made lots of changes with memory spaces and will submit a pull request before our DASH meeting in Munich. One change amongst others is that the Memory has to be in the type of the pointer. However it is not provided as template parameter in pointer conversion, assignment operator, etc. This way the memory space stays the same and the user cannot mix different memory spaces by accident.
The memory space has to be part of the pointer type because there are different semantics, e.g. with NVM and DRAM. We will hear more about this in our DASH meeting.

devreal:
> The memory space has to be part of the pointer type because there are different semantics, e.g. with NVM and DRAM.

The *semantics* exposed by `GlobPtr` should be the same no matter the underlying memory but the *implementation* of course differs.  So if a function accepts `GlobPtr<int>` it shouldn't matter what the underlying memory space is. But OK, let's discuss this at the F2F. There are other issues that I want to bring up there, e.g., using `GlobPtr` in DASH data structures.

devreal:
Stale, not needed anymore. (See #515)

PR #500, opened 2018-03-06
Fix local allocation that are not power of 2

devreal:
Previously, an allocation of 12 Byte would have chopped down to 8 and the next allocation could have overwritten parts of the first allocation. 

Thanks to @tuxamito for finding and reporting this one. Mea cupla!


tuxamito (approved):
This bug seems to be solved for the use cases that previously failed.

PR #502, opened 2018-03-09
Fixes to dash::fill

devreal:
1) Remove strided fill as it is not needed anymore and potentially hurts performance
2) Fix `omp parallel for` to actually perform work sharing


fmoessbauer (approved):
LGTM. While working on this piece of code, you could remove the useless if in L52

rkowalewski (approved):

PR #503, opened 2018-03-09
Feat halo

dhinf:
- some renamings
-  new stencil spec handling
  - HaloMatrixWrapper can take more than on stencil spec and builds the halo block with the maxima of    the stencil points 
  - stencil points now have coefficients
  - inner, boundary iterator managed by stencil operator now
  - HaloMatrixWrapper creates stencil operators dependent on the stencil spec
- tests optimized -> don't take ages on more than one node anymore


dash/include/dash/halo/Halo.h
```diff
@@ -11,55 +11,150 @@
...
 
+  /**
+   * Constructor
+   *
+   * Custom stencil point values for all dimensions and default
+   * coefficient = 1.0.
+   */
+  template <typename... Values>
+  constexpr StencilPoint(
+    typename std::enable_if<sizeof...(Values) == NumDimensions - 1,
```
devreal: I'd prefer to have a `static_assert` here to give a meaningful error message.

dash/include/dash/halo/Halo.h
```diff
@@ -11,55 +11,150 @@
...
+  }
 
   // TODO as constexpr
+  /**
+   * Returns maximum distance to center over all dimensions
+   */
   int max() const {
     int16_t max = 0;
     for(dim_t i(0); i < NumDimensions; ++i)
       max = std::max((int) max, (int) std::abs(this->_values[i]));
```
devreal, line 69: This looks like `std::max_element` to me.

dash/include/dash/halo/Halo.h
```diff
@@ -11,55 +11,150 @@
...
+  CoeffT _coefficient = 1.0;
+};  // StencilPoint
+
+/**
+ * A collection of stencil points (\ref Stencil)
+ * e.g. StencilSpec<dash::StencilPoint<2>, 2,2>({StencilPoint<2>(-1,0),
+ * StencilPoint<2>(1,0)}) -> north and south
+ */
+template <typename StencilPointT, dim_t NumDimensions,
+          std::size_t NumStencilPoints>
```
devreal: Is there a valid case for `StencilPointT<N>` and `N != NumDimensions`? If not, can't we default `StencilPointT` to `StencilPoint<NumDimensions>` to save the user some typing?

dash/include/dash/halo/Halo.h
```diff
@@ -11,55 +11,150 @@
...
-  using Self_t = StencilSpec<NumDimensions, NumStencilPoints>;
+  using Self_t = StencilSpec<StencilPointT, NumDimensions, NumStencilPoints>;
 
 public:
-  using stencil_size_t = std::size_t;
-  using Stencil_t      = Stencil<NumDimensions>;
-  using Specs_t        = std::array<Stencil_t, NumStencilPoints>;
+  using stencil_size_t  = std::size_t;
+  using stencil_index_t = std::size_t;
+  using Specs_t         = std::array<StencilPointT, NumStencilPoints>;
```
devreal: `Specs_t` -> `StencilArray_t`?

dash/include/dash/halo/Halo.h
```diff
@@ -11,55 +11,150 @@
...
-  constexpr const Stencil_t& operator[](std::size_t index) const {
+  /**
+   * Finds the stencil point index for a given \ref StencilPoint
+   *
+   * \return The index and true if the given stecil point was found,
+   *         else the index 0 and false.
+   *         Keep in mind that index 0 is only a valid index, if the returned
+   *         bool is true
+   */
+  const std::pair<stencil_index_t, bool> index(StencilPointT stencil) const {
```
devreal, line 144: `std::optional` is calling :)

dash/include/dash/halo/Halo.h
```diff
@@ -234,16 +409,20 @@ class RegionSpec : public Dimensional<uint8_t, NumDimensions> {
     return !(*this == other);
   }
 
-  // returns the highest dimension with changed coordinates
+  /**
+   * Returns the highest dimension with region values != 1
+   */
   dim_t relevant_dim() const { return _rel_dim; }
 
+  /// returns the number of coordinates unequal the center (1) for all
```
devreal: Comment stile (`///` vs `/* */`)

dash/include/dash/halo/Halo.h
```diff
@@ -294,8 +476,21 @@ class HaloSpec {
...
+      set_region_spec(stencil_combination);
+      while(next_region(stencil, stencil_combination)) {
+        set_region_spec(stencil_combination);
+      }
+    }
+  }
+
+  template <typename StencilSpecT, typename... Args>
+  HaloSpec(const StencilSpecT& stencil_spec, const Args&... stencil_specs)
+  : HaloSpec(stencil_specs...) {
```
devreal: What happens to the leading `stencil_spec` argument?

dash/include/dash/halo/Halo.h
```diff
@@ -294,8 +476,21 @@ class HaloSpec {
...
-    for(const auto& stencil : stencil_specs.specs()) {
+  HaloSpec(const StencilSpecT& stencil_spec) {
+    for(const auto& stencil : stencil_spec.specs()) {
+      auto stencil_combination = stencil;
+
+      set_region_spec(stencil_combination);
+      while(next_region(stencil, stencil_combination)) {
+        set_region_spec(stencil_combination);
+      }
+    }
```
devreal: The code above seems to be duplicated below...

dash/include/dash/halo/Halo.h
```diff
@@ -726,33 +949,36 @@ class HaloBlock {
...
+  using HaloExtsMax_t     = std::array<HaloExtsMaxPair_t, NumDimensions>;
+
 public:
+  /**
+   * Constructor
+   */
   HaloBlock(GlobMem_t& globmem, const PatternT& pattern, const ViewSpec_t& view,
-            const HaloSpec_t&  halo_reg_spec,
-            const CycleSpec_t& cycle_spec = CycleSpec_t{})
+            const HaloSpec_t&      halo_reg_spec,
```
devreal, line 988: It's entirely not clear to me what happens in this function. Some breakdown into smaller functions or comments would be nice :)

dash/include/dash/halo/HaloMatrixWrapper.h
```diff
@@ -45,26 +65,25 @@ class HaloMatrixWrapper {
...
-  HaloMatrixWrapper(MatrixT& matrix, const StencilSpecT& stencil_spec,
-                    const CycleSpec_t& cycle_spec = CycleSpec_t())
-  : _matrix(matrix), _stencil_spec(stencil_spec), _cycle_spec(cycle_spec),
-    _halo_reg_spec(stencil_spec), _view_local(matrix.local.extents()),
+  /**
+   * Constructor that takes \ref Matrix, a \ref GlobalBoundarySpec and a user
+   * defined number of stencil specifications (\ref StencilSpec)
+   */
+  template <typename... StencilSpecT>
+  HaloMatrixWrapper(MatrixT& matrix, const GlobBoundSpec_t& cycle_spec,
```
devreal, line 79: Same here: some additional documentation on what the code does would be nice

dash/include/dash/halo/HaloMatrixWrapper.h
```diff
@@ -200,104 +217,140 @@ class HaloMatrixWrapper {
...
     auto it_find = _region_data.find(index);
     if(it_find != _region_data.end())
       update_halo_intern(it_find->second, true);
   }
 
+  /**
+   * Waits until all halo updates are finished. Only useful for asyncronous
+   * halo updates.
+   */
   void wait() {
```
devreal: Eventually, region-specific `wait` and *test* would be nice :)

dash/include/dash/halo/HaloMatrixWrapper.h
```diff
@@ -200,104 +217,140 @@ class HaloMatrixWrapper {
...
-  const HaloMemory_t& halo_memory() const { return _halomemory; }
+  /**
+   * Returns the halo memory management object \ref HaloMemory
+   */
+  HaloMemory_t& halo_memory() { return _halomemory; }
 
+  /**
+   * Returns the underlying NArray
+   */
   MatrixT& matrix() { return _matrix; }
```
devreal, line 340: `const` versions of the two functions `matrix` and `halo_memory` would be nice

dash/include/dash/halo/Halo.h
```diff
@@ -68,38 +163,68 @@ class StencilSpec {
...
  */
-enum class Cycle : uint8_t {
+enum class BoundaryProp : uint8_t {
   /// No global boundary Halos
   NONE,
   /// Global boundary Halos with values from the opposite boundary
   CYCLIC,
   /// Global boundary Halos with predefined fixed values
-  FIXED
+  CUSTOM
```
devreal: I'd prefer `FIXED` instead of `CUSTOM`

dash/include/dash/halo/HaloMatrixWrapper.h
```diff
@@ -322,36 +375,64 @@ class HaloMatrixWrapper {
...
-      ++outputBegin;
+  /**
+   * Crates \ref HaloStencilOperator for a given \ref StencilSpec.
+   * Asserts whether the StencilSpec fits in the provided halo regions.
+   */
+  template <typename StencilSpecT>
+  HaloStencilOperator<Element_t, Pattern_t, StencilSpecT> stencil_operator(
+    const StencilSpecT& stencil_spec) {
+    for(const auto& stencil : stencil_spec.specs()) {
+      assert(stencil.max()
```
devreal: Use `DASH_ASSERT[_MSG]` instead of `assert`

dash/include/dash/halo/HaloMatrixWrapper.h
```diff
@@ -322,36 +375,64 @@ class HaloMatrixWrapper {
...
+             <= _halo_spec.extent(RegionSpec<NumDimensions>::index(stencil)));
     }
+
+    return HaloStencilOperator<Element_t, Pattern_t, StencilSpecT>(
+      _haloblock, _halomemory, stencil_spec, _view_local);
   }
 
 private:
   struct HaloData {
```
devreal, line 455: That's not the data but the handle for the transfer --> `HaloTransfer`?

devreal (changes requested):
Thanks a lot for the comments and improvements. Some remarks inline

dash/include/dash/halo/Halo.h
dhinf: max_element isn't shorter, because of std::abs i have to add a compare lamda

dash/include/dash/halo/Halo.h
dhinf: correct

dash/include/dash/halo/Halo.h
dhinf: The reason is, it isn't really fixed, because the user can change it any time.

dash/include/dash/halo/Halo.h
dhinf: when we use c++17 as default, i will change it :)

dash/include/dash/halo/Halo.h
dhinf: done

dash/include/dash/halo/Halo.h
dhinf: done

dash/include/dash/halo/Halo.h
dhinf: it was used in the for loop, now it is used in the function read_stencil_points

dash/include/dash/halo/Halo.h
dhinf: solved

dash/include/dash/halo/HaloMatrixWrapper.h
dhinf: done

dash/include/dash/halo/HaloMatrixWrapper.h
dhinf: done

dash/include/dash/halo/HaloMatrixWrapper.h
dhinf: done

dash/include/dash/halo/HaloMatrixWrapper.h
dhinf: its not a transfer. HaloStencilOperator uses the local memory and halo memory to fulfill diffenent operations like iterator or add values to the stencil points

fmoessbauer:
@dhinf The CI currently fails as there are build errors on clang 3.8.

**update**: Tested it locally with both clang-3.8.0 and 5.0.0 and it surprisingly worked.

devreal (approved):

dash/include/dash/halo/HaloMatrixWrapper.h
devreal: My comment was refering to `struct HaloData` ;)

dash/include/dash/halo/HaloStencilOperator.h
```diff
@@ -0,0 +1,267 @@
+#ifndef DASH__HALO_HALOSTENCILOPERATOR_H
+#define DASH__HALO_HALOSTENCILOPERATOR_H
+
+#include <dash/halo/iterator/HaloStencilIterator.h>
+
+namespace dash {
+/**
+ * The HAloStencilOperator provides stencil specific iterator and functions for
```
bertwesarg, line 8: typo: HAlo

dash/include/dash/halo/HaloStencilOperator.h
```diff
@@ -0,0 +1,267 @@
...
+    auto*           center = _local_memory;
+    pattern_index_t offset = 0;
+
+    if(MemoryArrange == ROW_MAJOR) {
+      offset = coords[0];
+      for(auto d = 1; d < NumDimensions; ++d)
+        offset = offset * _view_local.extent(d) + coords[d];
+    } else {
+      offset = coords[NumDimensions - 1];
+      for(auto d = NumDimensions - 2; d >= 0; --d)
```
bertwesarg, line 169: `d >= 0` is always true for `unsigned`. This issue comes up multiple times in this file.

dash/include/dash/halo/HaloStencilOperator.h
```diff
@@ -0,0 +1,267 @@
...
+  void set_value_at_inner_local(
+    const ElementCoords_t& coords, ElementT value, ElementT coefficient_center,
+    std::function<ElementT(const ElementT&, const ElementT&)> op =
+      [](const ElementT& lhs, const ElementT& rhs) { return rhs; }) {
+    auto*           center = _local_memory;
+    pattern_index_t offset = 0;
+
+    if(MemoryArrange == ROW_MAJOR) {
+      offset = coords[0];
+      for(auto d = 1; d < NumDimensions; ++d)
```
bertwesarg, line 165: This `auto` does maynot result in the same type as `NumStencilPoints`. This issue comes up multiple times in this file.

dash/include/dash/halo/iterator/HaloStencilIterator.h
```diff
@@ -0,0 +1,517 @@
...
+  using StencilP_t       = StencilPoint<NumDimensions>;
+  using ElementCoords_t  = std::array<pattern_index_t, NumDimensions>;
+  using StencilOffsets_t = std::array<signed_pattern_size_t, NumStencilPoints>;
+
+public:
+  /**
+   * Constructor
+   *
+   * \param haloblock \ref HaloBlock instance to use
+   * \param stencil_spec \ref StencilSpec to use
```
bertwesarg, line 55: `halomemory` parameter not documented.

dash/include/dash/halo/iterator/HaloStencilIterator.h
```diff
@@ -0,0 +1,517 @@
...
+    _stencil_offsets(stencil_offsets),
+    _local_memory((ElementT*) _haloblock.globmem().lbegin()),
+    _local_layout(_haloblock.pattern().local_memory_layout()), _idx(idx) {
+    if(Scope == StencilViewScope::INNER)
+      set_view_local(_haloblock.view_inner());
+
+    if(Scope == StencilViewScope::ALL)
+      set_view_local(_haloblock.view_inner_with_boundaries());
+
+    if(Scope == StencilViewScope::BOUNDARY)
```
bertwesarg, line 73: Does C++ guarantee, that only valid values for `Scope` can be passed. If not, that we should have a `static_assert()` here to guarantee it.

dash/include/dash/halo/iterator/HaloStencilIterator.h
```diff
@@ -0,0 +1,517 @@
...
+
+#include <dash/Types.h>
+
+#include <dash/halo/Halo.h>
+
+#include <vector>
+
+namespace dash {
+
+enum class StencilViewScope : std::uint8_t { INNER, BOUNDARY, ALL };
```
bertwesarg, line 12: Why the restriction to `std::uint8_t`? As this is only a template parameter, it should never occupy memory storage at runtime.

dhinf: Maybe a user wants to use it out of templates.

dash/include/dash/halo/iterator/HaloStencilIterator.h
dhinf: yes, it does (enum class)

PR #504, opened 2018-03-11
Bug cxx std intel

fmoessbauer:
- Fixes erroneous C|CXX_STANDARD usage
- Temporary manually adds `-std=...`, as CMake does not add these flags correctly for all compilers (this should be investigated further)
- Replaces `std::is_trivially_copyable` as intel does not support this


devreal (approved):
Looks good

PR #509, opened 2018-03-17
Fix dash::equal Impementation

fmoessbauer:
This issue is for tracking progress:

- [x] optimize current implementation
- [x] add unit tests
- [ ] fix for ranges which are not locally corresponding


fmoessbauer:
The PR fixes the current implementation in regards to the checked issues (see above).
Furthermore unit tests are added. The unit test for Overlapping / non-mapping ranges is currently disabled as this is not implemented.

There are way to much corner-cases for a quick fix of the issue. Hence I vote for merging this PR and keeping this as a known issue.

devreal (approved):

fmoessbauer:
@devreal : @rkowalewski will have a look at the remaining parts tomorrow morning.

fmoessbauer:
I will merge this PR today, as the expected fixes (for arbitrary ranges) are not ready yet. As this will probably require a rework anyways, we should do this in a new PR.

PR #512, opened 2018-03-17
Improvements to DASH installation process

### LINKED ISSUE #483: Installing `dash-test-mpi` conflicts with `/bin/dash` shell
The `dash-test-mpi` binary is installed into a `dash/test/mpi` subdirectory of the `bin` directory, though if one specifies `/bin` as this directory, it may overwrite/fail because of the existing `/bin/dash` shell. Having subdirectories in bin-directories is not very common any way. I think putting the binary into a `dash` subdirectory of `libexec` fits best.

devreal:
1) Do not install examples and test binaries (Fixes #483)
2) Do not install internal DART headers. 
3) Remove unused `CMakeList.txt` in `dash/examples` and `dash/test`
4) Cache MPI paths and libraries to fix issue with ninja build


dash/scripts/dash-test.sh
```diff
@@ -85,7 +85,7 @@ run_suite()
   # Number of failed tests in this run
   THIS_FAIL_COUNT=$(($NEW_FAIL_COUNT-$TOTAL_FAIL_COUNT))
   TOTAL_FAIL_COUNT=$NEW_FAIL_COUNT
-  if [ "$THIS_FAIL_COUNT" = "0" ]; then
+  if [ $TEST_RET -a "$THIS_FAIL_COUNT" = "0" ]; then
```
bertwesarg: While `-a` is POSIX, it is not portable to all shell implementations. In autoconf world we avoid it tough. I personally think that my eye are more trained to '&&': `if  [ $TEST_RET ] && [ "$THIS_FAIL_COUNT" = "0" ]; then`. Though now that `$TEST_RET` stands alone, it is also clear that you does not test what you expect. `TEST_RET` is always a non-empty string. And this is what you test for, i.e., it is always true. That should be `[ $TEST_RET -eq 0 ]`. And while you are at it, why not use an arithmetic test for the latter part too: `[ $THIS_FAIL_COUNT -eq 0 ]`.

bertwesarg (changes requested):

dash/scripts/dash-test.sh
bertwesarg: Also, when you catch the return code in line 79m you get not the one from the actual test but for the last command in pipeline, which is `grep -v '^#'`.

bertwesarg (changes requested):

dash/scripts/dash-test.sh
devreal: Thanks for looking at this and for the hints, they should be addressed in my recent changes.

bertwesarg (approved):

PR #514, opened 2018-03-20
Fix potential segfault in dart_get_handle

devreal:


fmoessbauer (approved):

BenjaProg:
Oh sorry just saw this now. Will make some tests with it!

BenjaProg:
right now i have so much segfaults i don't know where they come from... i need an idea on what's going on... then i will report

PR #515, opened 2018-03-20
Heterogeneous Memory Spaces

rkowalewski:
This pull request tracks progress about implementing heterogeneous memory spaces including the following:

- [x] Propose a Model for heterogeneous memory spaces as specified in C++17 Polymorphic Resources
- [x] Propose Memory Traits to specify capabilities on memory spaces.
- [x] Extend the model of polymorphic memory resources for global and local memory spaces
- [x] Simplify the Allocator Concept based on Allocation Policies
- [x] Provide Implementation for High Bandwidth Memory on Intel Xeon Phi KNL

- [ ] Provide CUDA Memory Space for integration with Alpakka, etc (delayed to a separate pull request)


fuchsto:
@rkowalewski We have to coordinate our work regarding this PR, let's have a meeting please.

CMakeLists.txt
```diff
@@ -376,6 +381,15 @@ if (ENABLE_HDF5)
 else()
   message(NOTE "HDF5                      disabled")
 endif()
+if (ENABLE_MEMKIND)
+  if (MEMKIND_FOUND)
+    message(INFO "MEMKIND                   enabled")
```
devreal, line 386: Should be `message(NOTE ...`

dash/examples/app.01.mindeg/main.cpp
```diff
@@ -13,6 +13,8 @@ using std::cout;
 using std::cin;
 using std::endl;
 
+using memalloc_ptr_t = decltype(dash::memalloc<int>(size_t{}));
```
devreal: This seems to be pattern now. Can we provide a wrapper to avoid the use of `decltype`?

dash/include/dash/Array.h
```diff
@@ -423,14 +420,15 @@ class AsyncArrayRef
...
+  class    PatternT,
+  typename MSpaceC>
 class ArrayRefView
 {
  public:
   typedef typename PatternT::index_type             index_type;
 
  private:
-  Array<T, index_type, PatternT>                  * _arr;
+  Array<T, index_type, PatternT, MSpaceC>                  * _arr;
```
devreal: Fix indentation

dash/include/dash/Array.h
```diff
@@ -944,35 +943,44 @@ class Array
...
-    deallocate();
+
+    std::swap(m_pattern, other.m_pattern);
+
+    // a) reset own data
+    m_data.reset();
+    // b) swap memory resources
+    std::swap(m_globmem, other.m_globmem);
+    // c) move the unique_gptr from other into this
+    unique_gptr_t __tmp{other.m_data.release(),
```
devreal, line 967: This also seems to be a pattern: can we define a `swap` that encapsulates this?

dash/include/dash/GlobPtr.h
```diff
@@ -154,93 +98,83 @@ class GlobPtr
...
+    : m_dart_pointer(other.m_dart_pointer)
+  {
+  }
 
   /**
    * Assignment operator.
    */
   template <typename T, class MemSpaceT>
-  self_t & operator=(const GlobPtr<T, MemSpaceT> & other)
+  self_t &operator=(const GlobPtr<T, MemSpaceT> &other)
```
devreal: Shouldnt't we check for const correctness and assignability here, too?

dash/include/dash/GlobPtr.h
```diff
@@ -319,11 +246,12 @@ class GlobPtr
...
    * Pointer increment operator.
    */
-  self_t & operator+=(index_type n)
+  self_t &operator+=(index_type n)
   {
     if (n >= 0) {
       increment(n);
-    } else {
+    }
+    else {
```
devreal, line 244: This change seems purely cosmetic (and frankly less attractive imho)

dash/include/dash/GlobRef.h
```diff
@@ -441,10 +421,19 @@ std::ostream & operator<<(
  * specialization for unqualified calls to swap
  */
 template<typename T>
-void swap(dash::GlobRef<T> && a, dash::GlobRef<T> && b){
+inline void swap(dash::GlobRef<T> && a, dash::GlobRef<T> && b){
   a.swap(b);
 }
 
+/**
+ * specialization for unqualified calls to swap
```
devreal, line 429: Fix documentation

dash/include/dash/Shared.h
```diff
@@ -182,9 +214,9 @@ class Shared {
   {
     DASH_LOG_DEBUG("Shared.cget()");
     DASH_LOG_DEBUG_VAR("Shared.cget", m_owner);
-    DASH_LOG_DEBUG_VAR("Shared.get", m_dart_gptr);
-    DASH_ASSERT(!DART_GPTR_ISNULL(m_dart_gptr));
-    return reference(m_dart_gptr);
+    DASH_LOG_DEBUG_VAR("Shared.get", m_glob_pointer);
+    DASH_ASSERT(static_cast<bool>(m_glob_pointer));
```
devreal, line 243: Is the static cast really needed? (It's missing in line 230)

dash/include/dash/Types.h
```diff
@@ -344,6 +347,21 @@ constexpr team_unit_t   UNDEFINED_TEAM_UNIT_ID{DART_UNDEFINED_UNIT_ID};
...
+typedef
+/* GCC Version < 4.9.0 */ \
+#if !defined(__clang__) && \
+    (defined(__GNUC__) && ((__GNUC__ == 4) && (__GNUC_MINOR__ < 9)))
+    struct {
+  long long __max_align_ll
+      __attribute__((__aligned__(__alignof__(long long))));
+  long double __max_align_ld
+      __attribute__((__aligned__(__alignof__(long double))));
+    }
```
devreal: Is this still needed since we moved to C++14? Also, the struct will be larger than 16 Byte (32B on 64bit systems, 8 for `__max_align_ll` + 16 for `__max_align_ld`)

dash/include/dash/algorithm/Find.h
```diff
@@ -50,17 +50,18 @@ GlobIter find(
...
-    const ElementType * l_range_begin = lbegin + l_begin_index;
-    const ElementType * l_range_end   = lbegin + l_end_index;
-
-    DASH_LOG_DEBUG("local index range", l_begin_index, l_end_index);
-
-    auto l_result = std::find(l_range_begin, l_range_end, value);
-    if(l_result == l_range_end){
-      DASH_LOG_DEBUG("Not found in local range");
-      g_index = std::numeric_limits<p_index_t>::max();
+   const auto* lbegin = static_cast<typename GlobIter::const_local_type>(
```
devreal: Is that `static_cast` really needed?

dash/include/dash/algorithm/MinMax.h
```diff
@@ -314,17 +327,16 @@ GlobInputIt min_element(
...
   /// Iterator to the initial position in the sequence
-  const GlobIter<ElementType, PatternType> & first,
+  const GlobIter & first,
   /// Iterator to the final position in the sequence
-  const GlobIter<ElementType, PatternType> & last,
+  const GlobIter & last,
   /// Element comparison function, defaults to std::less
   Compare                                    compare
-    = std::greater<const ElementType &>())
+    = Compare())
```
devreal: The formatting seems weird: put both on one line for better readability

dash/include/dash/algorithm/MinMax.h
```diff
@@ -357,7 +369,7 @@ const ElementType * max_element(
   const ElementType * last,
   /// Element comparison function, defaults to std::less
   Compare             compare
-    = std::greater<ElementType &>())
+    = Compare())
```
devreal: See above

dash/include/dash/allocator/AllocationPolicy.h
```diff
@@ -0,0 +1,422 @@
...
+#include <dash/internal/Logging.h>
+#include <dash/memory/MemorySpaceBase.h>
+
+std::ostream& operator<<(std::ostream& os, const dart_gptr_t& dartptr);
+
+namespace dash {
+
+enum class global_allocation_policy : uint8_t {
+  /// All units allocate the same amount of memory
+  collective,
```
devreal, line 19: There are cases where units allocate collectively but no the same amount of memory. Should that be a separate allocation policy?

dash/include/dash/allocator/AllocationPolicy.h
```diff
@@ -0,0 +1,422 @@
...
+ * independent of the other units. This local memory portion is allocated from
+ * a memory pool which is already attached to global memory. This variant is a
+ * good fit in cases where only one units needs to allocate memory while the
+ * other units do not contribute any local portion. It is used for example
+ * used to implement dash::Shared. Note: Both allocation and deallocation are
+ * non-collective. The user has to ensure that the owning unit does not
+ * release the memory while other units are still operating on it.
+ */
+template <class LocalAlloc, class LMemSpace>
+class LocalAllocationPolicy {
```
devreal: Out of curiosity: how would we fit the DART allocators in here? Would that be passed through the `LocalAlloc` argument passed to `do_global_allocate`?

dash/include/dash/allocator/internal/Types.h
```diff
@@ -0,0 +1,133 @@
...
+#include <cstddef>
+#include <dash/dart/if/dart_globmem.h>
+#include <ostream>
+#include <tuple>
+
+namespace dash {
+namespace allocator {
+
+template <typename LPtr, typename GPtr = dart_gptr_t>
+struct allocation_rec {
```
devreal, line 12: What is an `allocation_rec`? Allocation record? Please document

dash/include/dash/iterator/internal/GlobPtrBase.h
```diff
@@ -0,0 +1,366 @@
...
+}
+
+template <class T, class MemorySpaceT>
+dart_gptr_t decrement(
+    dart_gptr_t gptr,
+    typename MemorySpaceT::size_type offs,
+    MemorySpaceT const *mem_space,
+    memory_space_noncontiguous)
+{
+  DASH_THROW(dash::exception::NotImplemented, "not implemented yet");
```
devreal, line 286: Use `static_assert` here to avoid runtime surprises?

dash/src/memory/HostSpace.cc
```diff
@@ -0,0 +1,15 @@
+#include <dash/memory/HostSpace.h>
```
devreal, line 1: Out of curiosity: why are the memory spaces not implemented in the header? Is there some significant compilation overhead?

dash/src/memory/internal/MemorySpaceRegistry.cc
```diff
@@ -0,0 +1,72 @@
...
+  DASH_LOG_TRACE(
+      "MemorySpaceRegistry.erase",
+      "removing memory space",
+      std::get<IDX_TEAMID>(key),
+      std::get<IDX_SEGID>(key));
+
+  m_segments.erase(it);
+}
+
+std::vector<dash::internal::MemorySpaceRegistry::element_t>::iterator
```
devreal: Remove `dash::internal` from `dash::internal::MemorySpaceRegistry::element_t`

dash/include/dash/memory/internal/MemorySpaceRegistry.h
```diff
@@ -0,0 +1,60 @@
...
+    if (it != std::end(m_segments)) {
+      return it->second;
+    }
+
+    return nullptr;
+  }
+
+private:
+  static std::unique_ptr<MemorySpaceRegistry> m_instance;
+  static std::vector<element_t>               m_segments;
```
devreal, line 48: Why not use a ` std::unordered_map` and hash `segid_t` and `teamid_t` into a 32-bit integer? The linear search may otherwise become significant for larger numbers of segments, e.g., when using the dynamic data structures.

devreal (changes requested):
Thanks a lot for all the changes, definitely many good improvements in here! Some comments inline.

CMakeLists.txt
rkowalewski: hm...the other's have `INFO` as well if enabled.

dash/examples/app.01.mindeg/main.cpp
rkowalewski: We use `dash::memalloc` not too often. But yes I can give it a try. Even though I think it's not a bad pattern.

dash/include/dash/Shared.h
rkowalewski: Removed.

dash/include/dash/Types.h
rkowalewski: Good points. Let's remove it.

dash/include/dash/algorithm/Find.h
rkowalewski: Changed now.

dash/include/dash/algorithm/MinMax.h
rkowalewski: Yup, good point. I used `sed` sometimes, maybe that is the reason.

dash/include/dash/iterator/internal/GlobPtrBase.h
rkowalewski: In this case, not required actually. :)

dash/include/dash/Array.h
rkowalewski: The only way to do that is to implemented my own `dash::unique_ptr`. And that is worse compared to this IMHO. `std::unique_ptr` is limited somehow with it's API and this approach here is also the recommended from the `std::unique_ptr` [author](https://stackoverflow.com/questions/9167205/why-doesnt-unique-ptrreset-have-overloads-that-take-a-deleter/9167455#9167455).

dash/include/dash/GlobPtr.h
rkowalewski: This copy constructor is now deleted anyway. I will add it in future by considering the correct memory traits.

dash/src/memory/HostSpace.cc
rkowalewski: Because it does not rely on templates. ;)

dash/include/dash/memory/internal/MemorySpaceRegistry.h
rkowalewski: std::vector is way faster for a small number of entries than `unordered_map`. We can change it in future but for now I would stay with a `std::vector` because we have rarely more than 10 allocated global memory instances in a single DASH application.

CMakeLists.txt
devreal: Ahh you're right

dash/examples/app.01.mindeg/main.cpp
devreal: Something along the lines of `dash::memalloc_ptr_t<T>` should be easy to do

dash/include/dash/allocator/internal/Types.h
rkowalewski: You are right. It is actually only an internal type (internal namespace) and will be removed soon. It is used only in the EpochSynchronizedAllocator. And this will be changed to follow the new memory space concept in another pull request (near future). At the moment, only static containers are adapted to it. And this was a significant amount of work.

dash/src/memory/internal/MemorySpaceRegistry.cc
rkowalewski: Ok should be done now.

dash/examples/app.01.mindeg/main.cpp
rkowalewski: You convinced me, it is more readable. Done.

rkowalewski:
This pull request contains now a significantly refined implementation of memory spaces. We have now a clear separation of concerns between memory, allocators, containers and pointers. I will provide a more enhanced documentation in our internal wiki. Just to summarize:

- Passing around pointers in global memory: Pointer arithmetic, deferencing, etc. works correct on any global pointer on any unit, as defined by the memory traits.
- Memory has no `value_type` anymore as memory knows only about bytes (i.e. `std::byte`)
- Allocators use memory to allocate / deallocate objects according to the alignment requirements.
- Containers use now always RAII by wrapping the allocated memory segment in a `std::unique_ptr`.
- I removed lots of code, fixed pointer conversion and, `dash::distance` and pointer arithmetics in general.

rkowalewski:
As presented in the Stuttgart meeting, we can now allocate in any memory type. This pull request supports RAM and HBM (on Intel KNL).

CUDA and non-volatile (i.e. persistent memory) follow in a separate pull request. Just to limit the scope for this pull request here.

dash/examples/ex.04.memalloc/main.cpp
```diff
@@ -19,20 +19,22 @@ int main(int argc, char* argv[])
   auto myid = dash::myid();
   auto size = dash::size();
 
-  dash::Array< dash::GlobPtr<int> > arr(size);
+  using pointer_t = decltype(dash::memalloc<int>(size_t{}));
```
fmoessbauer: Can that be replaced with `dash::GlobMemAllocPtr<int>`? In an example this looks cleaner.

dash/include/dash/Matrix.h
```diff
@@ -95,23 +96,31 @@ template <
...
 template <
   typename T,
   dim_t NumDimensions,
   dim_t CUR,
-  class PatternT >
+  class Pattern,
+  typename LocalMemSpaceT>
 class LocalMatrixRef;
 
+namespace halo {
```
fmoessbauer, line 119: Why is this added to the `Matrix` class?

dash/include/dash/allocator/AllocationPolicy.h
fmoessbauer: @devreal: In my oppinion not as it is a different category, but do we test that?

dash/include/dash/coarray/CoEventIter.h
```diff
@@ -13,7 +13,12 @@ namespace coarray {
 class CoEventIter {
 private:
   using self_t = CoEventIter;
-  using gptr_t = GlobPtr<dash::Atomic<int>>;
+  //TODO rko: fix this hard coded information it is related to Coevent
+  //which uses dash::Array internally
+  using globmem_t =
```
fmoessbauer, line 18: Can't we use a `decltype` here?

dash/include/dash/coarray/CoEventRef.h
```diff
@@ -5,14 +5,21 @@
...
 namespace coarray {
 
 class CoEventRef {
 private:
   using self_t      = CoEventRef;
   using event_ctr_t = dash::Atomic<int>;
-  using gptr_t      = GlobPtr<event_ctr_t>;
+  //TODO rko: fix this hard coded information it is related to Coevent
+  //which uses dash::Array internally
+  using globmem_t =
```
fmoessbauer, line 19: Same here

dash/include/dash/internal/Math.h
```diff
@@ -3,52 +3,85 @@
...
 #include <numeric>
+#include <set>
+#include <utility>
+#include <climits> //For CHAR_BITS
 
 namespace dash {
 namespace math {
 
+template <unsigned Modulus, typename Integer>
+constexpr Integer RoundDown(Integer i)
```
fmoessbauer: Missing documentation, strange name.

fmoessbauer (commented):
There are many good things in this PR. However the amount of changes is massive. I'm not sure if a full review is actually possible. See my inline comments. After @devreal has approved, we can merge IMO.

dash/include/dash/allocator/AllocationPolicy.h
devreal: At the moment we do not test that but that is something I've been meaning to add for a while. MPI privides an info key that can steer some optimizations inside the MPI implementation if all units are guaranteed to allocate the same amount of memory but I'm not sure how big the influence is.

rkowalewski: > There are cases where units allocate collectively but no the same amount of memory. Should that be a separate allocation policy?

We test that for example with the CSR Pattern. For example in the `SortTest`, we show a use case where some units take part in the collective global allocation but contribute 0 "local" bytes.

rkowalewski: collective allocation is just a matter of synchronization. It does not say anything about the capacity. That's what MPI provides us.

dash/examples/ex.04.memalloc/main.cpp
rkowalewski: Must have overseen that. Done.

dash/include/dash/Matrix.h
rkowalewski: Because we can then forward declare `HaloMatrixWrapper` and make it a friend. Otherwise c++ does not allow that.

dash/include/dash/coarray/CoEventRef.h
rkowalewski: Hmm...not sure what you mean? The only way I see is to provide the pointer type as a template parameter to `CoEventIter`.

dash/include/dash/internal/Math.h
rkowalewski: Removed as it is not used anymore. ;)

rkowalewski:
Thank you for all your review comments.

dash/include/dash/allocator/AllocationPolicy.h
devreal: I see, I guess I am puzzled by the documentation `All units allocate the same amount of memory`. I guess what you mean is `All units collectively allocate memory`?

rkowalewski: Ah correct. Now I understand your question.

rkowalewski: Documentation updated.

devreal (approved):

devreal:
@rkowalewski Thanks for the great work!

PR #518, opened 2018-03-23
make sure dynamic windows are enabled if shared windows are not disabled

devreal:
For those building without CMake :)


fuerlinger (approved):
looks good, thanks!

rkowalewski (approved):

dart-if/include/dash/dart/if/dart_globmem.h
```diff
@@ -15,6 +15,12 @@
 #include <dash/dart/if/dart_util.h>
 #include <dash/dart/if/dart_types.h>
 
+// make sure dynamic windows are enabled if shared windows are not disabled
+#if !defined(DART_MPI_DISABLE_SHARED_WINDOWS) && \
+    !defined(DART_MPI_ENABLE_DYNAMIC_WINDOWS)
+#define DART_MPI_ENABLE_DYNAMIC_WINDOWS
```
fmoessbauer, line 21: Just out of curiosity: Why do we have `DISABLE` and `ENABLE` flags?

devreal: I think historically shared memory windows were the default so someone chose to *disable* them :(

PR #522, opened 2018-04-18
Use the CMake CXX Compiler in DASH Compiler Wrapper

rkowalewski:
The compiler wrapper does not use the correct c++ compiler. As an example, if I set `CC=clang` and `CXX=clang++` and execute CMake with these settings we use of course clang to compile our library. However, the CompilerWrapper is set to mpicxx (in case of MPI) which may be different from the underlying c++ compiler. This pull request addresses this issue.
Since I am not really a CMake expert reviews are welcome.


devreal (approved):
Good catch! Looks good to me :+1:

fmoessbauer (approved):

PR #523, opened 2018-04-18
Migrate CI to CircleCI v2, massively reducing CI time

fmoessbauer:
As CircleCI v1 is shut down in August, we have to migrate to v2.0. This brings the possibility to execute all targets / jobs in parallel, which reduces the overall CI time to ~8 minutes.

Another benefit is that the artifacts are now collected even if the job fails, hence it is possible to inspect the logs directly on Circle.


.circleci/config.yml
```diff
@@ -0,0 +1,73 @@
...
+          path: /tmp/build-logs
+      - store_test_results:
+          path: /tmp/build-tests
+
+  gnu_openmpi2:
+    docker:
+      - image: dashproject/ci:openmpi2
+    environment:
+      DASH_MAKE_PROCS: 2
+      DASH_MAX_UNITS: 3
```
devreal: Can we set `DASH_MAX_UNITS` to 4? My gut tells me that there might be a) test that require 4 units, and b) some problems don't trigger with less than 4 units. Both are just feelings, though.

dash/scripts/circleci/collect-artifacts.sh
```diff
@@ -0,0 +1,18 @@
...
+cd build-ci
+
+echo "Collect Logs"
+cp -v --parents -R ./*/*/*/*.log /tmp/build-logs/${CIRCLE_JOB}
+
+echo "Collect Tests"
+cp -v --parents -R ./*/*/*/dash-tests-*.xml /tmp/build-tests/${CIRCLE_JOB}
+
+echo "Artifacts copied to /tmp/build-logs/${CIRCLE_JOB}"
+echo "Testlogs copied to /tmp/build-tests/${CIRCLE_JOB}"
```
devreal: I suggest to move the output under the respective operation

devreal (changes requested):
Thanks for taking care of this :+1: 

Out of curiosity: does this branch already run the CircleCI2 now?

.circleci/config.yml
fmoessbauer: Let's try that. In circle v1 this was not possible as the tests run incredibly slow.

dash/scripts/circleci/collect-artifacts.sh
fmoessbauer: Yes, done!

fmoessbauer:
> does this branch already run the CircleCI2 now?
Yes, as you can see in the GitHub CI section where now 4 Circle targets are shown.

Btw: I added an additional openmpi3 container now. This was already on my list, but the container has not been marked for automatic builds on dockerhub. ...And building the container takes ~4 hours on there.

.circleci/config.yml
fmoessbauer: According to the artifacts, the topology stuff fails on CI. We should either disable this or run with just 3 units. Let's discuss this in the meeting.

fmoessbauer:
The Locality tests still do not run properly on CI, hence I disabled them. This is no regression compared to circle v1, as these tests are only active if running with at least 4 units.

After we have merged this PR, we have to re-run CI on all previous PRs as the merge-guards have changed as well.

rkowalewski:
Before merging we should fix and reenable the `TeamLocalityTest`. Otherwise it will sink into oblivion.

fmoessbauer:
@rkowalewski If @fuchsto is willing to do that, no problem. This is a internal problem in DART locality which does not support a mapping of multiple units to the same core. Even the `--map-by core` in CI seems to not ensure this.

Btw: In the old CI setup this testcase is not executed either, as the test is skipped if running with less than 4 units. So IMO the pragmatic way is to just merge this PR and enable the test after @fuchsto has fixed the bug.

devreal:
I agree with @fmoessbauer, as this is a bug in DART locality it should be handled separately. We cannot tie progress on so many fronts to broken implementations that won't be fixed any time soon...

rkowalewski (approved):
The changes look fine for me. I am still not lucky with skipped test cases however we should not block this improvement, I agree here.

PR #524, opened 2018-04-19
Feat halo

dhinf:
-added halo namespace for all halo stuff
- proxy classes for inner and boundary stencil operators
- view wise boundary iterators for boundary elements
- adapted tests


dash/include/dash/halo/HaloMatrixWrapper.h
```diff
@@ -267,8 +264,10 @@ class HaloMatrixWrapper {
    * Initiates a blocking halo region update for all halo elements.
    */
   void update() {
-    for(auto& region : _region_data)
-      update_halo_intern(region.second, false);
+    for(auto& region : _region_data) {
+      update_halo_intern(region.second);
+      dart_wait_local(&region.second.handle);
```
devreal: I suggest moving the wait into a separate loop to better overlap multiple transfers. If possible, using `dart_waitall_local` would be more efficient.

dash/include/dash/halo/HaloMatrixWrapper.h
```diff
@@ -277,16 +276,18 @@ class HaloMatrixWrapper {
...
+    }
   }
 
   /**
    * Initiates an asychronous halo region update for all halo elements.
    */
   void update_async() {
     for(auto& region : _region_data)
-      update_halo_intern(region.second, true);
+      update_halo_intern(region.second);
```
devreal: Please add braces

dash/include/dash/halo/HaloMatrixWrapper.h
```diff
@@ -296,7 +297,7 @@ class HaloMatrixWrapper {
   void update_async_at(region_index_t index) {
     auto it_find = _region_data.find(index);
     if(it_find != _region_data.end())
-      update_halo_intern(it_find->second, true);
+      update_halo_intern(it_find->second);
```
devreal: Please add braces

dash/include/dash/halo/Halo.h
```diff
@@ -55,18 +61,18 @@ class StencilPoint : public Dimensional<int16_t, NumDimensions> {
   constexpr StencilPoint(
     typename std::enable_if<sizeof...(Values) == NumDimensions - 1,
```
devreal, line 62: I suggest moving this check into a `static_assert` because it's not SFINAE.

dash/include/dash/halo/Halo.h
```diff
@@ -279,6 +405,19 @@ class RegionCoords : public Dimensional<uint8_t, NumDimensions> {
    */
   constexpr region_index_t index() const { return _index; }
 
+  static region_index_t index(dim_t dim, RegionPos pos) {
```
devreal: Documentation missing

dash/include/dash/halo/Halo.h
```diff
@@ -678,7 +837,7 @@ class RegionIter {
 
   GlobIter<ElementT, PatternT> global() const {
     auto g_idx = gpos();
-    return GlobIter<ElementT, PatternT>(&_globmem, _pattern, g_idx);
+    return GlobIter<ElementT, PatternT>(_globmem, &_pattern, g_idx);
```
devreal: `_pattern` is a pointer so `&_pattern` --> `*_pattern`

dash/include/dash/halo/Halo.h
```diff
@@ -1360,14 +1594,16 @@ class HaloMemory {
    * \return Pointer to the first halo element or nullptr if the
    *         region doesn't exist
    */
-  Element_t* pos_at(region_index_t index) { return _halo_offsets[index]; }
+  Element_t* first_element_at(region_index_t index) {
```
devreal: `first_element_at` returns a *pointer* to the first element :thinking: Why not have it return a reference?

dash/include/dash/halo/Halo.h
```diff
@@ -1360,14 +1594,16 @@ class HaloMemory {
...
+    return _halo_offsets[index];
+  }
 
   /**
    * Pointer to the first halo element
    *
    * \return Pointer to the first halo element
    */
-  Element_t* pos_begin() { return _halobuffer.data(); }
+  Element_t* first_element() { return _halobuffer.data(); }
```
devreal: See above

devreal (changes requested):
Some minor comments

dash/include/dash/halo/Halo.h
dhinf: look at the comment above, there is the answer

dash/include/dash/halo/Halo.h
dhinf: Normally you want a pointer, to iterate through the data. More interesting would be the return of an iterator. What you think?

dash/include/dash/halo/HaloMatrixWrapper.h
dhinf: I hate braces for one statement loops/if's . But for you, i will change it ;)

dash/include/dash/halo/Halo.h
dhinf: it's SFINAE. i have no better solution in the moment to avoid an ambiguous constructor.

devreal: Mhh, not sure I understand that. I see 2 ctor, both taking varargs, both with the same `std::enable_if`. Why not use a `static_assert` instead? What is the ambiguity?

devreal: Ahh, I think I now better understand what is going on. The constructors might be ambiguous if `CoeffT` has the same type as the values?

dhinf: it's ambiguous also in case of different value types. It's kind of strange. My first attempt was to declare the constructors explicit, to avoid implicit conversions. Didn't worked. 

Btw. the enable if's are not the same. The value of size_of is the same, but the total number of  arguments differs.

devreal (approved):

PR #525, opened 2018-04-19
include gtest using add_subdirectory instead of external project

fmoessbauer:
Advantages:
- faster build times
- more easy to maintain
- much cleaner approach

Do not pin gtest to a certain commit, but use head instead. Removed incompatible reference to gtest internal.


rkowalewski (approved):
Well done. Reviewing removed code is always a pleasure.

devreal:
Good riddance :+1:

devreal (approved):

PR #526, opened 2018-04-24
Disable Locality Tests with OpenMPI2

fmoessbauer:
…be ensured


rkowalewski (approved):
Looks good (unfortunately).

PR #527, opened 2018-04-27
Fix Clang Warnings and Rework Const Correctness of GlobRef

devreal:
Fix a bunch of warnings, most notably from the conversion between `GlobRef<T>` and `GlobRef<const T>`. The conversion operators are removed and references can now only be converted through implicit/explicit constructor calls, i.e., 

```C
GlobRef<int> gref = arr[0];
GlobRef<const int> cgref = gref; // implicit ctor, nonconst to const
GlobRef<int> gref2(cgref); // explicit ctor, const to nonconst
```


rkowalewski:
What I would suggest to remove as well are all the "erroneous" operators like `nonconst_value_type operator++(int);`, etc. This makes no sense and if the user wants to do that the user should do it explicitly. We may even argue about the right to exist of `operator++` on a `GlobRef` at all, because we implicitly have to apply two possibly very expensive communication operations. We should not hide such logic from the users. Moreover, the memory semantics are not well-defined if multiple units apply this operation concurrently to the same global memory address.

It is different in case of an `GlobAtomicRef` since the semantics are well defined here.

rkowalewski:
IMHO the logic in `GlobRef` is way too complex. It should be a lightweight object where users can simply read and write to. All the other operations may implemented in terms of these two "basic" operations explicitly. If the user needs well-defined atomic semantics he (or she) should use `GlobAtomicRef`.

CMakeExt/CompilerFlags.cmake
```diff
@@ -19,7 +19,7 @@ find_package(OpenMP)
 #  | -Weffc++                 | Spurious false positives                  |
 #  '--------------------------'-------------------------------------------'
 
-set (DART_C_STD_PREFERED "99")
+set (DART_C_STD_PREFERED "11")
```
fmoessbauer: As we only require a c99 capable compiler, this is not possible. We have to guard this similar to the implementation in PR #528 . Will fix it there.

fmoessbauer: Implemented here: https://github.com/dash-project/dash/pull/528/files#diff-0fa6dec18f5c6ec2e55c3b6de6f3ef62R22

While the flag is called `...PREFERED`, this denotes the minimum required version. This is due to how CMake handles this in `set_property(TARGET tgt PROPERTY C_STANDARD 11)`. To actually ensure that the compiler supports this, we add `C_STANDARD_REQUIRED`, which is a boolen telling that it's not only preferred but required.
https://cmake.org/cmake/help/v3.1/prop_tgt/C_STANDARD.html

Anyways: I also stumbled upon this...

devreal:
> What I would suggest to remove as well are all the "erroneous" operators like `nonconst_value_type operator++(int);`

Why is it erroneous? It increments the value at the memory location and returns the previous value *by value*, just as `operator++(int)` is supposed to. 

> IMHO the logic in GlobRef is way too complex. It should be a lightweight object where users can simply read and write to. 

It is still lightweight, only carries a `dart_gptr_t. Not sure what you mean there.

> All the other operations may implemented in terms of these two "basic" operations explicitly. 

They are already implemented in terms of put and get, just in the context of the GlobRef. I don't see any benefit of removing them. If you care about performance you won't do element-wise blocking increment anyway, so this is just for convenience. And we simply mimic the behavior of native references: 

```C
int val = 0;
++val; // not thread-safe, requires a load and a store. thread-parallel increment is undefined
```

> If the user needs well-defined atomic semantics he (or she) should use GlobAtomicRef.

And if he doesn't need atomic semantics (because he knows what he does) he can use `GlobRef` and doesn't have to implement the operations himself.

CMakeExt/CompilerFlags.cmake
devreal: I actually don't like the fact that we have preferred and required language versions. What is the benefit? If we want to be portable we cannot use C++14 features anyway as long as the required standard is C++11  (unless you clutter you code with `ifdef`s). 

In the case of C99, we stick to an almost decade-old standard. It's not that C11 has much news to offer, some features are handy though (anonymous struct members, for example). Any reason for not bumping the required version to C11?

rkowalewski:
> Why is it erroneous? It increments the value at the memory location and returns the previous value by value, just as operator++(int) is supposed to.

The erroneous thing is that we define `operator++` for a type (i.e., `GlobRef`) which does not operate on the type itself. This is not what happens in native C++ since references are only a opaque compiler guarantee which does not exist at runtime.
In the case of `GlobRef` we define operations (++,--, etc.) which do semantically not make sense on the `GlobRef` itself. And as already mentioned above, we "encourage" the user not only to write code with bad performance but also to write semantically invalid code since it may result in inconsistent memory.

We provide all these things just for convenience and I am not sure if that is a selling point to neglect HPC aspects in a HPC library.

devreal:
Our `GlobRef` is supposed to act like a native reference in C++: it points to a value (instead of holding the value itself), cannot be NULL, and provides operations on the value, e.g., :

```C
int var = 0;
int& a = var;
++a; // increment var to 1
```

The founding fathers of C++ laid down the rule that an operation on a reference does not increment the reference but applies to the value it points to. We define similar behavior for our references (which have to be explicit objects because we cannot change the language, of course).

> And as already mentioned above, we "encourage" the user not only to write code with bad performance

I'm happy to have a discussion about the memory model itself and whether we want to have *synchronous* memory access by default (hint: we don't if we care about performance). I will not argue about this here, I have made my attempts previously. All I am saying is that the operations are for convenience, they won't get faster if the user implements them himself, and they should not be used in performance critical code. Users should be aware that there is a difference in performance depending on whether you act on local or global memory. That won't change if we remove the operations on `GlobRef`. Maybe we need more documentation to clearly state this...

> ... but also to write semantically invalid code since it may result in inconsistent memory.

Would you argue to prohibit increment on `int&` because it's UB if the user calls it on the same reference from multiple threads? It needs a certain awareness to program in parallel, and we won't change that.

rkowalewski:
OK your statements and detailed explanations convinced me.

dash/include/dash/atomic/GlobAtomicAsyncRef.h
```diff
@@ -81,6 +82,28 @@ class GlobAsyncRef<dash::Atomic<T>>
...
   }
 
+
+  template <class...> struct null_v : std::integral_constant<int, 0> {};
+
+  GlobAsyncRef(const GlobAsyncRef<dash::Atomic<nonconst_value_type>>& gref)
+  : GlobAsyncRef(gref.dart_gptr())
+  { }
+
+//#if 0
```
rkowalewski: If this code is not valid remove it

rkowalewski (changes requested):
One comment inline. The other things are looking good.

dhinf:
seemed the right place to fix these unused warnings

rkowalewski (approved):
Please resolve the merge conflicts so we can merge it

rkowalewski:
While experimenting with all these things there are still some conceptual odds with const -> nonconst conversion and the other way around. I will provide some details later. So it is not ready to merge.

fmoessbauer:
I resolved the conflicts by setting C_STD_PREFERED to 99 as discussed in #528. If the compiler supports c11, it is used. Furthermore I fixed some naming issues, e.g. ` DASH_C_STD` vs ` DART_C_STD`.

dhinf (approved):

devreal:
@rkowalewski any update on the issues you mentioned?

rkowalewski:
@devreal Yes, sorry it took a while. Just finished it. Hopefully all tests are still running.

To summarize my changes:
- I moved the SFINAE checks for implicit vs. explicit constructos to a new header called `GlobRefBase` to remove some cyclic dependencies between headers. Moreover I removed all includes which are not really necessary
- const <-> nonconst conversion got some minor changes. I have removed duplicate constructors which are not necessary at all. All cases are documented in both `GlobRef` and `GlobAsyncRef`. Please have another review on it.
- Ownership of the `dart_handle_t` is now managed by a `std::unique_ptr` in `GlobAsyncRef`.
- I have moved some constructors from public to private. The reason is that I am not sure if all these constructors are really well defined, especially the few where we pass a `GlobPtr` as the single constructor argument. One can really misuse these constructors , for example if one constructs a `GlobRef<T>` by passing a  `GlobPtr<U>`. This needs some more thoughts what is really well defined and what not. It also interferes with some changes in #515.

EDIT: @dhinf I made a small minor change in the [Halo](https://github.com/dash-project/dash/pull/527/commits/fc870c2b3e2004446a47f71ac02c0aeca89e0709#diff-1c5445c8cdb55be95c7dfdba093d3480R833) stuff. Tests are still running, please double check it. The dead code is removed in another commit.

fmoessbauer (approved):
There are various good fixes in here. Thanks for taking care of that. IMO we can merge this PR and after merging try to review #536

dash/include/dash/GlobRef.h
```diff
@@ -72,121 +34,157 @@ class GlobRef
...
-  : GlobRef(gref.dart_gptr())
-  { }
-//#endif
+  template <class _T, class _Pat, class _M, class _Ptr, class _Ref>
+  friend class GlobViewIter;
 
+private:
   /**
-   * Constructor, creates an GlobRef object referencing an element in global
+   * PRIVATE: Constructor, creates an GlobRef object referencing an element in global
```
devreal, line 48: Are those private constructors used anywhere?

rkowalewski: Yes they are. That is the reason why I introduced class friendship (which makes sense anyway in these cases).

rkowalewski: In the ideal case a `GlobRef` should never be constructed by the users themselves. Only const <-> nonconst conversion should be supported. Hence, all Constructors excluding the conversions should be private.
However, this requires additional conceptual rework which would again cause a lot of changes in this pull request.

rkowalewski:
I have done the same changes to the `Atomic` specializations both on `GlobRef` and `GlobAsyncRef` and everything should be uniform now. While looking through the code I noted that `GlobAsyncRef<Atomic<T>>` does not work  with handles as `GlobAsyncRef<T>`. However, this may be another pull request.

devreal:
> I have done the same changes to the Atomic specializations both on GlobRef and GlobAsyncRef and everything should be uniform now. While looking through the code I noted that GlobAsyncRef<Atomic<T>> does not work with handles as GlobAsyncRef<T>. However, this may be another pull request.

Yes, I am planning to add a mechanism to signal instant local or global completion to put/get and atomics and get rid of the `*_blocking` variants. That way we can get rid of the handle inside `GlobAsyncRef`. I haven't had the time to complete that though.

PR #528, opened 2018-04-27
Test CI build also with c++11 (not 14)

fmoessbauer:
As we have decided to require only a c++11 capable compiler, this is now also checked in CI.

This PR also fixes a collision of the `-Werr` flag, infected by gtest and `-Weffc++` set by us.

**Todo**: The CI currently fails, as the Halo stuff is not c++11 compatible.


devreal:
As discussed offline, we should not have two competing standards to support. So either C++11 or C++14, both local and in CI. @rkowalewski argues for sticking for C++11 and discuss the move to C++14 at the next F2F.

Some remarks:
By moving to C++14, we might lose support for older compilers but we would gain some interesting features (most notably extended `constexpr`). From what I can see, the following compiler versions sufficiently support C++14: 

- Intel 17 (https://software.intel.com/en-us/articles/c14-features-supported-by-intel-c-compiler)
- GCC 5 (https://gcc.gnu.org/projects/cxx-status.html)
- We might lose the Cray C++ compiler for a while but that's fine with me :)

fmoessbauer:
@devreal , @rkowalewski : Given your statement above, I vote for merging this PR and doing the cleanup after the final decision.

rkowalewski:
I agree with @fmoessbauer. Regarding the upgrade to C++14 we should evaluate our most relevant platforms and compilers. DASH should work at least on the following HPC centers:

- SuperMUC
- HLRS
- NERSC Cori / Edison
- ORNL Summit (if we manage to get access)

Moreover I was able to build DASH with the Cray Compiler a couple of months ago. Before dropping support for Cray we should evaluate some benchmarks if it does not have a negative impact on performance.

devreal:
OK, let's evaluate that in the near future. I would, however, still argue that we disable support for C++14 entirely as the current setup only leads to confusion.

dhinf:
I vote for c++14. The view stuff of Tobias works  with c++14. I see no point to stick with c++11. It's the ugly brother/sister of C++14.

devreal:
I think we all agree that we should switch to C++14 (rather sooner than later) so let's investigate that soon. In the meantime, we need a consistent build system though.

rkowalewski:
I am not against C++14, I vote for that as well. However, we should not make such a change in a cloak-and-dagger operation just because it feels more comfortable and all developers build locally with leading edge compilers. We were C++11 compatible until now and before upgrading our compilation requirements we should (as already stated) investigate possible consequences. We are developing for HPC systems including the unloved legacy Cray Compiler.

dhinf:
So we use c++11 for the release and c++14 in the development branch? Works for me.

dhinf (dismissed):

fmoessbauer:
So in the meantime we need C++11 in CI. Do we agree to merge this PR as-is?

devreal:
@dhinf I think the decision on moving to C11 should be made together with moving to C++14.

dhinf:
you want to change to C11 in the PR #527. That's why i asked

devreal:
That's mostly because the compiler warns about anonymous struct members not being available in C99. However, all compilers we support happily suck up their pride and compile just fine :)

dhinf (approved):

PR #536, opened 2018-05-06
Integrate Clang Tidy into DASH

rkowalewski:
It is on my list for a while now and finally I have found at least some time. I gave [clang-tidy](https://clang.llvm.org/extra/clang-tidy/), which is basically a static analyzer to support you in following best practices of modern C++, a try on our DASH code base.
For the first run, I applied it only to our **algorithms** along with their **tests**. The number of fixes are quite limited but very interesting. Note that I have merged the changes of #527 since it removes a lot of warnings for clang-tidy as well and will be upstream soon.

Platform Requirements: Clang installation and [Clang Extra Tools](https://clang.llvm.org/extra/index.html). I have built both from source from the most recent major release (6.0). If everything worked as expected you should have clang-tidy in `<INSTALL_PREFIX>/bin` and there should be also a directory `<INSTALL_PREFIX>/share/clang/` with some scripts included. The relevant one here is `run-clang-tidy.py`. 

The workflow I have followed is straight-forward and very easy to follow.

NOTE: I have run all commands from the DASH source tree root directory.

1.  run the following command
~~~bash
/opt/llvm/6.0/share/clang/run-clang-tidy.py -p ./  \
-checks='*,-llvm-header-guard,-fuchsia-default-arguments,-fuchsia-overloaded-operator,-cppcoreguidelines-pro-bounds-pointer-arithmetic,-modernize-use-using,-google-readability-todo' \
-header-filter='.*algorithm.*' \
dash/test/algorithm \
-fix
~~~
Note The `-p` parameter. It tells clang-tidy that it should look for a compilation database (`compile_commands.json`) in the current directory. Simply copy your compilation database to the DASH root.
It follows the `-checks` parameter. More explanations are below. Then we apply a header filter (regex) to include only our algorithm header files.  We subsequently specify the path containing our C++ compilation units of algorithm tests to which clang-tidy is applied to. Finally, with `-fix` we tell the script to automatically apply suggested fixes.

2. Recompile everything to ensure that we did not introduce some syntax errors. This happened in my case, however they were very easy to fix-

3. Run unit tests to validate if no semantic errors have been introduced. This step is optional. In my case everything still worked smoothly.

4. Since the resulting code of clang-tidy is poorly formatted we have to reformat it. However, applying `clang-format` to the affected files will reformat all lines in these files which is definitely not what we want. I have found a very smart script which is included in Clang's extra tools source tree: [git-clang-format](https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/git-clang-format). It allows to reformat only changed lines in affected files. In order to do so follow these steps:

    1. Stage (`git add`) all your modified files from steps 1-3. This will ensure that you can always reset to the state of step 3 if something didn't work here as expected. Moreover, `git-clang-format` works only on staged files anyway.
    2. Run `git-clang-format --style file` in the DASH Root. This may take a while depending on the number of staged files. NOTE: This requires a .clang-format in the root directory or somewhere above it. `clang-format` will then take care of everything else. I have added my `.clang-format` file in this commit for a first start. It has just some minor adaptions from the Google Style to fit well to our current code base.
    3. Now you should see some additional modifications if you run `git status`. You can review everything with `git diff`.

5. If everything worked until here you have two possibilites:
    1. Commit first your staged files as affected by clang-tidy. Then you can stage and commit your files again as affected by `git-clang-format`. This is the approach I have followed here.
    2. First stage your changes as affected by `git-clang-format` as well and "merge" everything in a single commit.

6. Be lucky with your code which follows now best practices quite a bit more.

Some additional notes to the enabled checks.

First I enable all checks with `*`. Then I exclude some checks (Note the `-` sign) which no do not make sense in our case. Especially the fuchsia-checks. Check their documentation for more details. The other excluded checks:

- `-llvm-header-guard`: Clang-Tidy renames header guards according to Google / Clang standards. This is nothing critical and only causes lots of code changes.
- `-cppcoreguidelines-pro-bounds-pointer-arithmetic`: It follows the Cpp core guidelines which discourage pointer arithmetic. This leads to a lot of warnings in our code base since we use it excessively. And I think it is not critical anyway.
-`modernize-use-using`: Replace `typedefs` with `using`. It is more a matter of taste and it would result in a lot of changes in our code base. For this reason I disabled it.
- `-google-readability-todo`: It replaces `TODO`s with `TODO rkowalewski` for example. This is not what we want here. This may be a good thing for your own files.

In general clang-tidy has a lot of nice capabilities which are interesting for us. Especially const correctness, const expressions, etc. All these things may be found out with static analysis and it is a great support in developing correct library code. I am integrating it now in my VIM environment as part of my basic workflow. 

If you apply `clang-tidy` without the `-fix` switch you will see still a lot of warnings for the DASH code. Some are worth to consider, some are not. I am sure that especially with our containers (e.g. rule of five, const correctness, constexpression, etc.) we can do a lot of very good rework as more correct code leads usually to more performance which is of course the most important in our case.

This pull request here may serve as a discussion platform and experiences. I would be glad if anyone tries it out as well and maybe we can integrate it as a step in our DASH CI workflow in the near future.


fmoessbauer (approved):
Thanks for taking care of this. I skimmed over all the changes and there are lots of improvements in it. IMO we can / should merge this PR as is. Just one minor thing: please remove the .syntastic files.

rkowalewski:
Done. If the build is successful we can merge it.

Issue #532, opened 2018-05-02
How to set the diagonal of a matrix (and then some)


devreal:
Consider something like the following:

```C
dash::TeamSpec<2> ts(dash::size(), 1);
ts.balance_extents();
dash::Matrix<int, 2, dash::default_index_t, dash::TilePattern<2>> mat(20, 20, TILE(4), TILE(4), ts);

// zero the matrix
dash::fill(mat.begin(), mat.end(), 0.0);

// set the diagonal
for (size_t i = 0; i < mat.extent(0); ++i) {
  if (mat(i, i).is_local()) mat(i, i) = dash::myid();
}
```

I know what you're saying, that's trivial...

Here is what I get as a result:
```
    0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0 
    0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0 
    0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0 
    0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0 
    0     0     0     0     3     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0 
    0     0     0     0     0     3     0     0     0     0     0     0     0     0     0     0     0     0     0     0 
    0     0     0     0     0     0     3     0     0     0     0     0     0     0     0     0     0     0     0     0 
    0     0     0     0     0     0     0     3     0     0     0     0     0     0     0     0     0     0     0     0 
    0     0     0     0     1     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0 
    0     0     0     0     0     1     0     0     0     0     0     0     0     0     0     0     0     0     0     0 
    0     0     0     0     0     0     1     0     0     0     0     0     0     0     0     0     0     0     0     0 
    0     0     0     0     0     0     0     1     0     0     0     0     0     0     0     0     0     0     0     0 
    0     0     0     0     0     0     0     0     0     0     0     0     3     0     0     0     0     0     0     0 
    0     0     0     0     0     0     0     0     0     0     0     0     0     3     0     0     0     0     0     0 
    0     0     0     0     0     0     0     0     0     0     0     0     0     0     3     0     0     0     0     0 
    0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     3     0     0     0     0 
[...]

PR #538, opened 2018-05-07
Fix index calculations in GlobStaticMem

fmoessbauer:
@ddiefenthaler, @devreal  This PR adds a unit test for issue #532

### DIFF
--- a/dash/include/dash/memory/GlobStaticMem.h
+++ b/dash/include/dash/memory/GlobStaticMem.h
@@ -342,6 +342,7 @@ public:
 
   constexpr size_type local_size(dash::team_unit_t) const noexcept
   {
+    // TODO this only holds if all units allocate the same number of elements
     return _nlelem;
   }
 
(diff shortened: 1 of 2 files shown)


devreal:
Thanks! Should we use this PR to fix the issue? Otherwise we'd break CI everywhere else...

fmoessbauer:
@devreal: just to fix the issue and when fixed, we can merge it. @ddiefenthaler told me yesterday, that he will have a look at this today.

devreal:
Great! @ddiefenthaler please let us know if you need help with this!

fmoessbauer:
At first: @fuchsto and @rkowalewski  own me a beer!

While the fix is quite minimal, it revealed some conceptual problems as described in issue #539.

dash/include/dash/memory/GlobStaticMem.h
```diff
@@ -523,9 +524,10 @@ class GlobStaticMem
     DASH_LOG_TRACE_VAR("GlobStaticMem.at", lunit);
     // Apply global unit to global pointer:
     dart_gptr_setunit(&gptr, lunit);
+    // increment locally only
+    gptr.addr_or_offs.offset += local_index * sizeof(value_type);
     // Apply local offset to global pointer:
     pointer res_gptr(*this, gptr);
-    res_gptr += local_index;
```
devreal, line 528: Why does this not work? Shouldn't it be the same as incrementing the offset if the `dart_gptr_t` by `local_index * sizeof(value_type)`?

fmoessbauer: The difference is that the dart version only increments locally while the dash version first checks if the increment is past the local address range. As the local size of the memory on unit 1 is 96 (vs 144, unit 0) the dash version belives that a increment of 128 is a pointer to index 32 on the next unit. The actual bug are the capacity issues as stated in the linked issue.

devreal: Ahh, gotcha! I agree that it seems more efficient to operate on the `dart_gptr_t` directly, anyway. Thanks for fixing this!

devreal (approved):

PR #541, opened 2018-05-10
Use deterministic rngs in debug builds to avoid unwanted coverage

fmoessbauer:
changes

### DIFF
--- a/dash/test/algorithm/SortTest.cc
+++ b/dash/test/algorithm/SortTest.cc
@@ -11,6 +11,18 @@
 #include <cmath>
 #include <random>
 
+#ifndef DEBUG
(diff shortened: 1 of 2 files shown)


ddiefenthaler (approved):
Looks good.

It should be noted that this change only works as long as the coverage scripts stay at the debug-build mode.
Link to the current coverage build script: [build.cov.sh](https://github.com/dash-project/dash/blob/59f6d1cbeac35833f096cb9625e7cc03d0ab2037/build.cov.sh#L57)

fmoessbauer:
I added a slightly improved version do avoid cluttering the code with `#ifdef`s. @ddiefenthaler : Coverage builds are always debug builds ;)

PR #543, opened 2018-05-10
Fix Integer Overflow and Min / Max Reduction in dash::sort

fmoessbauer:


rkowalewski:
Thank you @fmoessbauer. Will fix it over the next days.

ddiefenthaler:
While you are at it, I guess there is an other integer overflow that occurs on 32bit machines.

I get this error if I change to an Array of type `long long int` in the `ExtremValues` test:
```
Expected equality of these values:
  _e
    Which is: -8589934472
  _a
    Which is: 4294967416
```

My bet is that the use of `size_t` is the problem here as it is smaller than `long long int` on 32bit machines.

rkowalewski:
Hm...yes I noticed it yesterday as well since not all tests have successfully passed. Something strange is still going on there. I will consider and test your assumption. :)

rkowalewski:
I think I have found the issue. It is another integer overflow, however, it is not what you first thought @ddiefenthaler. I do not use `size_t` for some arithmetic on the values of the array. It is just to index (i.e., count) the partitions. However, still interesting to know that `size_t` is smaller than `long long int` on 32-bit platforms. ;)

rkowalewski:
After a extensive debugging session I have finally fixed this issue. The root cause for the second integer overflow is #544. There is a temporary workaround, see [this commit](https://github.com/dash-project/dash/pull/543/commits/b1501f7c00be104156982641e863947fe54442a8). Thanks to all contributors in this pull request.

rkowalewski:
@fmoessbauer @ddiefenthaler I have removed your timeout as it does only slow down the unit tests.

dash/include/dash/algorithm/Sort.h
```diff
@@ -845,8 +843,36 @@ void sort(
   array_t g_partition_data(nunits * nunits * 3, dash::BLOCKED, team);
   std::fill(g_partition_data.lbegin(), g_partition_data.lend(), 0);
 
-  auto const min = static_cast<mapped_type>(g_min.get());
-  auto const max = static_cast<mapped_type>(g_max.get());
```
ddiefenthaler, line 849: I think there is a `barrier` missing before the call of this.
I can't tell if this is the cause of the issue here, but this is my assumption.

rkowalewski: This shouldn't be a problem as an implicit barrier is in the array constructors, so we can save a barrier here.

rkowalewski:
#544 was actually my fault, as described there. Before we close and merge this pull request we should review the suggested fix there. Then I can merge it into this branch and revert to the original implementation.

dash/include/dash/algorithm/Sort.h
```diff
@@ -830,24 +834,40 @@ void sort(
...
   using array_t = dash::Array<std::size_t>;
 
   std::size_t gsize = nunits * NLT_NLE_BLOCK * 2;
 
+  // implicit barrier...
   array_t g_nlt_nle(gsize, dash::BLOCKCYCLIC(NLT_NLE_BLOCK), team);
 
+  if (myid != owner) {
+    // the other units apply min / max reductions
+    g_min.get().op(dash::min<mapped_type>(), lmin);
```
devreal: Are we sure this is actually more efficient than using `allreduce`?  `allreduce` should be somewhere in *O(logn)* (see Rabenseifner's algorithm for example) and is usually a well optimized MPI functionality while what we're doing here is *O(n)*.

rkowalewski: Interesting thought. You may be right, however, I would argue that the O(logn) vs O(n) only applies in the very large scale (> 1000 nodes) which is also what previous benchmark results confirm. On the other hand, the one-sided counterpart causes less synchronization, i.e. `allreduce` implies synchronization in a barrier-like manner. So it is more efficient for a more or less low node count while the collective two-sided counterparts come into play with a higher node count.
In general, this consideration should be examined in more detail. There are several places in this algorithm where two-sided may be preferred.

devreal: > I would argue that the O(logn) vs O(n) only applies in the very large scale (> 1000 nodes) which is also what previous benchmark results confirm.

There might be a higher constant overhead involved for smaller numbers of units, although I would expect it to become significant way earlier than 1000 nodes. The problem with atomics is that only few implementations really support them in hardware. My measurements indicate that only Open MPI on IB networks actually does asynchronous progress. Cray is still struggling with it (according to one of their developers) and I haven't seen positive results from any other MPICH-based implementation (namely Intel MPI). For atomics implemented in software, this results in linear scaling of two-sided messages (to send the request for the atomic operation and receive the results) and work on the target. I'd be interested to see the sweet-spot in some measurements though.

> On the other hand, the one-sided counterpart causes less synchronization, i.e. allreduce implies synchronization in a barrier-like manner. So it is more efficient for a more or less low node count while the collective two-sided counterparts come into play with a higher node count.

Looking at the code, the benefit of one-sided communication seems completely voided by the immediately preceding and succeeding creations of global arrays, which are synchronizing operations. The processes are in lock-step anyway. Plus, the setup of a `dash::Shared` involves a collective operation as well...

Having said that, this should not impact the fixing of the bug this PR addresses. However, I think we need to carefully assess where collective operations make sense (and I think there is definitely room for improvements in other DASH algorithms as well).

devreal (approved):

dash/include/dash/algorithm/Sort.h
devreal: Just one more thought on this: Using `dash::Shared` to mimic reduce/allreduce bogs me everytime I see it (and I remember seeing it in the cowichan problems as well). We should stop doing that. The number of operations scales linearly, which quickly forms a bottleneck, and it involves multiple barriers (1 for creation + 1 to synchronize writing, +1 to synchronize reading/destruction in the case of `allreduce`). 

I would like to declare this an anti-pattern :stop_sign: Everytime you think "dash::Shared", you stop doing whatever it is you're doing, grab a coffee, and then use proper reductions where possible.

Instead we should think about introducing reductions at the DASH level (that encapsulate `dart_reduce`/`dart_allreduce`) to properly expose this functionality to the user. I had this on my list for the last F2F but we ran out of time before we could discuss this.

Just my 2 cents ;)

rkowalewski: It makes sense what you are saying and I definitely agree with that. However, since we can only speculate, lets do some benchmarks. I can write a small benchmark over the next weekend and run that for various unit / node counts, both shared and distributed memory.
I claim that achievable performance depends on two parameters. First, the number of units and nodes. The more cores involved the greater the benefit of two-sided MPI. The second parameter what may be interesting is density. As an example `dash::sort` may be called only on a specific range of the container. Additionally, we may have "holes" in our range in such a way that we have units which have an empty local range. This may be interesting for sparse data structures. Depending on the density, i.e. the ration of contributing units (non-empty local portion which is part of the range to be sorted) and the number of units in my team, one-sided communication may be more efficient. Let us say the ratio is 50-60% and we have the advantage in DASH that we know this at runtime, we may switch between collective or one-sided communication.

rkowalewski:
OK I have switched to the two-sided Min / Max Reduction now as it "should" be more efficient. However, we still should measure that. :)

fmoessbauer:
@rkowalewski : Please rename the PR according to the fix and after that we can merge.

Issue #479, opened 2018-01-12
Container: Default initialization of elements


devreal:
Consider the following code snippet:

```
  struct point_t {
    int x, y = 1;
  };

  std::vector<point_t> vec(N);
  assert(vec[0].y == 1); // succeeds

  dash::Array<point_t> arr(N);
  assert(vec[0].y == 1); // may or may not succeed
```

Point to make: Contrary to the STL, elements in DASH container are not initialized with their default value. Now, there is a good argument for this: performance! Many times users might want to skip implicit initialization because they perform the initialization themselves, e.g., using `dash::fill`. 

However, since we aim to be compatible with STL behavior and because I tapped into this trap myself (see https://github.com/dash-project/dash/pull/440/commits/b67c06572452f398eab8efe32cc4f8a3f58d9af1) I decided to raise this issue. What's more, this is mostly only relevant for derived types in DASH container because it is all too tempting to rely on the default values specified in the type definition. 

As far as I can see this behavior is not documented anywhere.

Should we change the behavior to provide default initialization or make it optional (one way or another)? Or just document this deviation from the STL behavior?

BenjaProg:
Hey here some input from my side:
Wouldn't it be failsafer for new developer if it behaves like in STL?
I suggest offering a constructor without initialization for performance aware programmers.
Is this a option? Or am I fooling myself anywhere?

PR #545, opened 2018-05-14
Added default initialization to dash::Shared (with unit tests)

rkowalewski:


devreal:
Thanks for the test case, I'll look at it later today.

ddiefenthaler:
How are atomics initalized? As the values are signed, zero is not the absolute minimum.
If the start value is negative the minimum will be not the expected value.
From looking at the failed CI runs all compared values are negative.

I can't reproduce this locally. I guess in my machine the atomics are allocated in a zeroed memory segment. I think that is why I can't reproduce it.

rkowalewski:
@ddiefenthaler you can either SSH into the CI machine or, like in my case, if you repeatedly execute it you will hopefully come into a situation which manifests the error.

ddiefenthaler:
I added additional expects, which would be required to perform the rest of the test without problems.
The results support my assumption. The wrong value is already stored in the atomic _before_ the `min`&#8209;operation is called on it. I can't see an error in the rest of the unit test.

For the related issue in PR #543 I think there is missing a `barrier` which leads to fetching the minimum or maximum before it is fully computed.

Maybe there is still another problem, but I can neither see it in CI nor on my local machine.

rkowalewski:
Ah now I got what you mean. Sorry, I didn't really know what I am doing here. Of course, we do not initialize the atomic value here, so it is undefined behavior. Stupid me. Actually, this refers to the same issue as in #479. This is something I am gonna change anyway in #515 , default initialization should be done for all containers. However, we can provide a temporary fix here, I am on the way to do this.

dash/include/dash/Shared.h
```diff
@@ -83,7 +105,7 @@ class Shared {
       _owner,
       _team->dart_id());
     team.barrier();
```
devreal: That barrier is superfluous.

dash/include/dash/Shared.h
```diff
@@ -58,6 +58,23 @@ class Shared {
...
+                   "finished delegating constructor");
+  }
+
+  /**
+   * Constructor, allocates shared value at single unit in specified team.
+   * The element is initialized with the given value.
+   */
+  Shared(
+    /// The value to initialize the element with
+    value_type    val,
```
devreal: Why not pass `val` by reference, i.e., `const valye_type&`?

devreal (changes requested):

dash/include/dash/Shared.h
ddiefenthaler: Good point, copied it over from the `set` function,
changed it there as well.

dash/include/dash/Shared.h
ddiefenthaler: Thanks for the information.

I left it alone as it was there before I tempered with the file.
It is removed now.

devreal (approved):

rkowalewski:
Thank you for your changes @ddiefenthaler.

Issue #549, opened 2018-05-17
DART local memory allocation leads to segfault


rkowalewski:
Problem: If we allocate locally on a unit which has different global and local unit ids (e.g., `dash::myid() == 2 && split_team.myid() == 0`), the team id handling is buggy in DART.

It is clear why that is the case. While allocating the local memory we set the global pointer to the **global** unit id based on `DART_TEAM_ALL` (see [here](https://github.com/dash-project/dash/blob/development/dart-impl/mpi/src/dart_globmem.c#L132)). While retrieving the local memory address of the global pointer we lookup the **local** unit id (see [here](https://github.com/dash-project/dash/blob/development/dart-impl/mpi/src/dart_globmem.c#L42)). This of course leads to a local NULL pointer if global and local unit do not match.

So we should rely either only on local or global unit ids.

devreal:
> Problem: If we allocate locally on a unit which has different global and local unit ids (e.g., dash::myid() == 2 && split_team.myid() == 0), the team id handling is buggy in DART.

It is not. Local allocations have no team attached so the dart_gptr carries a global unit ID. The test case in #550 is invalid as it assigns an ID from a random team to a gptr that should contain ~global unit IDs~ unit IDs from DART_TEAM_ALL. We have no way to check that in `dart_gptr_setunit`.

fmoessbauer:
In my opinion the local unit id's should be just a index-set of the corresponding global ids. However for all operations which provide a team parameter, local id's have to be used.

For dart_gptr however only global id's should be used, as otherwise these pointers cannot be passed to another team (or memory accessed by another team).

The conversion should be implicitly done in DART.

fmoessbauer:
... and as @devreal said: The unit-test is invalid.

devreal:
>  The conversion should be implicitly done in DART.

We have to store the team ID in the gptr because allocations are linked to a team, otherwise segment IDs are not unique (only the combination of team_id+segement_id can be guaranteed to be unique). It is also invalid to pass gptr to units outside of the team that allocated the memory, they could not access the underlying window. So storing relative IDs in gptr is perfectly fine.

rkowalewski:
Well the unit test is kind of invalid, right. It is more a workaround to manifest the error.

However, this is definitely a bug. From the practical perspective if we know that team unit id 0 allocates locally a global pointer and then explicitly sets the unit id of the global pointer again to 0, this should be an idempotent operation, right?! And it should never lead to an error.

[...]

PR #550, opened 2018-05-17
[WIP] unit test for bug #549

rkowalewski:


devreal:
Invalid: local memory allocations always refer to `DART_TEAM_ALL`.

PR #552, opened 2018-05-21
Bug 551 lapack

fmoessbauer:
This PR fixes the LAPACK cmake skript. The first assumption in the issue, that the number of arguments has changed was not correct. The only difference is that the `LOCATION` variable must not be empty. 

To ensure that this fix is backward compatible with cmake 2.8, an additional docker container is added, providing a minimal system environment:

```
openmpi 2.1
cmake version 2.8.12.2
gcc (Ubuntu 4.8.4-2ubuntu1~14.04.4) 4.8.4
g++ (Ubuntu 4.8.4-2ubuntu1~14.04.4) 4.8.4
```

The release build is also tested on the minimal environment in CI.

### DIFF
--- a/CMakeExt/LAPACK.cmake
+++ b/CMakeExt/LAPACK.cmake
@@ -4,7 +4,7 @@ find_package(BLAS)
 find_package(LAPACK)
 
 if (BLAS_FOUND)
-  check_library_exists(${BLAS_LIBRARIES} "cblas_sgemm" BLAS_IS_CBLAS)
+  check_library_exists(${BLAS_LIBRARIES} "cblas_sgemm" "C" BLAS_IS_CBLAS)
   if(${BLAS_IS_CBLAS})
   else()
     message(STATUS "No CBLAS Interface found, try manually")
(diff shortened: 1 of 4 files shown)


CMakeExt/LAPACK.cmake
```diff
@@ -4,7 +4,7 @@ find_package(BLAS)
 find_package(LAPACK)
 
 if (BLAS_FOUND)
-  check_library_exists(${BLAS_LIBRARIES} "cblas_sgemm" BLAS_IS_CBLAS)
+  check_library_exists(${BLAS_LIBRARIES} "cblas_sgemm" "C" BLAS_IS_CBLAS)
```
devreal, line 7: While we're at it: why don't we restrict the search for LAPACK to when support for LAPACK has been requested?

rkowalewski:
Thank you for your changes. One additional minor thing. While building OpenMPI you can also add the configure flags `--disable-oshmem` and `--disable-mpi-fortran`. It should additionally boost the build time and, as a result, CI runs are faster.

fmoessbauer:
@rkowalewski: This would only reduce the build time of the container, but not the CI as there pre-build containers are downloaded.

However we could add these flags in another PR

... and @devreal: yes, good point but I'm not a big fan of mixing fixes and  features in the same PR. Especially changes in the build system have so many side effects...

rkowalewski (approved):
Then I am fine.

PR #553, opened 2018-05-21
remove deprecated cmake policies

fmoessbauer:
This PR removes all deprecated (>3.10) policies. To ensure full compatibility with cmake 2.8.12, it should be held back until #552 is merged.

If possible, try to test this PR on real HPC systems.


devreal:
I tested the branch on the Cray system and it seems to work fine. Shouldn't we try to get rid of the other policies as well? I tried disabling all others and it seems to work fine (minus the one needed for ninja, that is).

fmoessbauer:
@devreal I have no problem with removing them. In the rare case that issues arise after removing them, we can add it again.

rkowalewski (approved):
If the final CI build is successful we should merge.

PR #554, opened 2018-05-22
Add sanity checks to dart_gptr_setunit

devreal:
Note that this prevents `dart_gptr_setunit` from being inlined...


devreal:
Ahh, this failing CI comes from the broken handling of local memory allocation in `dash::Shared`, as mentioned in #549

devreal:
Not worth it.

PR #559, opened 2018-06-01
Improve Coverage CI

fmoessbauer:
- Integrate coverage CI into DASH-CI scripts
- execute configuration with and without shared / dynamic windows


devreal (approved):
Looks good, thanks for taking care of this :+1:

PR #560, opened 2018-06-08
Feat halo

dhinf:
added some functionality and minor fixes


dash/include/dash/halo/Halo.h
```diff
@@ -76,6 +76,62 @@ class StencilPoint : public Dimensional<int16_t, NumDimensions> {
...
+    }
+
+    return coords;
+  }
+
+  /**
+   * Returns coordinates adjusted by stencil point
+   */
+  template<typename ElementCoordsT>
+  static ElementCoordsT stencil_coords(ElementCoordsT coords,
```
devreal: Why not implement the above function `stencil_coords` in terms of the static `stencil_coords` to reduce code duplication?

dash/include/dash/halo/Halo.h
```diff
@@ -76,6 +76,62 @@ class StencilPoint : public Dimensional<int16_t, NumDimensions> {
...
+    }
+
+    return coords;
+  }
+
+  /**
+   * Returns coordinates adjusted by stencil point
+   */
+  template<typename ElementCoordsT, typename ViewSpecT>
+  std::pair<ElementCoordsT,bool> stencil_coords_check(
```
devreal: Document what `stencil_coords_check` does different from `stencil_coords`

dash/include/dash/halo/Halo.h
```diff
@@ -76,6 +76,62 @@ class StencilPoint : public Dimensional<int16_t, NumDimensions> {
...
+    }
+
+    return std::make_pair(coords,halo);
+  }
+
+  /**
+   * Returns coordinates adjusted by stencil point
+   */
+  template<typename ElementCoordsT, typename ViewSpecT>
+  std::pair<ElementCoordsT,bool> stencil_coords_check_abort(
```
devreal: Does it abort the program? Where?

dash/include/dash/halo/StencilOperator.h
```diff
@@ -84,6 +84,34 @@ class StencilOperatorInner {
...
+   * The stencil points are multiplied with their coefficent (\ref StencilPoint)
+   * and the center is  multiplied with the given center coefficient.
+   * No test for halo elements. -> useful only for elements with stencil
+   * points pointing to inner elements. (no halo elements)
+   *
+   * \param coords center coordinate
+   * \param coefficient for center
+   * \param op operation to use (e.g. std::plus). default: std::plus
+   */
+  ElementT get_value_at(
```
devreal: Document that the value of the stencil is not modified

dash/include/dash/halo/StencilOperator.h
```diff
@@ -84,6 +84,34 @@ class StencilOperatorInner {
...
+    const ElementCoords_t& coords, ElementT coefficient_center,
+    std::function<ElementT(const ElementT&, const ElementT&)> op =
```
devreal: Make the function a template parameter, e.g., `BinaryFunc`, to avoid creating a `std::function` from lambdas

devreal (changes requested):
Small comments inline, looks good otherwise :+1:

dash/include/dash/halo/StencilOperator.h
```diff
@@ -439,22 +663,22 @@ class StencilOperator {
     return stencil_offs;
   }
 
-  pattern_index_t get_offset(const ElementCoords_t& coords) const {
-    pattern_index_t offset = 0;
-
+  StencilOffsets_t set_dimension_offsets() {
+    StencilOffsets_t      dim_offs;
+    signed_pattern_size_t offset = 0;
```
bertwesarg, line 671: My compiler tells me, that this was declared but never referenced

fmoessbauer:
There are compiler errors in clang-mpich using minimal build config:

https://4443-45930061-gh.circle-artifacts.com/0/tmp/build-logs/clang_mpich/20180828-142551--uuid-1f79b8a1/clang/Minimal/build.log
```
In file included from /root/project/dash/include/dash/halo/HaloMatrixWrapper.h:8:
/root/project/dash/include/dash/halo/StencilOperator.h:216:20: warning: 'constexpr' non-static member function will not be implicitly 'const' in C++14; add 'const' to avoid a change in behavior [-Wconstexpr-not-const]
    constexpr void operator()(const StencilOffsets_t& stencil_offs,
                   ^
/root/project/dash/include/dash/halo/StencilOperator.h:232:20: warning: 'constexpr' non-static member function will not be implicitly 'const' in C++14; add 'const' to avoid a change in behavior [-Wconstexpr-not-const]
    constexpr void operator()(const StencilOffsets_t& stencil_offs,
                   ^
/root/project/dash/include/dash/halo/StencilOperator.h:216:20: error: constexpr function's return type 'void' is not a literal type
    constexpr void operator()(const StencilOffsets_t& stencil_offs,
                   ^
/root/project/dash/include/dash/halo/StencilOperator.h:232:20: error: constexpr function's return type 'void' is not a literal type
    constexpr void operator()(const StencilOffsets_t& stencil_offs,
```

dhinf:
oh, void is a literal in c++14 but not in 11. Why we still use C++11?

devreal (approved):

dash/include/dash/halo/StencilOperator.h
devreal: @dhinf Can this function be removed?

dhinf: no, it is used.

PR #561, opened 2018-06-11
[WIP] dart Min/Max reduction

pascalj:
See #548. The benchmark itself will be removed. I ran it on 1 to 8 nodes with constant 32 processes and the results are pretty much what I would expect, although attaching the shared variable seems _really_ expensive.

1806jn0ro

How do we handle things like [this cast](https://github.com/dash-project/dash/compare/development...pascalj:bench-allreduce?expand=1#diff-017a40a28d52a5bca9d048272afceeb5R86)? Macros or should we simply do all this in C++?

PR #562, opened 2018-06-12
Provide DART_OP_MINMAX

devreal:
Based on #561 (and after talking to @pascalj), this PR implements MINMAX reduction for basic types in DART.

### DIFF
--- a/dart-impl/mpi/src/dart_mpi_op.c
+++ b/dart-impl/mpi/src/dart_mpi_op.c
@@ -7,7 +7,7 @@
 
 typedef void (*dart__mpi_min_max_reduce_t)(void*, void*, int*, MPI_Datatype*);
 
-#define DART_NAME_MINMAX_OP(__name) dart__mpi_min_max_reduce##__name
+#define DART_NAME_MINMAX_OP(__name) dart__mpi_min_max_reduce_##__name
 
 #define DART_DECLARE_MINMAX_OP(__name) \
 DART_INLINE void DART_NAME_MINMAX_OP(__name)(            \
@@ -29,6 +29,7 @@ DART_DECLARE_MINMAX_OP(__name){                             \
 
 DART_DEFINE_MINMAX_OP(byte,             char)
 DART_DEFINE_MINMAX_OP(short,            short int)
+DART_DEFINE_MINMAX_OP(int,              int)
 DART_DEFINE_MINMAX_OP(unsigned,         unsigned int)
 DART_DEFINE_MINMAX_OP(long,             int)
 DART_DEFINE_MINMAX_OP(unsignedlong,     unsigned long)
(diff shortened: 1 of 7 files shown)


dart-impl/mpi/src/dart_mpi_op.c
```diff
@@ -0,0 +1,117 @@
...
+static MPI_Op dart__mpi_minmax_reduce_ops[DART_TYPE_LAST];
+
+MPI_Op dart__mpi__op_custom(dart_operation_t op, dart_datatype_t type)
+{
+  DART_ASSERT_MSG(op == DART_OP_MINMAX, "Unknown custom operation!");
+  dart_datatype_t basetype = dart__mpi__datatype_base(type);
+  return dart__mpi_minmax_reduce_ops[basetype];
+}
+
+dart_ret_t dart__mpi__op_init()
```
pascalj, line 69: I presume this is not very expensive, else we should do it lazily.

dash/examples/bench.13.allreduce/main.cpp
```diff
@@ -0,0 +1,223 @@
...
+  const dash::util::BenchmarkParams & bench_cfg,
+  const benchmark_params            & params);
+
+
+measurement evaluate(
+              int reps,
+              std::string testcase,
+              benchmark_params params);
+
+int main(int argc, char** argv)
```
pascalj, line 51: Should we even keep this? No one will probably run it again ;-)

dart-impl/mpi/src/dart_mpi_op.c
```diff
@@ -0,0 +1,117 @@
...
+#include <dash/dart/mpi/dart_communication_priv.h>
+#include <dash/dart/base/assert.h>
+
+typedef void (*dart__mpi_min_max_reduce_t)(void*, void*, int*, MPI_Datatype*);
+
+#define DART_NAME_MINMAX_OP(__name) dart__mpi_min_max_reduce_##__name
+
+#define DART_DECLARE_MINMAX_OP(__name) \
+DART_INLINE void DART_NAME_MINMAX_OP(__name)(            \
+  void *lhs_, void *rhs_, int *len, MPI_Datatype *dptr)
```
pascalj: Sorry, I didn't have that in my version either. We can make the lhs `const` here and below.

pascalj (approved):
Looks good to me! Left some comments inline.

dart-impl/mpi/src/dart_mpi_op.c
devreal: You mean the creation of ops in `dart__mpi__op_init`? I don't think that is very expensive regarding the memory footprint. Doing it lazily would require locking to make it thread-safe and add overhead at application runtime instead of during initialization.

dash/examples/bench.13.allreduce/main.cpp
devreal: Yes, I would keep it, it's interesting to see and could be used for evaluation later.

dart-impl/mpi/src/dart_mpi_op.c
devreal: Technically we could but the MPI standard does not have the `const` in it's interface so we'd have to do some nasty casts when passing the function to MPI to suppress compiler warnings.

devreal: Of course we can make `lhs` inside the function body `const`, didn't think of that initially (not sure that that changes anything

PR #563, opened 2018-06-18
Add support for custom reduction operations

devreal:
This PR adds support for custom reduction operations and custom types in DART. As an example, this allows DASH to map any commutative reduction operation passed to `dash::accumulate` onto a call to `MPI_Allreduce`. 

This is accomplished by duplicating the underlying MPI type in `dart_op_create`, storing the provided operator and user-data (e.g., pointer to a lambda object) in a hash-map, and invoking them upon invocation of the DART wrapper reduction operator in MPI. There is some overhead involved due to querying of the hash map and issuing two additional indirect function calls (calling the function provided to DART and invoking the user-provided reduction operator) but the benefit of using the optimized MPI reduction instead of temporary global memory outweighs this by order(s) of magnitude. 

Measurements from the updated MinMax benchmark on 4x24 units on the Cray XC40: 
```
   96,  craympi,         dart_allreduce.minmax,  0.00613600
   96,  craympi,            dart_allreduce.min,  0.01256200
   96,  craympi,         dart_allreduce.shared,  0.48320600
   96,  craympi,         dart_allreduce.custom,  0.00812400
   96,  craympi,         dart_allreduce.lambda,  0.00818300
```

In case `dash::accumulate` can be mapped onto predefined DART types/operations the implementation avoids the overhead of the additional indirect calls.

Custom types are implemented as simple contiguous MPI types, which is sufficient to apply the reduction operations on arbitrary user-defined objects. The additional duplication is necessary to ensure thread-safety but can be avoided if the caller signals that the used DART type is temporary, i.e., local to the caller and not used by multiple threads.

The changes to `dash::accumulate` ensure that every unit receives the result (as it calls `dart_allreduce`). This PR also adds a variant that can be called on native pointers to provide a simple wrapper for the `MPI_Allreduce` functionality.


dash/examples/bench.14.accumulate/main.cpp
```diff
@@ -7,6 +7,7 @@
 #include <iostream>
 #include <iomanip>
 #include <string>
+#include <mpi.h>
```
fmoessbauer: Please guard this as we cannot assume that MPI is always available.

dash/include/dash/algorithm/Accumulate.h
```diff
@@ -10,9 +9,42 @@
...
+    }
+  }
+} // namespace internal
+
+
 /**
- * Accumulate values in range \c [first, last) as the sum of all values
- * in the range.
+ * Accumulate values in each process' range \c [first, last) using the
+ * given binary reduce function \c op, which should be a commutative operation.
```
pascalj: I don't understand why MPI says it _should_ be a commutative operation. If it isn't we have no guarantee to get the correct result (it's not a left fold anymore in our case). Even if MPI somehow does the right thing, we still apply `binary_op(result, init)` at the end.

dash/include/dash/algorithm/Accumulate.h
```diff
@@ -86,45 +165,23 @@ ValueType accumulate(
...
-  if (myid == 0) {
-    for (size_t i = 0; i < team.size(); i++) {
-      local_result lr = l_results[i];
-      if (lr.l_valid) {
-        result = binary_op(result, lr.l_result);
-      }
-    }
-  }
-  return result;
+  // TODO: can we figure out whether or not units are empty?
```
pascalj, line 216: A bit verbose, maybe:
```c++
  auto empty = false;
  for (dash::team_unit_t unit{0}; unit < pattern.team().size(); unit++) {
     auto lsizes = pattern.local_extents(unit);
     empty = empty || std::any_of(lsizes.begin(), lsizes.end(), [](size_t size) { return size != 0; });
  }
```

Not sure if something like this is defined already. I couldn't find it.

dash/include/dash/algorithm/Accumulate.h
devreal: That has nothing to do with MPI. We cannot guarantee the correct *global* order (as you said) so the operation has to be commutative. MPI allows for non-commutative operations but we cannot use them.

dash/include/dash/algorithm/Accumulate.h
devreal: That would only account for the whole range of the of allocation, not the range defined by the iterators. The monkey in my brain yells *blocks blocks blocks* but I'm not sure they are of any help here.

dash/include/dash/algorithm/Accumulate.h
pascalj: Okay, so it should be "which must be commutative" then.

fmoessbauer: This is not sufficient, as the function has to be linear. Consider e.g. `(a,b) : a^2+b^2`

dash/include/dash/algorithm/Accumulate.h
```diff
@@ -86,45 +165,23 @@ ValueType accumulate(
 template <
   class GlobInputIt,
   class ValueType,
-  class BinaryOperation >
+  class BinaryOperation = dash::plus<ValueType>,
```
rkowalewski, line 201: Please use our iterator traits to check if `GlobInputIt` is really a global iterator. This is more readable.

dash/include/dash/algorithm/Operation.h
```diff
@@ -45,19 +45,71 @@ class ReduceOperation {
...
+  : public std::integral_constant<dart_operation_t, DART_OP_UNDEFINED>
+{ };
+
+/**
+ * Query the DART operation for an arbitrary binary operation.
+ * Overload for DART operations.
+ */
+template<>
+template<typename BinaryOperation>
+struct dart_reduce_operation<BinaryOperation,
```
rkowalewski: Can you please add some comments to document this meta programming stuff. Just to understand what it does and what the `enable_if` requirements are.

dash/examples/bench.13.allreduce/main.cpp
```diff
@@ -153,10 +181,56 @@ measurement evaluate(int reps, std::string testcase, benchmark_params params)
...
+        if (a.min < res.min) {
+          res.min = a.min;
+        }
+        if (a.max > res.max) {
+          res.max = a.max;
+        }
+        return res;
+      };
+      dart_op_create(
+        &minmax_lambda<decltype(fn)>, &fn, true, new_type, true, &new_op);
```
rkowalewski, line 219: can you please document the parameters here, especially the boolean values. Either as a comment or, even better, in the form of variable names.

rkowalewski (changes requested):
comments inline

dash/include/dash/algorithm/Accumulate.h
pascalj: So the binop needs to be commutative and associative: it might be a good idea to rename the algorithm to `dash::reduce`. This not only prevents confusion about the order (like with `dash::for_each`): if we additionally require `init` to be the neutral element instead of an initial value, we could get rid of a lot of "empty node" code.

dash/include/dash/algorithm/Operation.h
devreal: I added some documentation on the `std::enable_if` and moved it to `dash::internal`.

dash/include/dash/algorithm/Accumulate.h
devreal: I extended the documentation a bit and included that the function has to be linear along with a description of the reduction process. 

I am also not happy with the way we copied the semantics of `std::accumulate`. First, we cannot guarantee a left fold (without significant performance issues) so the name of `dash::accumulate` might be confusing. Second, the `init` argument has to be the same on all units, otherwise the result is bogus. 

I'm happy to rename `dash::accumulate` to `dash::reduce` and change `init` to be a neutral element if there is consent to do so.

dash/examples/bench.13.allreduce/main.cpp
devreal: Done.

dash/include/dash/algorithm/Accumulate.h
rkowalewski: I have a similar problem in `dash::sort`. It is probably not the best approach but I have to know which units are empty.
See [here](https://github.com/dash-project/dash/blob/development/dash/include/dash/algorithm/Sort.h#L567)

rkowalewski (approved):

PR #566, opened 2018-06-21
Rename MPI_IMPL_ID to DASH_MPI_IMPL_ID

devreal:
Rename `MPI_IMPL_ID` to `DASH_MPI_IMPL_ID` and include it in `dash-mpicxx` to allow HDF5 support to properly build.


fmoessbauer (approved):
Thanks for taking care of this. Our own defines should always be prefixed with `DASH` or `DART`

Issue #569, opened 2018-07-02
Support for All to All Operations


rkowalewski:
DART should provide operations to support both blocking and nonblocking all-to-all operations.

- blocking: `MPI_Alltoall` and `MPI_Alltoallv`
- nonblocking: `MPI_Ialltoall` and `MPI_Ialltoallv`

Currently the only way to transpose an (N-)Array is to apply one-sided operations. This is, of course, not scalable with a large number of processes where logarithmic complexity is imperative, compared to linear complexity in the "naive" transpose.

rkowalewski:
An interesting use case for `dart_Ialltoallv` is `dash::sort` where all processors have to distribute local data to the corresponding target processors. This can be overlapped with local copy operations.

devreal:
I think there was once of student from TUM who had a `dart_alltoall` implementation. I will look into that

PR #570, opened 2018-07-23
Performance efficiency improvements for dash::sort

rkowalewski:
This branch has couple of changes to increase efficiency of `dash::sort`.

- Internal functions have been moved to a new header (`internal/Sort-inl.h`)
- There are various const correctness "fixes" to enable more aggressive optimization by the compiler
- Replace 2 collective min / max reductions with a single one as proposed by #562.
- Replace one-sided allreduce with a collective dart reduce.
- Added an all-to-all collective communication in DART since it is needed for efficient data exchange, especially in large-scale use cases. This can be replace with more enhanced versions if #569 is fixed.
- Replace the final local sort with a merge. This drastically increases the performance and may achieve a speedup up to 4 (benchmark results on projekt03) depending on the number of sequences. What we utilize here is the fact that all sequences which have been "received" from other units, are already sorted. We merge these sorted sequences pairwise in a binary tree.

I will provide additional benchmarks results soon.


dart-impl/mpi/src/dart_communication.c
```diff
@@ -2193,6 +2193,54 @@ dart_ret_t dart_allreduce(
...
+{
+  DART_LOG_TRACE("dart_alltoall() team:%d nelem:%" PRIu64 "", teamid, nelem);
+
+  CHECK_IS_BASICTYPE(dtype);
+
+  /*
+   * MPI uses offset type int, do not copy more than INT_MAX elements:
+   */
+  if (dart__unlikely(nelem > MAX_CONTIG_ELEMENTS)) {
+    DART_LOG_ERROR("dart_allreduce ! failed: nelem (%zu) > INT_MAX", nelem);
```
fmoessbauer: Can we recover from this? It is also not documented in the interface that there is a limit.

dash/include/dash/algorithm/Sort.h
```diff
@@ -1139,43 +468,59 @@ void sort(GlobRandomIt begin, GlobRandomIt end, SortableHash sortable_hash)
...
+  trace.exit_state("13:calc_final_send_count");
 
-  trace.enter_state("13:barrier");
-  team.barrier();
-  trace.exit_state("13:barrier");
+  trace.enter_state("14:calc_recv_count (all-to-all)");
+
+  std::vector<size_t> recv_count(nunits, 0);
+
+  dart_alltoall(
```
fmoessbauer: Wrap this call as it might be not successful

dash/include/dash/algorithm/Sort.h
```diff
@@ -1192,16 +537,13 @@ void sort(GlobRandomIt begin, GlobRandomIt end, SortableHash sortable_hash)
...
-    if (unit == unit_at_begin) {
-      it_copy = begin;
-    }
-    else {
-      // The array passed to global_index is 0 initialized
-      auto gidx =
-          pattern.global_index(static_cast<dash::team_unit_t>(unit), {});
-      it_copy = iter_type{&(begin.globmem()), pattern, gidx};
-    }
+    iter_type it_copy =
```
fmoessbauer: Uff... that could be worth a comment ;)

dash/include/dash/algorithm/internal/Sort-inl.h
```diff
@@ -0,0 +1,734 @@
...
+  }
+
+  std::vector<size_t> target_displs(nunits, 0);
+
+  auto const u_blocksize = g_partition_data.lsize();
+
+  for (auto const& border_idx : valid_partitions) {
+    auto const   left_u  = p_borders.left_partition[border_idx];
+    auto const   right_u = border_idx + 1;
+    size_t const val =
```
fmoessbauer, line 427: Probably add a comment?

fmoessbauer (commented):
Most of it looks really good. I noticed only some minor points. See my comments inline.

rkowalewski:
Thank you for your good review comments @fmoessbauer. Hopefully everything is addressed now.

dash/include/dash/algorithm/Sort.h
```diff
@@ -889,133 +222,134 @@ void sort(GlobRandomIt begin, GlobRandomIt end, SortableHash sortable_hash)
         [&is_skipped](size_t idx) { return is_skipped[idx] == false; });
   }
 
+  DASH_LOG_TRACE_RANGE(
+      "valid partitions",
+      std::begin(valid_partitions),
+      std::end(valid_partitions));
+
   if (valid_partitions.empty()) {
```
devreal, line 233: Would `valid_partitions` be empty on all units or would the one unit have one partition? If so are  there any collectives on the path taken by that unit? If so we have a problem because the non-empty unit will deadlock...

devreal (changes requested):
Looks good to me, just one thing I noticed...

dash/include/dash/algorithm/Sort.h
rkowalewski: `valid_partitions` is always equal on all units. ;) So, all units will take this path if no valid partitions have been found.

devreal (approved):

fmoessbauer:
Why didn't we notice the compilation error in our CI?

rkowalewski:
> Why didn't we notice the compilation error in our CI?

We were wondering about this as well and there is definitely something strange going on. @pascalj is fixing that. ;)

fmoessbauer:
Is this PR ready to merge?

fmoessbauer (approved):

rkowalewski:
Discussed in the F2F. Ready to merge if unit tests have successfully passed.

rkowalewski:
I am not sure why our codecov often hangs, any ideas @fmoessbauer? It just blocks merging and I do not really prefer the hard "sudo" way to force the merge.

fmoessbauer:
@rkowalewski: There is a problem with external branches (from forks). 

```
==> Uploading reports
    url: https://codecov.io
    query: branch=pull%2F570&commit=4f710f9976e02e755a229a418cd6018ebc8e1d9b&build=4682&build_url=&name=&tag=&slug=dash-project%2Fdash&service=circleci&flags=&pr=570&job=0
    -> Pinging Codecov
HTTP 400
Commit sha does not match Circle build.
```

rkowalewski:
Hm OK. Then I am not using my private repository anymore for pull requests.

PR #571, opened 2018-07-27
Delayed Initialization for dash::Mutex

rkowalewski:
Our `dash::Mutex` did not support delayed initialization until now. There are  use cases which require a global lock handle which can then be reused within the application.

This pull request added this feature and contains additional refactoring, mainly the replacement of the native pointer member in `dash::Mutex` with a `std::unique_ptr`.

### DIFF
--- a/dash/include/dash/Mutex.h
+++ b/dash/include/dash/Mutex.h
@@ -97,7 +97,7 @@ public:
 private:
   dash::Team const* _team{nullptr};
   std::unique_ptr<std::remove_pointer<dart_lock_t>::type, DestroyDARTLock>
-      _mutex{nullptr};
+      _mutex{DART_LOCK_NULL};
 };  // class Mutex
 
 }  // namespace dash
--- a/dash/include/dash/Mutex.h
+++ b/dash/include/dash/Mutex.h
@@ -73,6 +73,8 @@ public:
    * Collective initialization of the DART lock.
    *
    * This function is not thread-safe
+   *
+   * @return True if lock was successfully initialized, False otherwise
    */
   bool init();
 
--- a/dash/include/dash/Mutex.h
+++ b/dash/include/dash/Mutex.h
@@ -34,7 +34,7 @@ private:
   struct DestroyDARTLock {
     void operator()(dart_lock_t lock)
     {
-      if (nullptr != lock) {
+      if (DART_LOCK_NULL != lock && dart_initialized()) {
         auto ret = dart_team_lock_destroy(&lock);
 
         if (ret != DART_OK) {
--- a/dash/include/dash/Mutex.h
+++ b/dash/include/dash/Mutex.h
@@ -34,7 +34,7 @@ private:
   struct DestroyDARTLock {
     void operator()(dart_lock_t lock)
     {
-      if (DART_LOCK_NULL != lock && dart_initialized()) {
+      if (DART_LOCK_NULL != lock) {
         auto ret = dart_team_lock_destroy(&lock);
 
         if (ret != DART_OK) {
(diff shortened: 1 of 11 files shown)


fmoessbauer:
This PR looks good! However this also includes the design decision to create objects in an "invalid" state (ok, you check this by using assertions). I'm not a big fan of delayed initialisation, as this problem could be easily solved by using a global `std::unique_ptr` which is bound when initialisation is possible. This also solves many problems with recursive includes, as the size of the mutex struct does not have to be known at declaration time.

In the more general case it is questionable if all methods of an object have to be callable at all times. Ideally that should be the case, but almost no library supports that.

In short, merge it after @devreal has verified the DART stuff.

dash/include/dash/Mutex.h
```diff
@@ -30,50 +31,75 @@ class Mutex {
...
-  
-private:
-  dart_lock_t   _mutex;
-}; // class Mutex
 
-} // namespace dash
+private:
+  dash::Team const* _team{nullptr};
+  std::unique_ptr<std::remove_pointer<dart_lock_t>::type, DestroyDARTLock>
+      _mutex{nullptr};
```
devreal: I would suggest to introduce a constant `DART_LOCK_NULL` (= `(dart_lock_t*)NULL`) and use that instead of `nullptr` and `NULL` in the DASH and DART part as `dart_lock_t` is an opaque type (that happens to be `typedef`ed to a pointer).

devreal (changes requested):
Thanks for taking care of this, just one comment inline

devreal:
@fmoessbauer 

> I'm not a big fan of delayed initialisation, as this problem could be easily solved by using a global `std::unique_ptr` which is bound when initialisation is possible. 

I'm not sure I can follow you here. How would make sure the lock is initialized after the runtime initialization has been called?

fmoessbauer:
@devreal: I meant it this way:

```cpp
std::unique_ptr<dash::mutex> mx;

int main(){
  dash::init();
  mx = std::make_unique<dash::mutex>();
  // use it

  // otherwise the mutex has to register on the global deallocation list
  mx.reset();
  dash::finalize();
  // mutex has to be deconstructed at this point in time
}
```

rkowalewski:
> I'm not sure I can follow you here. How would make sure the lock is initialized after the runtime initialization has been called?

If I understood it correctly we can define a global `std::unique_ptr` pointing initially to a `nullptr` and reset it after runtime initialization with a new `Mutex`. While this is possible it is neither the optimal solution nor a better one IMHO because we can always reset the `unique_ptr`. It smells like misusing `unique_ptr` as a wrapper to postpone initialization.

However, in general I agree with @fmoessbauer that our concept of delayed initialization is kind of special.

rkowalewski:
Well seems like we wrote a comment concurrently. OK to solve your use case I would expect that DART finalizes all globally allocated memory, teams, etc.

And then we just have to check if dart is initialized in the internal deleter of our `dash::Mutex`. Or am I missing something?

dash/include/dash/Mutex.h
rkowalewski: Good point. Maybe you can "overtake" this issue @devreal. I thought about this however, I was not sure how to define a `DART_LOCK_NULL` handle since we have opaque members in it.

devreal: I will push something later today

devreal:
I changed the implementation of the DART locks such that they are registered with the team and the segments are released upon destruction of the team. The `dash::Mutex` wrapper then still calls `dart_team_lock_destroy` to release the `malloc`'ed memory, which may happen at any time, e.g., at the end of `main`.

devreal (approved):

fmoessbauer (approved):

fmoessbauer:
Is there still something to change, or can we merge?

rkowalewski:
Ready to merge. ;)

PR #574, opened 2018-08-01
DART non-collective global memory allocator

devreal:
This PR introduces `dart_allocator_t`, which can be used to allocate global memory non-collectively from a previously collectively allocated global memory pool. The functionality is similar to `dart_memalloc` but allows to control the pool size (`dart_memalloc` might at some point make use of `dart_allocator_t`). 

To integrate this with DASH, we would need to integrate this with the MemorySpaces somehow, something I have not looked at yet. I vaguely remember that @rkowalewski is working on this?

Allocators have been proposed by @fmoessbauer in #572


dart-if/include/dash/dart/if/dart_globmem.h
```diff
@@ -257,6 +257,89 @@ dart_ret_t dart_gptr_setflags(
...
+ * The allocator should be destroyed after all allocations have been free'd
+ * using \ref dart_allocator_destroy.
+ */
+typedef struct dart_allocator_struct * dart_allocator_t;
+
+/**
+ * Create a new allocator for non-collective global memory allocations.
+ * This operation is collective among the units in \c team.
+ *
+ * \param pool_size The size of the local memory pool from which global memory
```
fmoessbauer: I guess the pool size is in bytes? That should be mentioned in the documentation. What sizes are allowed? I guess the body allocator requires a 2^n number?

dart-if/include/dash/dart/if/dart_globmem.h
```diff
@@ -257,6 +257,89 @@ dart_ret_t dart_gptr_setflags(
...
+  size_t             pool_size,
+  dart_team_t        team,
+  dart_allocator_t * new_allocator);
+
+/**
+ * Allocate global memory from a DART allocator instance created using
+ * \ref dart_allocator_new. The operation is not a collective operation but only
+ * involves the local unit. The resulting \ref dart_gptr_t can be communicated
+ * to other units for global memory access.
+ *
```
fmoessbauer, line 302: Please add some notes on the alignment. Normally I would not bother, but we do HPC ;)

dart-if/include/dash/dart/if/dart_globmem.h
```diff
@@ -257,6 +257,89 @@ dart_ret_t dart_gptr_setflags(
...
+ * \param gptr      Pointer to the \ref dart_gptr_t, will be reset to
+ *                  \c DART_GPTR_NULL.
+ * \param allocator The allocator from which the memory was allocated.
+ *
+ * \return \c DART_OK on success, any other of \ref dart_ret_t otherwise.
+ *
+ * \threadsafe
+ * \ingroup DartGlobMem
+ */
+dart_ret_t dart_allocator_free(
```
fmoessbauer, line 333: Who is allowed to free? Only the owner or any unit?

dart-impl/mpi/src/dart_mem.c
```diff
@@ -213,33 +213,35 @@ int dart_buddy_free(struct dart_buddy * self, uint64_t offset)
...
+      else {
+        left += length;
+        index = index * 2 + 2;
+      }
+      break;
+    }
+  }
 
   dart__base__mutex_unlock(&self->mutex);
   // TODO: is this ever reached?
```
fmoessbauer, line 265: Maybe add assertion

dash/test/dart/DARTMemAllocTest.cc
```diff
@@ -169,3 +169,42 @@ TEST_F(DARTMemAllocTest, SegmentReuseTest)
     DART_OK,
     dart_team_memfree(gptr2));
 }
+
+
+TEST_F(DARTMemAllocTest, AllocatorTest)
```
fmoessbauer: Can we test allocation / deallocation for various pool-sizes, including impossible allocations. AFAIK the buddy allocation is not extensively tested yet.

We should also consider fragmentation.

fmoessbauer (changes requested):
See my comments inline

dash/test/dart/DARTMemAllocTest.cc
```diff
@@ -169,3 +169,42 @@ TEST_F(DARTMemAllocTest, SegmentReuseTest)
...
+  dart_allocator_t allocator;
+  dart_gptr_t gptr;
+
+  // collectively create a new allocator
+  dart_allocator_new(1024, DART_TEAM_ALL, &allocator);
+  //ASSERT_NE_U(allocator, NULL);
+
+  // allocate some local memory on a single unit and distribute the result
+  if (dash::myid() == 0) {
+    dart_allocator_alloc(16, DART_TYPE_INT, &gptr, allocator);
```
rkowalewski: As far as I can see this does not work if we have more than 16 units, right? This may lead to some surprises.
Moreover, please do not pass magic numbers to DART API calls. Use readable names instread., i.e. 
`size_t const nels = 16`

rkowalewski (changes requested):

dart-if/include/dash/dart/if/dart_globmem.h
devreal: We currently don't have any guarantees (unfortunately) on the memory allocation, mostly because MPI does not provide us with any guarantees and we haven't implemented it ourselves. I would thus not make any statements here (in reality, natural alignment should be guaranteed).

fmoessbauer (approved):
Now it looks good IMO. But we should wait for @rkowalewski s response as well.

fuchsto:
@devreal 
Dynamic global allocation based on epochs has been introduced with my implementation of `dash::UnorderedMap` and `dash::List` (both working proofs of concept, but in prototype state) and extended with the (fully functional) implementation of `dash::Graph`.
Epoch-based allocation however uses collective commits, so I'm eager to see how these containers can benefit from your non-collective techniques.
We can discuss this in the DASH project meeting. Coincidentally, I will present `dash::Graph` anyways.

rkowalewski (approved):

PR #575, opened 2018-08-05
Report errors on CI

pascalj:
The current `eval | tee`... calls ignore the error code of the `eval` invocation. That way we never know when a build breaks. See #570 for an example. The source wasn't building at all and the runners still reported success.

Note that `EXEC_WRAP` isn't used any more (right?) so I removed it.
I used a `OUTPUT="$CMD"; RETVAL=$?; echo $OUTPUT...`-construct so we get the correct return type (becuase `pipefail` is a bashism). Also note that the errors in the mpich-version are now reported and a failing build will also be reported.

### DIFF
--- a/dash/scripts/dash-ci-deploy.sh
+++ b/dash/scripts/dash-ci-deploy.sh
@@ -254,8 +254,8 @@ mkdir -p build
 rm -Rf ./build/*
 (cd ./build && cmake $BASEPATH $BUILD_SETTINGS | nocolor && \
  await_confirm && \
- make -j $MAKE_PROCS VERBOSE=1 $MAKE_TARGET
+ make -j $MAKE_PROCS VERBOSE=1 $MAKE_TARGET || exit 1
  if echo "$BUILD_TYPE" | grep "Coverage"; then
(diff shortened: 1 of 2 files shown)


fmoessbauer:
Thanks for taking care of this. I'm going to have a closer look at this tomorrow.

fmoessbauer (commented):
The changes look good and luckily simplify the CI scripts. The failures are not build-related, but have been present for some time without notice. As this is all DART related, @devreal should have a look at it. If that is not possible in the next days, i vote for disabling this tests to get the CI working again.

Unfortunately the EXPECT-wrappers from @rkowalewski vanish the original location of the error. I recommend to remove them or find a way to get both type-safe comparisons and symbol information.

devreal:
I will have a look at the failing DART tests today

pascalj:
Thanks @fmoessbauer and @devreal! Just rebased and the CI seems to be happy again.

fmoessbauer (approved):
Yey, if the CI is happy, I'm happy too and we can merge ;)

PR #577, opened 2018-08-09
CI: Print filename and line in failed assertions

devreal:
This information is lost in the `::testing::internal::EQAsserter` as the filename and line of that class is used. Eventually we should use `static_if` in a macro once we are able to in order to properly print the filename.


fmoessbauer (approved):
Perfect, thanks

PR #578, opened 2018-08-10
Delayed Initialization for dash::Shared

rkowalewski:
Similar to #571 we need delayed initialization for `dash::Shared` as well. There are further improvements:

- Like our other containers we provide now a `local()` interface as a shortcut for the owning unit.
- Replace the internal `shared_ptr` encapsulating the `globmem` instance with a `unique_ptr`. I cannot see any reason for `shared_ptr`. Please correct me if I am wrong.

### DIFF
--- a/dash/include/dash/Shared.h
+++ b/dash/include/dash/Shared.h
@@ -87,7 +87,7 @@ public:
   {
     DASH_LOG_DEBUG_VAR("Shared.Shared(value,team,owner)()", owner);
     if (dash::is_initialized()) {
-      init(val);
+      DASH_ASSERT_RETURNS(init(val), true);
     }
   }
 
--- a/dash/include/dash/Shared.h
+++ b/dash/include/dash/Shared.h
@@ -1,6 +1,9 @@
 #ifndef DASH__SHARED_H_
 #define DASH__SHARED_H_
 
+#include <memory>
+#include <dash/std/memory.h>
+
 #include <dash/dart/if/dart_types.h>
 
 #include <dash/memory/GlobStaticMem.h>
@@ -9,7 +12,6 @@
 
 #include <dash/iterator/GlobIter.h>
 
-#include <memory>
 
 
 namespace dash {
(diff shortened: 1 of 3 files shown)


devreal:
> Replace the internal `shared_ptr` encapsulating the globmem instance with a `unique_ptr`. I cannot see any reason for shared_ptr. Please correct me if I am wrong.

Is there a specific reason we have `shared_ptr` in the other containers?

fmoessbauer:
Well, as copying was allowed, the shared pointer was necessary to avoid deallocating the globmem. However I do not know of any location where we used that.

rkowalewski:
> Well, as copying was allowed, the shared pointer was necessary to avoid deallocating the globmem. However I do not know of any location where we used that.

Exactly. That is why I deleted. It is consistent to our other containers which do not provide copy constructor / assignment either. Such a heavyweight copy should not be hidden, the user can do it explicitly.

> Is there a specific reason we have shared_ptr in the other containers?

Which container still uses that?

rkowalewski:
I am not really happy to move the std memory header into the dash directory. It should be a separate directory in the root. This means additional install directories in our CMake Build. If you prefer this, I can do that as well.

devreal:
>> Is there a specific reason we have shared_ptr in the other containers?

>Which container still uses that?

My bad, I misread your description, never mind.

dash/include/dash/Shared.h
```diff
@@ -168,20 +193,36 @@ class Shared {
...
+    DASH_LOG_DEBUG_VAR("Shared.get", m_owner);
+    DASH_LOG_DEBUG_VAR("Shared.get", m_dart_gptr);
+    DASH_ASSERT(!DART_GPTR_ISNULL(m_dart_gptr));
+    return const_reference(m_dart_gptr);
+  }
+
+  /**
+   * Native pointer to the starting address of the local memory of
+   * the unit that initialized this dash::Shared instance.
+   */
```
devreal, line 206: Please document that `nullptr` is returned on any other unit.

devreal (changes requested):
One comment, otherwise fine with me :+1:

devreal (approved):

PR #581, opened 2018-08-30
For hostnames longer/equal DART_LOCALITY_HOST_MAX_SIZE dash crashes. 

dhinf:
The reason is a wrong string length handed over to gethostname.

### DIFF
--- a/dart-if/include/dash/dart/if/dart_types.h
+++ b/dart-if/include/dash/dart/if/dart_types.h
@@ -12,6 +12,7 @@
 #include <stdlib.h>
 #include <stdint.h>
 #include <stdbool.h>
+#include <limits.h>
 
 #ifdef __cplusplus
 extern "C" {
@@ -325,7 +326,7 @@ typedef enum
 dart_locality_scope_t;
 
 /** Maximum size of a host name string in \ref dart_hwinfo_t */
-#define DART_LOCALITY_HOST_MAX_SIZE       ((int)(30))
+#define DART_LOCALITY_HOST_MAX_SIZE        (HOST_NAME_MAX)
 /** Maximum size of a domain tag string in \ref dart_hwinfo_t */
 #define DART_LOCALITY_DOMAIN_TAG_MAX_SIZE ((int)(32))
 /** Maximum number of domain scopes in \ref dart_hwinfo_t */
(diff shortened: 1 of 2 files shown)


dhinf:
an alternative could be:
```
  if(gethostname(hw.host, DART_LOCALITY_HOST_MAX_SIZE) != 0)
    hw.host[DART_LOCALITY_HOST_MAX_SIZE-1] = '\0';
```
maybe more accurate and safer?

devreal:
For reference, the culprit is hidden in the **Notes** section of the man page:

> Having performed the copy, the function then checks if the length of the nodename was greater than or equal to len, and if it is, then the function returns -1 with  errno  set  to `ENAMETOOLONG`; in this case, a terminating null byte is not included in the returned name.

This breaks with the standard definition :man_facepalming: 

Also from the man page: 
 > POSIX.1 guarantees that "Host names (not including the  terminating  null  byte)  are  limited  to  `HOST_NAME_MAX`  bytes".   On  Linux, `HOST_NAME_MAX` is defined with the value 64, which has been the limit since Linux 1.0.

Our limit currently is 30, which seems arbitrary.

Two things to do here: fix the maximum length of the hostname that we allow by setting `DART_LOCALITY_HOST_MAX_SIZE` to `HOST_NAME_MAX` and, as @dhinf suggested in his comment, ensure that the string is always NULL-terminated.

dhinf:
Our HPC system nodes have hostnames > 30 chars. Changed same weeks ago and the reason i ran into the bug.

devreal (approved):
Thanks for taking care of this!

smutch:
Hi all.  I just tried to build on OS X and ran into the following error:

```
dash/src/dart-if/include/dash/dart/if/dart_types.h:364:16: error: use of undeclared identifier 'HOST_NAME_MAX'
    char  host[DART_LOCALITY_HOST_MAX_SIZE];

dash/src/dart-if/include/dash/dart/if/dart_types.h:331:45: note: expanded from macro 'DART_LOCALITY_HOST_MAX_SIZE'
#define DART_LOCALITY_HOST_MAX_SIZE        (HOST_NAME_MAX)
```

It looks like `HOST_MAX_SIZE` is not defined on OS X and other BSD systems.

fmoessbauer:
Hi @smutch , thanks for your report.
@devreal : We could easily fix that by checking if  `HOST_NAME_MAX` is defined and otherwise defaulting to 256?

devreal:
There seem to be other macros on OS X: https://git.devuan.org/walshb/openssh/commit/a2c95c1bf33ea53038324d1fdd774bc953f98236

We can fall-back to 256 if none of that is available...

PR #583, opened 2018-09-13
Bug dart buddy new mem allocation

tuxamito:


dart-impl/mpi/src/dart_mem.c
```diff
@@ -39,11 +39,18 @@ struct dart_buddy {
 char* dart_mempool_localalloc;
 struct dart_buddy  *  dart_localpool;
 
-static inline int
+static inline unsigned int
```
devreal, line 42: Why not make this `size_t`?

dart-impl/mpi/src/dart_mem.c
```diff
@@ -39,11 +39,18 @@ struct dart_buddy {
...
 num_level(size_t size)
 {
-  unsigned int level  = 1;
-  while ((((unsigned int) 1) << level) < size) {
+  unsigned int level = 1;
+  size_t shifter = 0x02;
+
+  if(size > (9223372036854775808U)) {
```
devreal: What is this error for?

dart-impl/mpi/src/dart_mem.c
tuxamito: Because it is only going to store values [0 - 63], and it the returned value of this function is stored in an unsigned integer that is part of a data structure, that is probably used in some other place... not sure what would be the consequences.

devreal: Ahh, you're right, of course.

devreal (approved):

PR #591, opened 2018-09-21
Rename dash::accumulate -> dash::reduce

devreal:
As agreed upon at the F2F, this PR renames `dash::accumulate` to `dash::reduce`, mimicking the semantics of C++17 `std::reduce`. 

In addition to pure name changes, the changes also include inferring the return type from the iterator's `value_type` (instead of type of the initial value) and fixes some additional inconsistencies (pass by value instead of reference, default value for `init`, superfluous include in `Transform.h`).


fuchsto:
The major difference between `std::reduce` and `std::accumulate` is that `std::reduce` does not guarantee in-order traversal. So this is convenient for DASH because in-order traversal is unsatisfiable whenever element space is partitioned. 
Fine.

But in-order semantics *on the result* can be satisfied with suitable map-reduce schemes, like prefix scan.
So you need both! Semantics of reduce and accumulate are very similar and reduce is ususally the better choice, but this does not eliminate the need for `accumulate`. This is also what the standard committee had to say about this. In C++17, there still is `std::accumulate` for this reason.

I personally am not happy about this situation at all (iteration order should be a policy thing, and it's tightly coupled with parallelisation, where policies already exist, but only for `std::reduce`, not `std::accumulate` because people have the misconception that you cannot parallelize it anyways) but we should honor the standard to avoid adding to the confusion. 

That said, the current implementation of `dash::accumulate` is not in-order anyways.
So I'm fine with this PR, but we should add a _real_ `dash::accumulate` that guarantees in-order reduction.
Your thoughts?

devreal:
> So I'm fine with this PR, but we should add a real dash::accumulate that guarantees in-order reduction.
Your thoughts?

Thanks for your thoughts on this :) We can provide a naive (and slow) implementation of `dash::accumulate` with in-order element traversal. It would be easier to motivate putting effort into a more efficient implementation if we have an actual use-case that requires `accumulate` instead of `reduce` semantics though (just like @rkowalewski put lots of efforts into `dash::sort`, simply because he needed an efficient implementation). I think for most applications `reduce` is sufficient.

fuchsto:
Ah, no, in-order *reduction* is not slow. In-order *traversal* is (and I don't see why we should provide that, `dash::for_each` doesn't, and noone cares).
`std::reduce` assumes commutative reduction operations, so the order of reduction of partial results is not guaranteed.
In-order reduction is to say:  For `binary_reduce_op(a,b)`, `a` is a reduction of elements preceding all elements in reduction `b`.
That's why I mentioned prefix scan as an example. Prefix scan does not require in-order traversal but requires in-order reduction.

rkowalewski:
I am fine with this PR as well, despite we should change few things.

- First I would not "document" how the algorithm works. A two-step process might be a possible implementation but there are also other options.
- I would change the wording from `undefined order` to non-deterministic order, leading to non-deterministic results if the operator is not both commutative and associative
- We shouldn't always deduce the return type from the `iterator_traits`. If the user specifies a `init` value which has a different `value_type` compared to the `iterator_type` we should return a `value_type`. The same is done in C++17.

Finally, one minor thing. I am not a big fan of a default `init` value, I do not like this in C++ either. The reason is that one can easily forget this and the identity is not always the default constructed value. If we want to be "cool" maybe we can specialize a compile time type to deduce the identity based on the operation, if possible. If not possible we require the user to explicitly specify it. That would even be a small proposal for open-std.org IMHO.

rkowalewski (changes requested):
See my previous comment.

devreal:
> First I would not "document" how the algorithm works. A two-step process might be a possible implementation but there are also other options.
> I would change the wording from undefined order to non-deterministic order, leading to non-deterministic results if the operator is not both commutative and associative

Ack, will change that.

> Finally, one minor thing. I am not a big fan of a default init value, I do not like this in C++ either. The reason is that one can easily forget this and the identity is not always the default constructed value.

Fine with me, we can make the `init` argument mandatory. 

> We shouldn't always deduce the return type from the iterator_traits. If the user specifies a init value which has a different value_type compared to the iterator_type we should return a value_type. The same is done in C++17.

I have to disagree here. I don't like the way this is done in C++ as it is just too easy to get wrong. Consider, for example, the following:

```C
std::vector<uint64_t> vec(...);
std::reduce(vec.begin(), vec.end(), 0);
```

You'll be surprised to see what happens with numbers `>UINT_MAX` (the correct way of course would be to invoke `std::reduce(vec.begin(), vec.end(), uint64_t{0});`. The argument you made above about the default `init` being easily wrong applies here as well.

rkowalewski:
> You'll be surprised to see what happens with numbers >UINT_MAX (the correct way of course would be to invoke std::reduce(vec.begin(), vec.end(), uint64_t{0});. The argument you made above about the default init being easily wrong applies here as well.

I ran into this problem actually while implementing `dash::sort` and it was really hard to debug such a bug as it only manifested in medium to large scale tests. Fine with me, but then we have to document this limitation.

rkowalewski (approved):

PR #592, opened 2018-09-21
Git: Ignore all hidden files by default

devreal:
Plenty of tools (gdb, editors, you name it) tend to create hidden files, which are tedious to filter through a whitelist. Just blacklist all hidden files, we can still add them manually.


rkowalewski (approved):

.gitignore
```diff
@@ -22,18 +22,11 @@ dash/include/dash/util/StaticConfig.h
...
 ipm/*/GEN.*
 my.*.sh
-.syntastic*
-.clang*
 tags
 *.orig*
 *.bac*
 *.mine*
 RELEASE/*
-dash/include/dash/util/StaticConfig.h
```
fmoessbauer, line 32: If I remember correctly, this file is generated by @fuerlinger's build scripts, so it should be ignored.

fmoessbauer (commented):
Good Idea.

.gitignore
devreal: Yup, but the entry was duplicated :)

PR #593, opened 2018-09-21
Bump C++ standard version C++14

devreal:
As agreed upon at the F2F, we bump the minimum standard version to C++14. 

Updated minimum compiler versions required:

- GCC: 5.0
- Clang: 3.4
- Intel: 17.0

Still to test: 

- [x] Cray (me)
- [x] IBM  (@rkowalewski)

Note: we won't be able to compile DASH with the default Red Hat compiler, which by now is over 3 years old. We agreed that any halfway decent system should provide a C++14 capable compiler. C++14 is required by the updates to the DASH Halo implementation to be merged with #560. 

@fmoessbauer Any changes required for CI?


devreal:
Fixed a warning that came up after switching to C++14.

rkowalewski:
I will do my task next week.  ;)

devreal:
Update from my side: I tried compiling with CCE 8.7.4 but that already fails with C++11 (unimplemented features and a segfault along the way). It has reported some interesting warnings for which I will post a PR tomorrow. I filed bug reports and will retry periodically (as I have done in the past) but for now we should not consider Cray a roadblock for going to C++14.

pascalj:
I tested this with IBM xlc/xlC and unfortunately the older CMake versions have [problems](https://gitlab.kitware.com/cmake/cmake/issues/17784) detecting it. However, after a small workaround Dash compiles successfully.

fmoessbauer:
@devreal : I updated the containers so that the CI works again. All containers - except minimal - already use ubuntu 18.04 where gcc-7 is the default version. For minimal I now explicitly install gcc-5 However gcc-5 has some issues with PAPI:

```
[  3%] Building C object dart-impl/base/CMakeFiles/dart-base.dir/src/internal/papi.c.o
cd /root/project/build-ci/20181015-184318--uuid-2e4c7a13/gnu/Release/build/dart-impl/base && /usr/bin/gcc   -I/root/project/dart-impl/base/include -I/root/project/dart-impl/base/src -I/root/project/dart-if/include -I/opt/papi/include  --std=c11  -mtune=native -march=native -mfpmath=sse -fopenmp   -D_XOPEN_SOURCE=700 -O3 -DNDEBUG  -Ofast -DDASH_RELEASE -fPIC    -DDART_MPI_ENABLE_UNIFIED_MEMORY_MODEL -DDART_ENABLE_PAPI -std=gnu11 -o CMakeFiles/dart-base.dir/src/internal/papi.c.o   -c /root/project/dart-impl/base/src/internal/papi.c
In file included from /root/project/dart-impl/base/include/dash/dart/base/internal/papi.h:12:0,
                 from /root/project/dart-impl/base/src/internal/papi.c:10:
/opt/papi/include/papi.h:579:7: error: unknown type name 'caddr_t'
       caddr_t pr_off;         /**< pc start address (offset) */
       ^
...
/opt/papi/include/papi.h:834:7: error: unknown type name 'caddr_t'
       caddr_t end;            /**< user requested end address of an address range */
       ^
In file included from /root/project/dart-impl/base/include/dash/dart/base/internal/papi.h:12:0,
                 from /root/project/dart-impl/base/src/internal/papi.c:10:
/opt/papi/include/papi.h:1063:50: error: unknown type name 'caddr_t'
    int   PAPI_profil(void *buf, unsigned bufsiz, caddr_t offset, 
```

https://5025-45930061-gh.circle-artifacts.com/0/tmp/build-logs/minimal/20181015-184318--uuid-2e4c7a13/gnu/Release/build.log

rkowalewski:
As far as I remember this is a known bug in PAPI and should be fixed with a later version (> PAPI 5.5.0). Which version do we use on our CI?

fmoessbauer:
Currently we use [5.4.1](https://github.com/dash-project/dash/blob/d03c4b16e385f6c7cad3d7ce207a8ba14c5d617a/dash/scripts/docker-testing/minimal/dockerfile#L19) but we can lift that to 5.5 easily. Just give me your OK on that ;)

rkowalewski:
@fmoessbauer  you are admin for certain reasons. ;)

devreal:
Upgrading PAPI to 5.5.0 is fine with me.

fmoessbauer:
@rkowalewski : I know, but some decision should be made in a team ;) I'm currently re-building the container, but as openmpi is build after papi, I have to rebuild openmpi... Get a coffee and in the meantime the CI should work again.

fmoessbauer:
We have to merge with admin privs, as the patch target is not hit... but there are no code-related changes, so we actually cannot reach it. 
**Edit**: Everyone can merge, patch is not mandatory.

fmoessbauer (approved):

devreal:
Looks good to me now, thanks for taking care of CI :+1:

PR #594, opened 2018-09-21
Add dash::bcast

devreal:
Simple implementation of `dash::bcast` with overloads for `dash::Shared` as well as local ranges. The latter also supports non-contiguous ranges (e.g., `std::list`) by creating a temporary contiguous buffer and broadcasting from and into it.


fmoessbauer:
We should add support for the `dash::Coarray`. This should be quite simple by calling `dash::coarray::cobroadcast`: https://github.com/dash-project/dash/blob/development/dash/include/dash/coarray/Utils.h#L142

devreal:
Update: I renamed `dash::bcast` to `dash::broadcast` (to better reflect the way C++ STL functions named vs MPI functions) and added support for Coarray in `dash::broadcast`.

devreal:
For the record: the coverage test fails because someone somewhere thought it's a good idea to print out arbitrary types that may or may not be printable if logging is turned on in DASH. I may or may not fix this at some point...

fmoessbauer:
@devreal : Agree, this has lead to many broken runs... Fixed it.

devreal:
@fmoessbauer I was working on a type-trait to figure out whether we can print the value: 

```C
template<
  class StreamType,
  class ValueType>
struct is_printable
{
private:
  template<typename, typename>
  static constexpr std::false_type test(...);

  template <typename T=StreamType, typename U=ValueType>
  static constexpr
  decltype(operator<<(std::declval<T>(), std::declval<U>()), std::true_type{}) test(int);

public:

  static constexpr bool value = decltype(test<StreamType, ValueType>(0))::value;
};
```

I'm not there yet as I keep seeing compiler errors. I might give this one or two more tries :)

fmoessbauer:
@devreal : I also thought about a trait, but IMO it's not worth it. If you want to give it a try, just revert my commit and I'm happy with it ;)

rkowalewski:
I would rather change the way how our logging scripts work since it is anything but efficient. ;)
I mean similar to [this here](https://github.com/dash-project/dash-apps/blob/master/splash2/barnes-hut/Logging.h). Then you can write simple logging statements  like:
~~~
LOG("mynleaf: " << Local.mynleaf << ", myncell: " << Local.myncell);
~~~

The advantage is that we do not need the trait (traits are sometimes the diplomatic wording for C++ TMP hack). We have more flexibility, better performance compared to our current Logging approach (yes, our slow logging sucks). And it is a separate issue anyway. 
Using this kind of solution reduces the generated assembly to around a half the size of the original logging macro on Godbolt with gcc 8.2.

EDIT: Yes, changing it completely would break the whole codebase. We can keep the macros as they are right now and redirect them to the new approach in order to allow a smooth transition.

EDIT2: The [original source](https://stackoverflow.com/questions/19415845/a-better-log-macro-using-template-metaprogramming) where I have stolen it for a more detailed discussion.

rkowalewski (dismissed):
Fine for me now.

rkowalewski:
Actually I wanted to approve a different pull request (#591)

devreal:
Update: Good news! I came up with a way to distinguish between printable and non-printable types by introducing an overload in `Logging.h`. Yay!

The bad news: GCC (tested up to 8.2) fails to find our custom `operator<<`s that handle ranges (defined in `StreamConversion.h`) in the `decltype` statement used in SFINAE'ing the overload. Boooh!

As a result, printing ranges (such as the `_extent` array of a team) would not work anymore (it will compile but print something along the lines of `UNKNOWN` as the wrong overload is chosen). I suggest we stick to removing the output of the member value in `dash::Shared`. I will file a bug report against GCC.

@rkowalewski Can you file an issue to track/discuss the issues with the logging system in DASH?

devreal:
GCC Bug filed here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=87452

rkowalewski:
Logging issue is tracked in #598. Ready to merge?

fmoessbauer (approved):
@devreal : After resolving the conflict, IMO we can merge ;)

PR #595, opened 2018-09-28
Minor fixes to DASH CMake files

devreal:


rkowalewski:
Looks good. But just to get it. What is the issue, why did we need these "fixes"?

devreal:
Sorry, should have mentioned that: The fix to the MPI CMake  file is merely to chose the right flag to suppress inclusion of the C++ wrappers for Intel MPI. 

For threads, including the `-pthread` flag may not work on all compilers, e.g., the Cray compiler does not know that flag but instead includes pthreads by default. The patch only sets these flags if the CMake module deems it as required.

rkowalewski (approved):

PR #596, opened 2018-10-02
Fix problems reported by Cray Compiler

devreal:
This PR fixes a bunch of problems reported by the Cray compiler (seen while working on  #593):

- Replace usage of `std::pair` with `dash::Pair`, which is a PODType and thus allows for `offsetof` to be used
- Remove `-Wno-unsed` from the `dash/` directory as it should a) not be set outside of `CMakeExt/Compiler.cmake` and b) is not supported by all compilers; This came with a larger set of fixes to remove unused variables.
- Fix some other minor issues

### DIFF
--- a/dart-impl/base/src/logging.c
+++ b/dart-impl/base/src/logging.c
@@ -22,7 +22,7 @@
 /* Width of line number field in log messages in number of characters */
 #define LINE_WIDTH 4
 /* Maximum length of a single log message in number of characters */
-#define MAX_MESSAGE_LENGTH 256;
+#define MAX_MESSAGE_LENGTH 256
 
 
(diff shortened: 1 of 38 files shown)


devreal:
Gna, `dash::Pair<Key, Value>` is not trivial if Key is `const` (as is the case for `UnorderedMap::value_type`). So that part of the PR doesn't fix the warning issued by the Cray compiler...

fmoessbauer:
Is this PR supersede by #600? I guess that there will be many conflicts between these two.

devreal:
Yes, there is definitely overlap that will create conflicts. I will go over it again and check whether there are fixes in here that should be included in #600 too.

devreal:
Superseded by #600, closing

PR #599, opened 2018-10-18
Remove enforcement of C++11 to enable C++14

rkowalewski:
Since #593 did not fully enable C++14 due to CI "hacks". Here is at least a "fix" which works. Not sure if everything is fine now. Our CMake scripts do not really look CMake compliant.


devreal (approved):
Looks good!

fmoessbauer (approved):

PR #600, opened 2018-10-23
Enable -Wall and -Werror in developer builds

rkowalewski:
I did lots of fixes to enable `-Wall` and `-Werror` in our development builds. It took not too many changes. However, some things got broken:
- Cotire build
- LAPACK integration

Our `CompilerFlags.cmake` is really a mess and I invite all maintainers to join this effort, especially for a proper cleanup of our compiler flags. As an example, we have both `-std=c++14` and `-std=gnu++14`. If you are looking into the compilation commands there are further unintentional details. Do we need `-march=native` and `-mtune=native`?


dart-impl/base/include/dash/dart/base/mutex.h
```diff
@@ -111,7 +111,7 @@ dart__base__mutex_trylock(dart_mutex_t *mutex)
 #ifdef DART_HAVE_PTHREADS
   int ret = pthread_mutex_trylock(&mutex->mutex);
   DART_LOG_TRACE("dart__base__mutex_trylock: lock %p aqcuired: %s",
-                 mutex, (ret == 0) ? "yes" : "no");
+                 (const void *)mutex, (ret == 0) ? "yes" : "no");
```
devreal: Why do we need the explicit cast here (and elsewhere)?

dart-impl/mpi/src/dart_communication.c
```diff
@@ -701,7 +701,7 @@ dart_ret_t dart_accumulate_blocking_local(
 
   CHECK_UNITID_RANGE(team_unit_id, team_data);
 
-  DART_LOG_DEBUG("dart_accumulate() nelem:%zu dtype:%ld op:%d unit:%d",
+  DART_LOG_DEBUG("dart_accumulate() nelem:%zu dtype:%ld op:%ld unit:%d",
```
devreal, line 704: Here `op` is printed as `%ld` where above it is printed as `%p`.

dart-impl/mpi/src/dart_mpi_op.c
```diff
@@ -133,7 +135,7 @@ static void dart__mpi__op_invoke_custom(
 {
   struct dart_operation_struct *dart_op = get_op(*dptr_);
   DART_LOG_TRACE("Invoking custom operation %p (op=%p, ud=%p)",
-                 dart_op, dart_op->op, dart_op->user_data);
+                 (const void *)dart_op, (const void *)(uintptr_t)dart_op->op, dart_op->user_data);
```
devreal: Why do we need the intermediate cast to `uintptr_t`?

devreal (commented):
Thansk for taking the time to work on this. Some of the fixes were already in #596 but I'm fine with superseding this PR. 

I agree that out `CompilerFlags.cmake` is a mess and I thought about some ways of cleaning it up. Something along the lines of a table containing all the flags for the different compilers could be feasible. 

Some comments/questions inline.

dart-impl/mpi/src/dart_mpi_op.c
rkowalewski: If we do not apply the intermediate cast, the compiler complains that you cannot cast from a smaller type to a different larger size. This way, the compiler eats it.

dart-impl/base/include/dash/dart/base/mutex.h
rkowalewski: The standard says (§ 7.19.6.1):
`p`  The argument shall be a pointer to void. The value of the pointer is converted to a sequence of printing characters, in an implementation-defined manner.

rkowalewski: The compiler prefers to have it explicit.

dart-impl/mpi/src/dart_communication.c
rkowalewski: OK done. ;)

rkowalewski:
An update to the locality stuff in DART. Since it has some `const` -> `non-const` cast issues I disable these warnings for the relevant code lines. And since it is a hack I explicitly named the file `compiler_tweaks.h`. This is nothing to adapt for other issues and should be never adapted in any other situations. However, the annotated code takes some time and requires probably to refactor internal APIs.

dart-impl/base/include/dash/dart/base/mutex.h
devreal: This seems overly strict to me and makes printing a pointer `-Wall`-clean ugly and cumbersome. Implicit casts to and from `void*` are perfectly valid in C, e.g., `malloc` and `free`, and afaics it does not make any difference for `printf`'s `%p` what type the pointer has. Are all compilers warning about this? 

IMO, unless there is a really good reason for enforcing these casts we should rather suppress them.

dart-impl/mpi/src/dart_mpi_op.c
devreal: Ahh I see, it might be better then to consistently print `dart_op_t` as `%ld` (it is an integer after all, right?)

dart-impl/base/include/dash/dart/base/mutex.h
rkowalewski: We can suppress this warning with the proper compiler flag. Then we can remove all these casts.

dart-impl/mpi/src/dart_mpi_op.c
rkowalewski: Agree. Since we define `dart_operation_t` we can print it as `%ld`. What I do not really like is that we often log opaque MPI types as well and we cannot know how to properly format them. I suggest we never log opaque MPI types.

dart-impl/base/include/dash/dart/base/mutex.h
devreal: :+1:

devreal:
One thing I noticed over at #596: We're setting ` -Wno-unused -Wno-sign-compare` in `dash/CMakeLists.txt`, which may break for compilers not supporting these flags. This PR sets `-Wno-sign-compare` which is fine for me. Instead of suppressing `-Wunused` we should clean up our code and get rid of unused variables/functions.

devreal:
I ported a minor fix from #596. The rest over there is fixing unused variables, see my previous comment.

rkowalewski:
I repeatedly applied `clang-tidy` and the more I experiment with it the more I like it. While there are many false positives certain checks like `modernize-*` and `cppcoreguidelines-*` work without any conflicts and are definitely good changes. Will update our internal documentation and propose an updated `clang-tidy` configuration file.

rkowalewski:
Yes, sorting includes, etc. are no functional improvement but considered as best practices and, for example, big projects like Google Chromium do this too. Usually, this thing is part of the build stage.

> This has been a policy I simply adopted and changing it does not provide any improvement but makes tracking changes hard.

Which policy?

Anyway, I will revert it. Thank you Git.

dash/include/dash/Team.h
```diff
@@ -76,12 +77,12 @@ class Team
...
   inline iterator begin()
   {
-    return iterator(0);
+    return {0};
   }
 
   inline iterator end()
   {
-    return iterator(size());
+    return {static_cast<int>(size())};
```
ddiefenthaler, line 85: This construction looks a little bit unintuitive.

dash/src/util/BenchmarkParams.cc
```diff
@@ -189,7 +189,9 @@ void BenchmarkParams::print_pinning()
       << std::setw( 5)     << "core"
       << std::setw( 5)     << "CPU"
       << '\n';
-  for (global_unit_t unit{0}; unit < dash::size(); ++unit) {
+  for (global_unit_t unit{0};
+       unit < static_cast<dash::global_unit_t>(dash::size());
```
ddiefenthaler, line 193: This cast does not look right.

ddiefenthaler (commented):
The CompilerFlags.cmake file is simplified quite a bit.

Currently there are some Warnings-flags disabled in the normal developer flags.
This is to allow the debug-build to pass with `-Werror` enabled.
For example with `-Werror=pedantic` and C99, DART will not build.
With the `EXT_COMPILER_WARNINGS` these warnings will be activated, and `-Werror` will be disabled.

The rest of the Pull Request looks good, with some comments inline.
Some minor indentation problems are already fixed.
I'm also not too sure about pure layout changes like [here](https://github.com/dash-project/dash/pull/600/files#diff-07de7d1a5858b6f4d5003bb11fbf373eR63) for example.

dash/src/util/BenchmarkParams.cc
rkowalewski: The problem here is actually that `dash::size()` returns a `size_t` while `dash::team_unit_t` and `dash::global_unit_t` are 16-bit signed integers. So to fix that we should return an unsigned 16-bit integer from `dash::size`. However, this is a separate pull request.

rkowalewski:
Thank you @ddiefenthaler

dash/include/dash/Team.h
rkowalewski: why? It is actually the same "problem" as explained below.

ddiefenthaler: I wasn't refering to the cast here.
I meant the implicit list initialization of the iterator.
The `begin()`-method has the same notation.

After I found the appropriate explanation on cppreference this looks fine,
while perhaps still a little bit unintuitive.

dash/src/util/BenchmarkParams.cc
ddiefenthaler: Ok, but such a change to `dash::size` sounds like it should be very well considered.

dash/include/dash/Team.h
rkowalewski: This unintuitive thing is called uniform initialization. Yes, it is not always straightforward but that ugly syntax is to what the C++ gurus agreed on.

ddiefenthaler (approved):

dash/src/util/BenchmarkParams.cc
devreal: Actually, the unit IDs in DASH are 32-bit (we currently only have 24bit in the `dart_gptr_t` iirc because of the flags field that is not used). But I agree, we should consider fixing this discrepancy for 0.4.0.

devreal:
Thanks for the good work, y'all!

PR #602, opened 2018-10-30
Move clang-format file into root directory

devreal:
This PR will make our `.clang-format` rules more visible and easier to use (`clang-format` is looking for that file in the current directory or any parent directory)


fmoessbauer (commented):
Another two files in the root directory... For what reason did we have the contrib folder?
@rkowalewski : I guess there was once a discussion to move the dev-only stuff to this folder?

devreal:
Right now the `contrib` folder only contains these two files, hidden. For new users (or developers not aware that we have a `clang-format` stylefile) it's virtually invisible...

rkowalewski:
> I guess there was once a discussion to move the dev-only stuff to this folder

Yup. The original intent was that we do not want to "pollute" our root directory with files which are not a required part of our toolchain. I am also not aware that everybody uses `clang-format`. So I used this approach as other projects do it in a similar way. We deleted also the `.syntastic_*`files for that reason.

But for me it's fine that we move it to the root directory and we should definitely use a common formatting. @pascalj and me are using this one in the repository and it is quite pretty. If you have some suggestion feel free to change it.

rkowalewski:
In my opinion we can move some of the `build.*` files into the contrib folder. The most relevant are `build.minimal.sh`, `build.dev.sh` and `build.sh`. I think almost nobody uses the other files.

rkowalewski:
small update to our `.clang-format`. The settings are based on Google and I just added the relevant customizations for our code base. This way, our config is easier to maintain and gets new features without explicitly adding it in our config.

PR #603, opened 2018-11-02
Add dash::Future::valid

devreal:
Used to check whether a future has been properly initialized (which may not be the case in a `std::vector<dash::Future>`). This is used in the DASH port of Lulesh.

### DIFF
--- a/dash/include/dash/Future.h
+++ b/dash/include/dash/Future.h
@@ -12,20 +12,41 @@
 
 namespace dash {
 
+/**
+ * Implementation of a future used to wait for an operation to complete
+ * and access the value returned by that operation.
+ */
 template<typename ResultT>
 class Future
 {
-private:
+public:
   typedef Future<ResultT>                self_t;
+
+  /**
+   * Callback function returning the result value.
+   */
   typedef std::function<ResultT (void)>  get_func_t;
+
+  /**
+   * Callback function to test for the availability of the result value.
+   */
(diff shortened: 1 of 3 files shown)


dash/include/dash/Future.h
```diff
@@ -112,6 +111,11 @@ class Future
     return _value;
   }
 
+  bool valid()
+  {
+    return (_ready || _get_func != nullptr);
```
fmoessbauer, line 214: What is the semantic of `_ready`? Is it possible to have `_ready==false` and `_get_func != nullptr`? Please document the semantics and meaning of the private members.
... and the public interface seems to be undocumented too.

fmoessbauer (changes requested):
While this should fix the issue Douglas had, we should use this and add documentation.

fmoessbauer:
... and a unit test for `valid` is missing as well. Sorry for being pedantic, but missing tests have cause so much trouble in the past.

dash/include/dash/Future.h
devreal: That is why I start to hate putting up PRs. There is documentation missing everywhere so "just fixing something" really isn't possible... Will add documentation though.

fmoessbauer (approved):
Thanks for the documentation and the unit test. This looks pretty good now ;)

rkowalewski (approved):
Fine for me as well

douglaspasqualin:
Thank you guys. I'll test again.

PR #605, opened 2018-11-08
Last preparations for 0.3.0

devreal:
This PR includes the GoogleTest in the generated release tarball and adjusts the GoogleTest integration accordingly. It also disables formatting tests for non-developer tests.


rkowalewski:
OK thank you. Not sure if we can simply check out `GoogleTest` from the perl script itself. But for now, fine. Thank you for these fixes.

rkowalewski (approved):

PR #607, opened 2018-11-15
Feature dash::Vector

fabian-jung:
Vector implementation file dash/include/dash/vector
Example file ex.12.vector
Test container/VectorTest
Added the include directive for dash::vector to the dash/Container.h meta header


dhinf:
Not ready for development yet.

PR #608, opened 2018-11-19
Bump DASH release number to 0.4.0

devreal:
Bumping internal version number to 0.4.0 after the 0.3.0 went through.


fmoessbauer (approved):

devreal:
One thing I realized later: the current default installation in `build.sh` and friends is `~/opt/dash-0.3.0/`, should we change that to 0.4.0, too? It might lead to confusion for people relying on this path...

fmoessbauer:
@devreal : Good point. Let's change that too.

rkowalewski (approved):

PR #609, opened 2018-11-19
Add information to BenchmarkParams

devreal:
This PR adds the current DASH version and git commit hash as well as the compiler version to the benchmark output. Having the git commit hash in the benchmark output is helpful to investigate possible performance regressions from older benchmark runs.

This PR also moves the definition of the DASH version to `dash/Version.h` and thus would supersede #608.


devreal:
The version went back into the CMake file as it is needed for the doxygen documentation. The `Version.h` is now generated from the information in the CMake file (and the git hash was added there).

fmoessbauer:
@fuerlinger : I suppose this change breaks your Makefiles. Can you provide an updated version?

fuerlinger:
Yes, I'll update my makefiles. Should not be a problem.

rkowalewski:
OK cool. Exactly what I need now for another benchmark session. :)

rkowalewski (approved):

rkowalewski:
What is with this PR? Can we merge it?

devreal:
Forgot to merge, done.

PR #610, opened 2018-11-21
constexpr compile errors with ICC 2019

rkowalewski:
- `constexpr` lambdas are not allowed before C++17, see this [proposal](http://open-std.org/JTC1/SC22/WG21/docs/papers/2015/n4487.pdf)
- `constexpr` of `RegionSpec` makes no sense since not all members can be constructed by a `constexpr`

### DIFF
--- a/dash/test/memory/GlobStaticMemTest.cc
+++ b/dash/test/memory/GlobStaticMemTest.cc
@@ -178,7 +178,7 @@ TEST_F(GlobStaticMemTest, MakeUniqueMoveSemantics)
   constexpr size_t lcapA = 20;
   constexpr size_t lcapB = lcapA / 2;
 
-  constexpr auto cap_in_bytes = [](size_t lcap) {
+  auto cap_in_bytes = [](size_t lcap) {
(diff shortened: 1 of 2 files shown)


dash/test/memory/GlobStaticMemTest.cc
```diff
@@ -178,7 +178,7 @@ TEST_F(GlobStaticMemTest, MakeUniqueMoveSemantics)
   constexpr size_t lcapA = 20;
   constexpr size_t lcapB = lcapA / 2;
 
-  constexpr auto cap_in_bytes = [](size_t lcap) {
+  auto cap_in_bytes = [](size_t lcap) {
```
devreal: I guess `const` would make sense here though

rkowalewski: All right, done.

rkowalewski:
Can we merge this? I need this fix. ;)

pascalj:
Yes, this also resolves issues with the current GCC.

pascalj (approved):
LGTM

PR #611, opened 2018-11-22
[WIP] Boost dash::sort with even more parallelism

rkowalewski:
We need more parallelism to exploit the power of many-core nodes. The underlying algorithm itself will be rewritten to eliminate barriers. This includes the following major changes:

- Algorithmic improvements:
  - [x] reduce communication overhead (`alltoall` communications)
  - [x] overlap communication and the final merge step as efficiently as possible (@pascalj)
  - [ ] Instead of perfect partitioning we provide another variant where we do not require an in-place sort. Users can provide a larger output buffer with the same pattern, but each unit has a certain threshold of additional local storage. Example: `dash::sort(first, last, out, sort_hash)`.

- Minor changes
  - integrate Intel Parallel STL (based on Intel TBB) into DASH to exploit shared memory parallelism more efficiently.
  - Allow `DART_UNDEFINED_UNIT_ID` as a valid unit for DART communication routines #617 
  
- Further Impacts
  - Threadsupport is now enabled by default in CI. Other tests fail, needs some investigation.

Note: This list will grow.


rkowalewski:
> We definitely need a way to configure the number of threads from the outside at runtime, e.g., through an environment variable. That inevitably leads to the wider question of how we want to handle runtime configuration (config files? env variables? both? who is in charge of the parsing? right now DART and DASH both do their own thing but that is sub-optimal...)

Currently we support three env variables to configure multi-threading which is built-in in our locality stuff. See the [documentation](https://github.com/dash-project/dash/blob/development/dash/include/dash/util/UnitLocality.h#L306) of `UnitLocality.num_domain_threads()`:

- `DASH_DISABLE_THREADS`: If set, disables multi-threading at unit scope and this method returns 1
- `DASH_MAX_SMT`: If set, virtual SMT CPUs (hyperthreads) instead of physical cores are used to determine availble threads.
- `DASH_MAX_UNIT_THREADS`:  Specifies the maximum number of threads available to a single unit.

I suppose this is also built into DART somehow since the locality interface is implemented down there.

PR #612, opened 2018-11-26
Fix iteration over deallocators in dash::Team

devreal:
List iterators are invalidated if the element they are pointing at is removed, leading to deallocators being dropped and containers trying to deallocate memory upon destruction, potentially after `dash::finalize`.

### DIFF
--- a/dash/include/dash/Team.h
+++ b/dash/include/dash/Team.h
@@ -314,11 +314,12 @@ public:
   void free()
   {
     DASH_LOG_DEBUG("Team.free()");
-    for (auto dealloc = _deallocs.rbegin();
-         dealloc != _deallocs.rend();
-         ++dealloc) {
+    std::list<Deallocator>::iterator next = _deallocs.begin();
+    std::list<Deallocator>::iterator dealloc;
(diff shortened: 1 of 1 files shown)


rkowalewski:
Is there a test case or situation where it lead to problems? At a first look it is not completely clear to me how your code differs from the original one. And since we are still in a single-threaded environment (at least in DASH) no errors should have been occured at all, right?

Anyway we should refactor the handling of `dash::Team` and their relationship with containers regarding memory deallocation in a future release.

devreal:
@rkowalewski  The problem in the original code is that the `dealloc` iterator gets invalidated because the deallocator deregisters itself from that very same list (see `dash::Array::deallocate` for example, matrix does the same). My solution is to get the next iteration's deallocator (well, the iterator pointing to it) before calling the current deallocator. 

I found this while running Lulesh (which allocates ~26 DASH containers, for whatever reason). It didn't occur previously so I guess this was triggered by some of the changes in the memory management done prior to the 0.3.0 release.

And yes, it's messy and we should fix it at some point.

rkowalewski (approved):
Ah, got it. LGTM.

PR #629, opened 2019-02-21
FIX: Remove team dealloc functions in move assignment and construction

rkowalewski:
After move assignment and move construction in `dash::Array` we should also unregister the dealloc functions in the original team. Otherwise, we get segmentation faults during `dash::finalize`.

### DIFF
--- a/dash/include/dash/GlobPtr.h
+++ b/dash/include/dash/GlobPtr.h
@@ -426,7 +426,7 @@ public:
   }
 
   constexpr auto get_unit() const noexcept {
-    return m_dart_pointer.unitid;
+    return team_unit_t{m_dart_pointer.unitid};
   }
 
   /**
--- a/dash/include/dash/GlobPtr.h
+++ b/dash/include/dash/GlobPtr.h
@@ -425,6 +425,10 @@ public:
     m_dart_pointer.unitid = unit_id.id;
   }
 
+  constexpr auto get_unit() const noexcept {
+    return m_dart_pointer.unitid;
+  }
+
   /**
    * Check whether the global pointer is in the local
    * address space the pointer's associated unit.
--- a/dash/include/dash/GlobPtr.h
+++ b/dash/include/dash/GlobPtr.h
@@ -98,7 +98,7 @@ public:
   /**
    * Constructor, specifies underlying global address.
    */
-  explicit constexpr GlobPtr(dart_gptr_t gptr)
+  explicit constexpr GlobPtr(dart_gptr_t gptr) DASH_NOEXCEPT
     : m_dart_pointer(gptr)
   {
   }
@@ -106,7 +106,7 @@ public:
   /**
    * Constructor for conversion of std::nullptr_t.
    */
-  explicit constexpr GlobPtr(std::nullptr_t)
+  explicit constexpr GlobPtr(std::nullptr_t) DASH_NOEXCEPT
     : m_dart_pointer(DART_GPTR_NULL)
   {
   }
@@ -143,7 +143,7 @@ public:
               typename dash::remove_atomic<value_type>::type>::value>
 
       ::type>
-  constexpr GlobPtr(const GlobPtr<From, GlobMemT> &other)
+  constexpr GlobPtr(const GlobPtr<From, GlobMemT> &other) DASH_NOEXCEPT
     : m_dart_pointer(other.m_dart_pointer)
   {
   }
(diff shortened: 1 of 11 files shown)


dash/include/dash/iterator/GlobIter.h
```diff
@@ -321,6 +321,10 @@ class GlobIter : public std::iterator<
       DASH_LOG_TRACE_VAR("GlobIter.dart_gptr", idx);
       DASH_LOG_TRACE_VAR("GlobIter.dart_gptr", offset);
     }
+
+    if (offset == 1) {
```
devreal: What is this for?

rkowalewski: Good question. In general, the implementation of `GlobIter` is really awkward and we cannot simply fix it because many other places in our code rely on this. Will invest more research to figure it out. The whole thing looks like broken state machine IMHO.

rkowalewski: Will push more changes later today / tomorrow.

rkowalewski: Should look better now. I removed duplicate code.

rkowalewski:
While fixing this stuff here I noticed that `GlobIter.operator[](index_type gidx)` are implemented wrong.

Given an instance `GlobIter<...> it`:
1. `operator[](index_type n)` returns `*(BEGIN + n)`. Should be changed to `*(it + n)`.
2. `operator*` redirects to `operator[]`. So if we fix the previous issue this implementation is either broken.

I can fix it tomorrow. But as always, a first attempt lead to lots of broken tests.

EDIT: "Fixed" now as well.

dash/include/dash/GlobPtr.h
```diff
@@ -425,6 +425,10 @@ class GlobPtr {
     m_dart_pointer.unitid = unit_id.id;
   }
 
+  constexpr auto get_unit() const noexcept {
```
devreal: Why not just `unit()`? As in `Team::unit()`?

rkowalewski: I prefer your suggestion as well, but...`set_unit` already existed. So I decided this to be consistent. We can rename it, but this will certainly impact more source files.

devreal: Any idea what `set_unit` is used for? Conceptually, it exposes a low-level detail (one element of the tuple `[unit, segment, offset]`) to change from the outside, without touching the segment ID or the offset, which leaves you in a potentially inconsistent state. Maybe we should get rid of `set_unit` entirely?

rkowalewski: Well semantically it is valid to change the unit of a particular segment. Given a `GlobPtr` to the beginning of a segment we can always "jump" to the next unit. We use it exactly for this.

On the other hand it enables to modify the pointer to point to Mordor leaving you alone in hell. But this is what we can do in good old C/C++ as well.

dash/include/dash/matrix/internal/Matrix-inl.h
```diff
@@ -478,7 +479,8 @@ constexpr typename Matrix<T, NumDim, IndexT, PatternT, LocalMemT>::const_iterato
 Matrix<T, NumDim, IndexT, PatternT, LocalMemT>
 ::end() const noexcept
 {
-  return const_iterator(_begin + _size);
+    return const_iterator(
+        const_cast<GlobMem_t *>(&_glob_mem), _pattern, _begin.pos() + _size);
```
devreal, line 483: Why not use the copy-constructor here? (and above)

dash/include/dash/GlobRef.h
```diff
@@ -145,7 +145,7 @@ class GlobRef
   /**
    * Move Assignment: Redirects to Copy Assignment
    */
-  self_t& operator=(self_t&& other) {
+  self_t& operator=(self_t&& other) noexcept {
```
devreal, line 148: In other places you use `DASH_NOEXCEPT`. Is there a difference?

dash/include/dash/matrix/internal/Matrix-inl.h
rkowalewski: Because we cannot assign a `const_iterator` to an `iterator`. They are two different types and also the template parameters are different. Actually, const correctness is still a burden because we do not really implement it the right way.

dash/include/dash/GlobRef.h
rkowalewski: There is a difference. `DASH_NOEXCEPT` is only `noexcept` if we are building in release mode. I implemented it for `GlobIter` now and we should refactor this in a separate PR to be consistent in all classes.

devreal (approved):

Issue #615, opened 2018-11-28
Thoughts on NUMA issues with DASH containers


devreal:
I want to lay out some observations I made during the last couple of months.

**Background**

MPI implementations register the memory backing windows with the network device on systems that use high-performance networks in order to leverage RDMA capabilities (observed both on IB and Cray Aries with Open MPI). As part of this process, all pages in the registered memory range are pre-faulted (in the case of `MPI_Win_allocate`), which is done by the thread allocating the window (usually the master thread). Afterwards the pages are non-movable, non-swappable, and subsequently pinned to the NUMA domain on which the thread was running. It is my understanding that pages that are already allocated (e.g., for memory passed to `MPI_Win_create`) are not moved and remain on the NUMA domain on which they were allocated.

I am strongly assuming that other communication libraries such as GASPI do the same.

**The Problem**

This can become a problem for two reasons:

1) Open MPI allocates shared memory on each node running more than one process to speed-up node-local communication (similar to our shared memory optimization). We found that the root-process on the node registers the complete shared memory range, which leads to NUMA effects on all processes not running on the same socket. There is an issue open to fix this so this problem will go away eventually but it is something to keep in mind for the time being (other MPI implementations might be affected as well). 

2) What is more interesting is a scenario in which we have one process per node that uses threads to exploit multi-core performance. Here, the master thread will allocate and register the window memory for the whole window (and thus the DASH container), essentially performing the *first touch* and thus nullifying any attempt to be NUMA-aware in the OpenMP part. This is a severe problem as most of our memory comes from MPI windows.

**Possible Mitigation**

We need to give DASH users a chance to properly initialize the memory of the container before that memory is registered with the network device. That means that we have to a) allocate memory ourselves, b) expose the memory to the user, and c) pass that memory into `MPI_Win_create` eventually (instead of using `MPI_Win_allocate` in the first place). 

DASH containers today support allocation on construction and delayed allocation (both of which directly allocate the window). We could add a third option to allow the construction of the container using local memory, after which only local access is allowed (throwing an exception on any attempt to access global memory). Thus the user would be able to use `.local` and `.lbegin()` on the container (or block/view/whathaveyou). Eventually, the local memory is committed into global memory (by creating the window) and global memory access is possible from that point on. Here is an example:

```C
dash:Array<double> arr;
arr.allocate_local(LARGE_N); // uses malloc/MPI_Alloc_mem
#pragma omp parallel for
for (int i = 0; i < LARGE_N; ++i) {
[...]

PR #630, opened 2019-02-28
Rework Container Type Traits and GlobRef Conversion Rules

rkowalewski:
Unfortunately, we cannot implement the following example:

~~~cpp
struct Parent {
  int x;
};
struct Child : public Parent {
  int y;
};
static_assert(dash::is_container_compatible_v<Child>:, "FAILS");
//The reason is...
static_assert(std::is_standard_layout_v<Child>, "FAILS");
//... see https://en.cppreference.com/w/cpp/named_req/StandardLayoutType for more details, 
// especially the example with class inheritance.

// At the same time,
static_assert(std::is_trivially_copyable_v<Child>, "PASSES");
~~~

AFAIK we chose these constraints because we want to be able to do a simple `memcpy` in DART by using Shared Memory Windows. But `std::memcpy()` does not require a [StandardLayoutType](https://en.cppreference.com/w/cpp/named_req/StandardLayoutTyp). A [TrviallyCopyable](https://en.cppreference.com/w/cpp/named_req/TriviallyCopyable) type suffices as [documented](https://en.cppreference.com/w/cpp/string/byte/memcpy). So we should weaken our container requirements to allow a `dash::Array<Child>`.

We reworked `GlobRef` some time ago and the relatively strict requirements lead to one consequence which we have to deal with. We should permit the following conversion:

~~~cpp
  dash::Array<Child> array{100};

  //...Initialize with whatever

  // This is a child and we read 8 bytes...
  auto asChild = array[10].get();
  /*
   * Here we explicitly cast it as Parent. In consequence, we read only 4
   * bytes (i.e., sizeof Parent).
   */
  auto asParent = static_cast<dash::GlobRef<Parent>>(array[10]).get();
~~~

EDIT: My suggestion is to replace `is_standard_layout` with `is_trivially_default_constructible`. Then, a `Child` is container compatible and this should suffice for our requirements. Thoughts on this?


devreal:
> AFAIK we chose these constraints because we want to be able to do a simple `memcpy` in DART by using Shared Memory Windows. But `std::memcpy()` does not require a [StandardLayoutType](https://en.cppreference.com/w/cpp/named_req/StandardLayoutType).

There are two different things: we need `is_trivially_copyable` for *any* copying in and out of the DASH containers (has nothing to do with shared memory, we just cannot run an assignment operator when copying from unit A to unit B). The restriction on `is_standard_layout` did not come from me (in fact, I don't remember having seen it before :D) but there seems to be a good reason: 

From https://en.cppreference.com/w/cpp/named_req/StandardLayoutType:
> Has no virtual functions or virtual base classes

This restriction is important as we cannot copy vtables across process boundaries (unless we force things like ALSR to be disabled). Afaics, it doesn't include `is_trivially_default_constructible` but that is important since we don't run c'tors on the elements in the container (maybe I'm missing that in the definition of the standard layout type?)

So, to amend your proposal: `!is_polymorphic<T> && 
is_trivially_default_constructible<T> && is_trivially_copyable<T>` ?

rkowalewski:
> There are two different things: we need `is_trivially_copyable` for _any_ copying in and out of the DASH containers (has nothing to do with shared memory, we just cannot run an assignment operator when copying from unit A to unit B).

That is what I meant, actually. In other words, we want to keep compatibility to C, MPI, etc.

> Has no virtual functions or virtual base classes.

Correct. vtables shouldn't be allowed. But a standard layout gives a much stronger guarantee. We can include `! is_polymorphic<T>`, however, we do not really need it. A polymorphic class is never trivially copyable.

devreal:
You're right, `is_trivially_copyable` includes `!is_polymorphic`. Please ignore `!is_polymorphic` then :)

rkowalewski:
Minor change: We cannot rely on `is_trivially_default_constructible` but only on `is_trivially_copyable`. As an example, `is_trivially_default_constructible` does not permit default initialization. This also excludes `dash::Pair` which has been valid until now.

And after thinking again we do not really need trivial construction. All we need is default construction which gives us still the possiblity to construct elements while constructing the container, although we do not really consider this currently.

So in summary:

~~~cpp
template <class T>
struct is_container_compatible :
  public std::integral_constant<bool,
              std::is_default_constructible<T>::value
              && std::is_trivially_copyable<T>::value
         >
{ };
~~~

devreal:
For the record: default construction of elements in containers has been brought up in https://github.com/dash-project/dash/issues/479.

rkowalewski:
> For the record: default construction of elements in containers has been brought up in #479.

We discussed it already in a F2F and agreed on not doing that right? It is currently done only for `dash::Shared`. For the others we have to keep in mind NUMA #615.

rkowalewski:
Like always, such PR get larger than expected. I am documenting everything now with sweet `static_assert`s in our test suite. These rules show if we can convert `GlobRef<T>` to `GlobRef<U>` by considering all const modifiers.

While this approach demonstrates it for `GlobRef` we can adapt it for `GlobIter`, `GlobPtr` and other friends.

dash/test/iterator/GlobAsyncRefTest.cc
```diff
@@ -186,8 +169,8 @@ TEST_F(GlobAsyncRefTest, ConstTest)
   // assignment const -> const
   dash::GlobAsyncRef<const int> agref2 = carr.async[0];
   // explicit conversion const->non-const
-  dash::GlobAsyncRef<int> agref3 =
-                            static_cast<dash::GlobAsyncRef<int>>(carr.async[0]);
+  //dash::GlobAsyncRef<int> agref3 =
+  //                          static_cast<dash::GlobAsyncRef<int>>(carr.async[0]);
```
devreal, line 173: I guess I'm missing something here, but why do we disallow explicit const->non-const conversion?

devreal:
Sorry, this PR completely slipped through. Just a minor question, seems fine to me otherwise

Issue #631, opened 2019-03-05
`constexpr constructor must initialize direct base class` with Intel 2019


bertwesarg:
Got this from someone which compiled with Intel 2019 and `-std=c++14`.

Is this some known problem, which is already solved?

```
/work/xu077052/spec/hpc2020-007/benchspec/HPC/620.dashmg_s/../111.libdash/src/include/dash/halo/Halo.h(532): error: constexpr constructor must initialize direct base class "dash::Dimensional<uint8_t={unsigned char}, 3>"
  class RegionSpec : public Dimensional<uint8_t, NumDimensions> {
                            ^
          detected during:
            instantiation of class "dash::halo::RegionSpec<NumDimensions> [with NumDimensions=3]" at line 780
            instantiation of "dash::halo::HaloSpec<NumDimensions>::HaloSpec(const StencilSpecT &) [with NumDimensions=3, StencilSpecT=StencilSpecT]" at line 83 of "/work/xu077052/spec/hpc2020-007/benchspec/HPC/620.dashmg_s/../111.libdash/src/include/dash/halo/HaloMatrixWrapper.h"
            instantiation of "dash::halo::HaloMatrixWrapper<MatrixT>::HaloMatrixWrapper(MatrixT &, const dash::halo::HaloMatrixWrapper<MatrixT>::GlobBoundSpec_t &, const StencilSpecT &...) [with MatrixT=MatrixT, StencilSpecT=<StencilSpecT>]" at line 122 of "multigrid3d.cpp"
```

PR #632, opened 2019-03-05
removed unnecessary inheritance

dhinf:
bug fix for issue #631

### DIFF
--- a/dash/include/dash/halo/Halo.h
+++ b/dash/include/dash/halo/Halo.h
@@ -528,7 +528,7 @@ private:
  * The region extent applies to all dimensions.
  */
 template <dim_t NumDimensions>
-class RegionSpec : public Dimensional<uint8_t, NumDimensions> {
+class RegionSpec {
 private:
   using Self_t = RegionSpec<NumDimensions>;
 


bertwesarg:
Could you test it with Intel 19? I could send this one-liner to the original report, but it may take some time to get feedback.

fuchsto (changes requested):
The subclassing was not unnecessary, it indicates the type category.
This is why public inheritance was used. The formal reading is "RegionSpec<D> IS_A Dimensional<uint8_t, D>".

The compiler error is right, though, the fix is to add missing delegate constructor calls in RegionSpec.

dhinf:
RegionSpec is not really "a" Dimensional. The member RegionCoords is Dimensional. But RegionSpec only combines RegionCoords with some other properties.

fuchsto:
Well, the question whether "A is a B" (`A : public B`) means:
"Can you pass an instance of `A` to every context that expects a `B`?
In this example, let's assume:

```c++
SizeT volume(const Dimensional & d) { return std::accumulate( ... , std::multiplies); }
RegionSpec<...> rs(...);
auto region_volume = volume(rs);
```

I don't see why something like that should not be okay, so `RegionSpec` could act as `Dimensional`.
Back then, I added some other types for Cartesian domains, perhaps something else is a better fit
We'll see tomorrow.

devreal:
What is the way forward here? I don't remember this was discussed at the F2F but it keep biting...

dhinf:
i still think that no inheritance should be used. I don't use the data structure of the base class or any methods.

devreal:
Can the `RegionSpec` be implemented in terms of `Dimensional`?

dhinf:
Not really, because RegionSpec combines RegionCoords (Dimensional), the extent and some other useful things. In an older version i didn't used RegionCoords, but instead Dimensional to store the coordinates. Later i devided them. And at this point i missed to delete the inheritance.

devreal:
> In an older version i didn't used RegionCoords, but instead Dimensional to store the coordinates

What's the reason to not use `Dimensional` to store the coordinates anymore?

dhinf:
Because i needed RegionCoords separately for other components. But seriously, i don't understand the problem here. We have so much other, more urgent, stuff to do.

devreal:
@fuchsto I think we should merge this PR as is. Can you accept `RegionSpec` not being a `Dimensional`?

PR #633, opened 2019-03-07
Add Boost as a dependency

pascalj:
We could eliminate much of our code if we just used Boost, e.g. much of the iterators and const-correctnes stuff @rkowalewski is currently working on. This PR adds the dependency to DASH.

So far this will only add the include directory to the library target as well
as to the compiler wrapper.

If at some point some Boost module must be linked, it should be added to
Boost.cmake.

How can I add Boost to our CI Docker images?

### DIFF
--- a/CMakeExt/Boost.cmake
+++ b/CMakeExt/Boost.cmake
@@ -1,0 +1,5 @@
+find_package(Boost REQUIRED COMPONENTS
+        ## If you need to link a boost library, add it below. For example:
+        # chrono
+        ## will link boost_chrono for you and also add it as a dependency.
+        )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,6 +117,7 @@ endif()
 
 # Load build modules to locate libraries after environment setup
 # has been loaded:
+include(${CMAKE_SOURCE_DIR}/CMakeExt/Boost.cmake)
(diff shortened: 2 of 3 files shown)


devreal:
Please be careful here. From my experience boost is a pain in the butt, not readily available on many machines (it took us ages to transition to C++14 because there *may* be machines that have ancient compilers...), and not stable between releases. Adding a new mandatory dependency further complicates building DASH. 

It is not clear to me which libraries you want to use and how that will help with the stuff @rkowalewski does/did. I think we should find a consensus on whether we want to pull in boost in the first place instead of blindly rushing it in, without a clear explanation of the benefits and essential things like the minimum version required etc. If we decide to include boost, can we integrate it into our source tree / build system to avoid external mandatory dependencies? If not, how do we avoid tripping over incompatibilities?

dhinf:
I'm also not a big fan of adding boost. Like devreal already mentioned, boost will complicate things, especially on HPC systems. Put it on the agenda for the next meeting.

devreal:
As per the discussion at the F2F today:

1) A majority thinks that relying on an existing boost installation is dangerous and will make installing DASH tedious at best, e.g., due to incompatibilities between different boost versions and the effort needed to install the correct version. Thus, we would need to ship and deploy the boost version we need. 
2) It is not clear what the benefits would be in terms of code complexity in DASH. An example would be helpful to judge that, potentially in the context of this PR.

PR #635, opened 2019-03-12
GlobIter: Make GlobMemType a public typedef

bertwesarg:


bertwesarg:
@pascalj this is needed to get the executor in mephisto work again:

https://github.com/mephisto-hpc/testing/commit/60f6084ca9791169eadfe3f5ddf72d7694cc021c#diff-0d729d073571dd4e46129f851327de2eR39

fuchsto (changes requested):
Iterators and memory as independent concepts is a crucial principle in the STL.
The conceptual discrimination here is not formalese, it is most definitely existential.
Or to avoid category theory, in software engineering terms:
Let's remember that *every* implementation of the Global Iterator concept had to provide a public type `memory_type` which again is required to be a model of the Global Memory concept.

I had a look at the Mephisto code that motivated this request.
We actually need execution- and memory spaces as proposed (but not realized?) in Kokkos:

https://github.com/kokkos/kokkos/issues/457#issuecomment-297494430

As a workaround for now, we could introduce `dash::global_iterator_traits<GlobIterT>::memory_type`, default-defined as `void` and specialized for ("befriended") global iterators which are actually memory-bound.

bertwesarg:
> As a workaround for now, we could introduce `dash::global_iterator_traits<GlobIterT>::memory_type`, default-defined as `void` and specialized for ("befriended") global iterators which are actually memory-bound.

Ok, but than why did you removed @pascalj and @rkowalewski? @rkowalewski implemented the memory spaces in DASH and @pascalj implemented the mephisto part. I was only trying to get it to work again.

fuchsto:
The changes in this PR are not related to memory spaces or Mephisto but on iterator concepts. Global memory concepts, maybe (both developed and implemented by myself, but why should that be relevant?)
I did not remove them from requests for review with malicious intent.

PR #639, opened 2019-03-25
Set default window type to non-shared static

devreal:
This PR changes the build scripts and CMakeList to configure DART to use regular windows instead of dynamic + shared memory windows in `dart_team_memalloc_aligned`.

Background: 
The use  of dynamic windows causes all sorts of troubles:

1) It prevents the use of RDMA capabilities on any decent hardware.
2) Open MPI chokes if `MPI_THREAD_MULTIPLE` is enabled and no RDMA-capable network is found (thanks to @bertwesarg for bringing this back up) because the shared-memory component does not support dynamic windows (how could it?) and the `pt2pt` component lacks support for `MPI_THREAD_MULTIPLE` (not sure this will ever be fixed). Hence, DASH won't even work on a single node.

The MPI implementation will likely short-cut through shared memory for on-node communication so the benefit of doing it ourselves really seems marginal. On the other hand, changing the default avoids some bad surprises.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,9 +56,9 @@ option(ENABLE_ASSERTIONS
 option(ENABLE_UNIFIED_MEMORY_MODEL
        "Specify whether to assume unified memory model" on)
 option(ENABLE_SHARED_WINDOWS
-       "Specify whether shared memory features are enabled" on)
+       "Specify whether shared memory features are enabled" off)
 option(ENABLE_DYNAMIC_WINDOWS
-       "Specify whether to use dynamic MPI windows for collective allocation" on)
+       "Specify whether to use dynamic MPI windows for collective allocation" off)
 option(ENABLE_DEFAULT_INDEX_TYPE_LONG
        "Specify whether to use int64_t as default index type" on)
 option(ENABLE_LIBNUMA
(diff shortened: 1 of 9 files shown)


bertwesarg (commented):
Could mention somewhere, why shared and dynamic windows are off by default. Not that someone thinks, better enable it.

bertwesarg:
I would not consider the `build*.sh` scripts a good place to document CMake variables. why not inside `CMakeLists.txt` into the `options()`?

Btw, does other combinations instead of `off`&`off` or `on`&`on` make sense too?

rkowalewski (approved):
LGTM

devreal:
> I would not consider the build*.sh scripts a good place to document CMake variables. why not inside CMakeLists.txt into the options()?

We can move it to the `options` in the CMake file. So far these descriptions are kept quite short though and the description of these variables would have to be split into two.

> Btw, does other combinations instead of off&off or on&on make sense too?

There is one case where dynamic windows without shared windows make sense. That is related to #615: dynamic windows do not register the memory and thus don't fault the pages, leading to better NUMA behavior. There are also some MPI implementations that don't really do well with shared memory window IIRC. So as long  as we don't have the NUMA issue under control I'd leave the two options separate but we should consider merging them at some point.

PR #644, opened 2019-04-30
Improvements to DART-MPI

### LINKED ISSUE #642: CMake `ENABLE_UNIFIED_MEMORY_MODEL` does not have an effect

devreal:
This PR fixes #642 by:

1) Removing the flag;
2) Checking windows for the unified memory model and printing a warning if that's not available;
3) Calling `MPI_Win_sync` if shared memory communication is enable.

In case of `MPI_WIN_SEPARATE`, we have no way to correctly synchronize stores and puts from remote. This model breaks with our assumption that we can mix local stores and puts from remote into the same window without explicit synchronization in between. Hence a big fat warning that DASH cannot guarantee consistency...

It also allocates the pre-allocated memory block through `MPI_Win_allocate` if shared memory windows are disabled and adds some info keys and asserts where possible.


dhinf (changes requested):
dart-impl/mpi/src/dart_communication.c : 

CHECK_EQUAL_BASETYPE(src_type, dst_type); is used redundantly in some functions (e.g. dart_put)

bertwesarg:
I cannot comment much on the `MPI_Win_sync` changes, but otherwise looks good to me. Thanks.

devreal:
@dhinf Good catch, should be fixed now.

dhinf (approved):

rkowalewski:
Just for clarifications...Can you list the rationales why you added these changes, please.  Also the comment referring to example 11.2.1 is not clear to me. I cannot find such an example in the MPI-3.1 report after a quick look.

devreal:
> Just for clarifications...Can you list the rationales why you added these changes, please. Also the comment referring to example 11.2.1 is not clear to me. I cannot find such an example in the MPI-3.1 report after a quick look.

Sure: 

1) Example 11.21 on page 468 states: 
    > The following example demonstrates the proper synchronization in the unified memory model when a data transfer is implemented with load and store in the case of windows in shared memory (instead of `MPI_PUT` or `MPI_GET`) and the synchronization between processes is performed using point-to-point communication. The synchronization between processes must be supplemented with a memory synchronization through calls to `MPI_WIN_SYNC`, which act locally as a processor-memory barrier.

    Hence the call to `MPI_Win_sync` in combination with flush. According to the example, we would actually need a second `sync` just before reading from shared memory but that requires the user to call `container.flush()` after the external synchronization (barrier, MPI send/recv) is done. I don't see how we can do that ourselves.

2) For the separate memory model: Point `S3` on page 455 states: 
    > A put or accumulate must not access a target window once a store or a put or accu-
mulate update to another (overlapping) target window has started on a location in
the target window, until the update becomes visible in the public copy of the win-
dow. Conversely, a store to process memory to a location in a window must not start
once a put or accumulate update to that target window has started, until the put or
accumulate update becomes visible in process memory. In both cases, the restriction
applies to operations even if they access disjoint locations in the window

    If I correctly understand the rational following this point, the problem is the bi-directional synchronization, i.e., which store-updates to copy into the public window and which put/accumulate-update to copy from the public into the private window: 
    > The reason for this constraint is that, on some architectures, explicit coherence restoring operations may be needed at synchronization points. A different operation may be needed for locations that were updated by stores and for locations that were remotely updated by put or accumulate operations.

    I opted to not abort the run, the user may try his luck with the separate memory model. He has been warned. Also, I checked the source code for MPICH (does not use the separate memory model afaics) and Open MPI (uses the separate memory model on non-x86 Portals4 platforms. There may be other exotic platforms out there but I don't expect this to bite us anytime soon. If so we have go back to the drawing board. Mainly, this PR gets rid of `sync` calls where they are not needed, i.e., if shared memory windows are disabled.

PR #645, opened 2019-04-30
Revise math::factorize()

bertwesarg:


dash/include/dash/internal/Math.h
```diff
@@ -161,44 +143,16 @@ std::map<Integer, int> factorize(Integer n)
...
-  }
-  if (n > 2) {
-    // Check if the remainder is prime:
-    auto it = primes.find(n);
-    if (it == primes.end()) {
-      primes.insert(n);
-    }
+  const auto factorization = factorize(n);
+  for(auto it = factorization.begin(); it != factorization.end(); ++it) {
+    primes.insert(it->first);
```
devreal, line 157: Small further improvement: `primes.insert(factorization.begin(), factorization.end())`

dash/include/dash/internal/Math.h
```diff
@@ -107,7 +107,7 @@ double sigma(const Iter& begin, const Iter& end)
  * \code
  *   long number  = 2 * 2 * 2 * 4 * 7 * 7;
  *   auto factors = dash::math::factorize(number);
- *   // returns map { (2,3), (4,1), (7,2) }
+ *   // returns map { (2,5), (7,2) }
  * \endcode
  */
 template <typename Integer>
```
devreal, line 113: Add a `static_assert` that `Integer` is indeed an integral type?

devreal (approved):
Looks good to me. I wanted to ask for tests but realized that we have `AutobalanceTest.Factorize` :) Just two minor comments

dash/include/dash/internal/Math.h
bertwesarg: are you sure? `factorization` is a `std::map` where the iterators gives you `std::pair`s

devreal: Ahh, you're right of course. Never mind then :)

devreal (changes requested):
Tests fail. Example: 

```
[  ERROR   ] [UNIT 0] in /root/project/dash/test/pattern/TilePatternTest.cc:56
Expected equality of these values:
  dash::size()
    Which is: 2
  teamspec_2d.size()
    Which is: 1
[  ERROR   ] [UNIT 1] in /root/project/dash/test/pattern/TilePatternTest.cc:56
Expected equality of these values:
  dash::size()
    Which is: 2
  teamspec_2d.size()
    Which is: 1
[  FAILED  ] TilePatternTest.Tile2DimTeam2Dim
[  ERROR   ] [UNIT 0] in /root/project/dash/test/TestPrinter.h:149
Failed
Testcase failed at least on one unit
```

dash/include/dash/internal/Math.h
bertwesarg: It should also be `unsigned`? And the exponents too?

devreal: IMO it's better to add a runtime assert for positive numbers to avoid casting a negative number to `unsigned`. The factors should be unsigned though, yes.

bertwesarg: `static_assert` is compile time

bertwesarg: I think I got it:

* [x] add `static_assert` to check that `Integer` is an integral type
* [x] add `assert` that `n => 0`
* [x] change mapped type from `int` to `unsigned`

correct?

bertwesarg:
Can someone triggers the CI again please.

dash/include/dash/internal/Math.h
devreal: That's what I meant, yes :)

bertwesarg:
> Tests fail. Example:
> 
> ```
> [  ERROR   ] [UNIT 0] in /root/project/dash/test/pattern/TilePatternTest.cc:56
> Expected equality of these values:
>   dash::size()
>     Which is: 2
>   teamspec_2d.size()
>     Which is: 1
> [  ERROR   ] [UNIT 1] in /root/project/dash/test/pattern/TilePatternTest.cc:56
> Expected equality of these values:
>   dash::size()
>     Which is: 2
>   teamspec_2d.size()
>     Which is: 1
> [  FAILED  ] TilePatternTest.Tile2DimTeam2Dim
> [  ERROR   ] [UNIT 0] in /root/project/dash/test/TestPrinter.h:149
> Failed
> Testcase failed at least on one unit
> ```

CI is now green.

devreal (approved):
Looks good to me now :)

PR #649, opened 2019-05-16
TilePattern: Fix coordinate <-> index conversion

### LINKED ISSUE #620: TilePattern: erroneous conversion between global index and global coordinates
Consider the following matrix:

```C
  using PatternT = typename dash::TilePattern<4>;
  using TiledMatrix = dash::Matrix<double, 4, dash::default_index_t, PatternT>;
  TiledMatrix A(8, 8,                              // 8x8 tiles
                dash::TILE(2), dash::TILE(2),      // each superblock has 2x2 tiles
                2, 2,                              // each tile has 2x2 elements
                dash::TILE(2), dash::TILE(2), ts);
```

That is meant to represent a 16x16 two-dimensional matrix with 4x4 super-blocks of 2x2 tiles a 2x2 elements, i.e., super-blocks contain 4x4 elements split into 2x2 tiles. The idea is that I can access individual tiles {x, y} via A[x][y].  See #619 for some more details.

A matrix could look like this (with `00` denoting block 0 on unit 0, `12` denoting block 2 on unit 1, etc):
```
+--------------------------------+--+--+--+--+
|00|00|01|01|10|10|11|11|04|04|  |  |  |  |  |
+--------------------------------------------+
|00|00|01|01|10|10|11|11|04|04|  |  |  |  |  |
+--------------------------------------------+
|02|02|03|03|12|12|13|13|  |  |  |  |  |  |  |
+--------------------------------------------+
|02|02|03|03|12|12|13|13|  |  |  |  |  |  |  |
+--------------------------------------------+
|  |  |  |  |  |  |  |  |  |  |  |  |  |  |  |
+--------------------------------------------+
[...]
+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
```

For each tile, I need to query the global pointer to the first element and this is where things go south. Take the block global coordinates `{0,2}`: querying the global index of its first element via `A.pattern().global_at({0, 2, 0, 0})` yields 16 (skipping all 16 elements in blocks `00`--`03`). That seems correct. However, turning that global offset back into global coordinates (as happens when I call `A.begin() + 16`), the coordinates returned by `CartesianIndexSpace::coords(16)` are `{0, 4, 0, 0}`, pointing to the wrong element (the first element of block `04`).

The `GlobIter` I get is `dash::GlobIter<d>(idx:16, gptr:dash::GlobPtr<d>(u000000|f00|s0001|t0000|o0000000000000080))`. However, it is my understanding that offset 16 should point to the first element in block `10`, as the iteration order across blocks is `00`->`01`->`02`->`03`->`10` (4 preceding blocks a 4 elements), not `00`->`01`->`10`->`11`->`04`. Please correct me if I'm wrong here.

I tried tracking this down for a while. The reason seems to be that `CartesianIndexSpace::coords` uses the tile-based offsets, which breaks the iteration order, i.e., `_offset_row_major = {24, 4, 2, 1}`, i.e., it ignores the second row of tiles. The correct offsets to use for this conversion would be `{48, 8, 4, 1}`. I'm not sure how to take it from here so it would be great if someone could take a look into it...

devreal:
The conversion between coordinates and indices should not be done through directly throught the `CartesianIndexSpace` because that has no notion of tiles. 

To make things consistent, this PR removes the exposure of the `[local_]memory_layout` in the patterns and adds a missing `coords` member that takes a viewspec. All conversion should be done through the patterns as the results might differ from the conversion the `memory_layout` provides for some pattern types.

Fixes #620

### DIFF
--- a/dash/include/dash/Cartesian.h
+++ b/dash/include/dash/Cartesian.h
@@ -698,6 +698,8 @@ public:
     }
     // Tiles in at least one dimension
     // TODO
+    DASH_THROW(dash::exception::NotImplemented,
+               "CartesianIndexSpace::resize(IndexType) not implemented for tiles!");
   }
 
   /**
@@ -737,6 +739,8 @@ public:
     }
     // Tiles in at least one dimension
     // TODO
+    DASH_THROW(dash::exception::NotImplemented,
+               "CartesianIndexSpace::at(IndexType) not implemented for tiles!");
   }
 
   /**
@@ -758,11 +762,6 @@ public:
     for (auto d = 0; d < NumDimensions; ++d) {
       coords[d] = point[d] + viewspec[d].offset;
     }
-    if (!_distspec.is_tiled()) {
-      // Default case, no tiles
-      return parent_t::at(coords);
-    }
-    // Tiles in at least one dimension
     return at(coords);
   }
 
@@ -778,6 +777,8 @@ public:
     }
     // Tiles in at least one dimension
     // TODO
+    DASH_THROW(dash::exception::NotImplemented,
+               "CartesianIndexSpace::coords(IndexType) not implemented for tiles!");
   }
 
 private:
--- a/dash/include/dash/Pattern.h
+++ b/dash/include/dash/Pattern.h
@@ -564,21 +564,6 @@ public:
   const std::array<size_type, NumDimensions> & extents() const;
 
   /**
-   * Cartesian index space representing the underlying memory model of the
-   * pattern.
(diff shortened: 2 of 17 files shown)


dhinf (approved):

devreal:
@fuchsto Please check whether the changes to the pattern concept and the new implementations of `coords()` in the tiled patterns are sound.

rkowalewski (approved):
We should merge IMHO

Issue #657, opened 2019-07-11
`dash::copy` copies the wrong values in local to global


bertwesarg:
Extending the example in #656 even further by copying the values back to the larger team by the smaller team, reveals that the local-to-global mode of `dash::copy` also does not work

```c++
#include <iostream>

#include <libdash.h>

int
main(int ac, char *av[])
{
    using TeamSpecT = dash::TeamSpec<2>;
    using MatrixT = dash::NArray<double, 2>;
    using PatternT = typename MatrixT::pattern_type;
    using SizeSpecT = dash::SizeSpec<2>;
    using DistSpecT = dash::DistributionSpec<2>;

    dash::init(&ac, &av);

    auto size_spec = SizeSpecT(8, 8);
    auto dist_spec = DistSpecT(dash::BLOCKED, dash::BLOCKED);

    auto& team_all = dash::Team::All();
    TeamSpecT team_all_spec(team_all.size(), 1);
    team_all_spec.balance_extents();
    MatrixT grid_more(size_spec, dist_spec, team_all, team_all_spec);
    dash::fill(grid_more.begin(), grid_more.end(), (double)team_all.myid());
    team_all.barrier();

    auto gextents = grid_more.extents();

    if (team_all.myid() == 0) {
        std::cout << "init\n";
        for (    uint32_t y = 0; y < gextents[0]; ++y) {
            std::cout << "[" << std::setw(2) << y << "] := {";
            for (uint32_t x = 0; x < gextents[1]; ++x)
                std::cout << " " << std::setw(2) << (double)grid_more[y][x];
            std::cout << "}\n";
        }
    }

    dash::Team& team_fewer= team_all.split(4);
    team_all.barrier();

    if (!team_fewer.is_null() && 0 == team_fewer.position()) {
        TeamSpecT team_fewer_spec(team_fewer.size(), 1);
        team_fewer_spec.balance_extents();

        MatrixT grid_fewer(size_spec, dist_spec, team_fewer, team_fewer_spec);
        dash::fill(grid_fewer.begin(), grid_fewer.end(), -1.0);

        auto lextents= grid_fewer.pattern().local_extents();

        for (uint32_t y = 0; y < lextents[0]; ++y) {
            auto gcorner_fewer = grid_fewer.pattern().global({y, 0});
            auto gbegin = grid_more.begin() + grid_more.pattern().global_at(gcorner_fewer);

            auto loffset = grid_fewer.pattern().local_at({y, 0});
[...]

Issue #656, opened 2019-07-11
`dash::copy` copies the wrong values in global to local


bertwesarg:
Extending the example in #653 gives now the wrong values in the target matrix:

```c++
#include <iostream>

#include <libdash.h>

int
main(int ac, char *av[])
{
    using TeamSpecT = dash::TeamSpec<2>;
    using MatrixT = dash::NArray<double, 2>;
    using PatternT = typename MatrixT::pattern_type;
    using SizeSpecT = dash::SizeSpec<2>;
    using DistSpecT = dash::DistributionSpec<2>;

    dash::init(&ac, &av);

    auto size_spec = SizeSpecT(16, 16);
    auto dist_spec = DistSpecT(dash::BLOCKED, dash::BLOCKED);

    auto& team_all = dash::Team::All();
    TeamSpecT team_all_spec(team_all.size(), 1);
    team_all_spec.balance_extents();
    MatrixT grid_more(size_spec, dist_spec, team_all, team_all_spec);
    dash::fill(grid_more.begin(), grid_more.end(), (double)team_all.myid());
    team_all.barrier();

    if (team_all.myid() == 0) {
        auto gextents = grid_more.extents();
        for (    uint32_t y = 0; y < gextents[0]; ++y) {
            std::cout << "[" << y << "] := {";
            for (uint32_t x = 0; x < gextents[1]; ++x)
                std::cout << " " << (double)grid_more[y][x];
            std::cout << "}\n";
        }
    }

    dash::Team& team_fewer= team_all.split(4);
    team_all.barrier();
    if (!team_fewer.is_null() && 0 == team_fewer.position()) {
        TeamSpecT team_fewer_spec(team_fewer.size(), 1);
        team_fewer_spec.balance_extents();

        MatrixT grid_fewer(size_spec, dist_spec, team_fewer, team_fewer_spec);
        dash::fill(grid_fewer.begin(), grid_fewer.end(), -1.0);

        auto lextents= grid_fewer.pattern().local_extents();

        for (uint32_t y = 0; y < lextents[0]; ++y) {
            auto gcorner_fewer = grid_fewer.pattern().global({y, 0});
            auto gbegin = grid_more.begin() + grid_more.pattern().global_at(gcorner_fewer);

            auto loffset = grid_fewer.pattern().local_at({y, 0});
            dash::copy(gbegin, gbegin + lextents[1],
                       grid_fewer.lbegin() + loffset);
        }
        team_fewer.barrier();

[...]

PR #659, opened 2019-07-13
Rewrite dash::copy

devreal:
This PR addresses #657  and #656 by making `dash::copy` adhere to the iteration space of the iterators passed. This requires splitting up the copy into transfers to/from contiguous global memory regions. 

Copying should now be correct for both global-to-local and local-to-global. Larger (>1 cache line size, maybe better be a pagesize?) local copies are deferred until all remote transfers are kicked off in an attempt to overlap remote and local copies.

~For `GlobIter` and `GlobPtr` the detection of contiguous memory regions is fairly straightforward, for `GlobViewIter` (or any view-based iterator)  finding boundaries of contiguous regions is in O(n). Ideas on how to improve that are welcome (@fuchsto @dhinf @rkowalewski can the viewspec be used for that?)~
All iterators (except for `GlobPtr`) make use of views to find contiguous memory boundaries,

### DIFF
--- a/dash/include/dash/algorithm/Copy.h
+++ b/dash/include/dash/algorithm/Copy.h
@@ -236,7 +236,6 @@ GlobOutputIt copy_impl(
                  "out_first:", out_first);
   typedef typename GlobOutputIt::size_type  size_type;
   typedef typename GlobOutputIt::value_type  value_type;
-  typedef typename std::remove_const<value_type>::type  nonconst_value_type;
   const size_type num_elem_total = dash::distance(begin, end);
   if (num_elem_total <= 0) {
     DASH_LOG_TRACE("dash::internal::copy_impl", "input range empty");
@@ -253,9 +252,9 @@ GlobOutputIt copy_impl(
 
   ContiguousRangeSet<GlobOutputIt> range_set{out_first, out_last};
 
-  std::vector<local_copy_chunk<nonconst_value_type>> local_chunks;
+  std::vector<local_copy_chunk<value_type>> local_chunks;
 
-  nonconst_value_type* in_first = const_cast<nonconst_value_type*>(begin);
+  auto in_first = begin;
 
   //
   // Copy elements to every unit:
@@ -272,8 +271,7 @@ GlobOutputIt copy_impl(
 
     // handle local data locally
     if (cur_out_first.is_local()) {
-      nonconst_value_type* dest_ptr =
-                        const_cast<nonconst_value_type*>(cur_out_first.local());
+      value_type* dest_ptr = cur_out_first.local();
       // if the chunk is less than a page or if it is the only transfer
       // don't bother post-poning it
       if (num_elem_total == num_copy_elem ||
@@ -281,8 +279,7 @@ GlobOutputIt copy_impl(
         std::copy(src_ptr, src_ptr + num_copy_elem, dest_ptr);
       } else {
         // larger chunks are handled later to allow overlap
-        local_copy_chunk<nonconst_value_type> chunk{src_ptr, dest_ptr, num_copy_elem};
-        local_chunks.push_back(chunk);
+        local_chunks.push_back({src_ptr, dest_ptr, num_copy_elem});
       }
     } else {
       auto dst_gptr = cur_out_first.dart_gptr();
--- a/dash/include/dash/algorithm/Copy.h
+++ b/dash/include/dash/algorithm/Copy.h
@@ -168,8 +168,10 @@ ValueType * copy_impl(
 
     // handle local data locally
     if (cur_in.is_local()) {
-      // if the chunk is less than a cache line don't bother post-poning it
-      if (l2_line_size > num_copy_elem*sizeof(ValueType)) {
+      // if the chunk is less than a cache line or if it is the only transfer
+      // don't bother post-poning it
+      if (num_copy_elem == num_elem_total ||
+          l2_line_size > num_copy_elem*sizeof(ValueType)) {
         std::memcpy(dest_ptr, cur_in.local(), num_copy_elem*sizeof(ValueType));
       } else {
         // larger chunks are handled later to allow overlap
@@ -264,10 +266,12 @@ GlobOutputIt copy_impl(
 
     // handle local data locally
     if (cur_out_first.is_local()) {
-      // if the chunk is less than a cache line don't bother post-poning it
       nonconst_value_type* dest_ptr =
                         const_cast<nonconst_value_type*>(cur_out_first.local());
-      if (l2_line_size > num_copy_elem*sizeof(ValueType)) {
+      // if the chunk is less than a cache line or if it is the only transfer
+      // don't bother post-poning it
+      if (num_elem_total == num_copy_elem ||
+          l2_line_size > num_copy_elem*sizeof(ValueType)) {
         std::memcpy(dest_ptr, src_ptr, num_copy_elem*sizeof(ValueType));
       } else {
         // larger chunks are handled later to allow overlap
--- a/dash/include/dash/algorithm/Copy.h
+++ b/dash/include/dash/algorithm/Copy.h
@@ -105,9 +105,9 @@ namespace internal {
 
 template<typename ValueType>
 struct local_copy_chunk {
-  ValueType *src;
-  ValueType *dest;
-  size_t     size;
+  const ValueType *src;
+        ValueType *dest;
+  const size_t     size;
 };
 
 template<typename ValueType>
--- a/dash/include/dash/algorithm/Copy.h
+++ b/dash/include/dash/algorithm/Copy.h
@@ -3,6 +3,7 @@
 
 #include <dash/Future.h>
 #include <dash/Iterator.h>
+#include <dash/iterator/internal/ContiguousRange.h>
 
 #include <dash/algorithm/LocalRange.h>
 
@@ -105,19 +106,17 @@ template <
   typename ValueType,
   class GlobInputIt >
 ValueType * copy_impl(
-  GlobInputIt                  in_first,
-  GlobInputIt                  in_last,
+  GlobInputIt                  begin,
+  GlobInputIt                  end,
   ValueType                  * out_first,
   std::vector<dart_handle_t> & handles)
 {
-  DASH_LOG_TRACE("dash::copy_impl()",
-                 "in_first:",  in_first.pos(),
-                 "in_last:",   in_last.pos(),
+  DASH_LOG_TRACE("dash::copy_impl() global -> local",
+                 "in_first:",  begin.pos(),
+                 "in_last:",   end.pos(),
                  "out_first:", out_first);
-  auto pattern = in_first.pattern();
-  typedef typename decltype(pattern)::index_type index_type;
-  typedef typename decltype(pattern)::size_type  size_type;
-  size_type num_elem_total = dash::distance(in_first, in_last);
+  typedef typename GlobInputIt::pattern_type::size_type  size_type;
+  size_type num_elem_total = dash::distance(begin, end);
   if (num_elem_total <= 0) {
     DASH_LOG_TRACE("dash::copy_impl", "input range empty");
     return out_first;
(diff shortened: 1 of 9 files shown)


dash/include/dash/algorithm/Copy.h
```diff
@@ -534,169 +409,18 @@ ValueType * copy(
...
-    }
-  } else {
-    DASH_LOG_TRACE("dash::copy", "no local subrange");
-    // All elements in input range are remote
-    out_last = dash::internal::copy_impl(in_first,
-                                         in_last,
-                                         dest_first,
-                                         handles);
-  }
+  // All elements in input range are remote
```
bertwesarg: does this comment still holds?

dash/include/dash/algorithm/Copy.h
```diff
@@ -799,84 +523,14 @@ GlobOutputIt copy(
...
-  } else {
-    // All elements in output range are remote
-    DASH_LOG_TRACE("dash::copy", "no local subrange");
-    out_last = dash::internal::copy_impl(
-                 in_first,
-                 in_last,
-                 out_first,
-                 handles);
-  }
+  // All elements in output range are remote
```
bertwesarg: does this comment still holds? the LOG_TRACE below too

dash/include/dash/algorithm/Copy.h
```diff
@@ -93,6 +94,22 @@ dash::Future<ValueType *> copy_async(
...
 
+template<typename ValueType>
+struct local_copy_chunk {
+  ValueType *src;
+  ValueType *dest;
+  size_t     size;
+};
+
+template<typename ValueType>
+void
```
bertwesarg: all other functions have their return type in the same line as the function name

dash/include/dash/algorithm/Copy.h
```diff
@@ -105,108 +122,81 @@ template <
...
+
+    DASH_ASSERT_GT(num_copy_elem, 0,
+                    "Number of element to copy is 0");
+    auto dest_ptr = out_first + num_elem_copied;
+
+    // handle local data locally
+    if (cur_in.is_local()) {
+      // if the chunk is less than a cache line or if it is the only transfer
+      // don't bother post-poning it
+      if (num_copy_elem == num_elem_total ||
```
bertwesarg: shouldn't this case use `std::copy` than instead of `std::memcpy`?

dash/include/dash/algorithm/Copy.h
```diff
@@ -534,169 +409,18 @@ ValueType * copy(
```
bertwesarg: `team` and `use_memcpy` not used anymore

dash/include/dash/algorithm/Copy.h
```diff
@@ -534,169 +409,18 @@ ValueType * copy(
 
   DASH_LOG_TRACE("dash::copy()", "blocking, global to local");
 
-  ValueType * dest_first = out_first;
-  // Return value, initialize with begin of output range, indicating no
-  // values have been copied:
-  ValueType * out_last   = out_first;
   // Check if part of the input range is local:
```
bertwesarg: comment outdated?

dash/include/dash/algorithm/Copy.h
```diff
@@ -105,108 +122,81 @@ template <
...
-                 "in_last:",   in_last.pos(),
+  DASH_LOG_TRACE("dash::copy_impl() global -> local",
+                 "in_first:",  begin.pos(),
+                 "in_last:",   end.pos(),
                  "out_first:", out_first);
-  auto pattern = in_first.pattern();
-  typedef typename decltype(pattern)::index_type index_type;
-  typedef typename decltype(pattern)::size_type  size_type;
-  size_type num_elem_total = dash::distance(in_first, in_last);
+  typedef typename GlobInputIt::pattern_type::size_type  size_type;
```
bertwesarg: still using `GlobInputIt::pattern_type::size_type` instead of just 'GlobInputIt::size_type`

dash/include/dash/algorithm/Copy.h
```diff
@@ -105,108 +122,81 @@ template <
...
+  // Copy elements from every unit:
+  //
+
+  for (auto range : range_set) {
+
+    auto cur_in        = range.first;
+    auto num_copy_elem = range.second;
+
+    DASH_ASSERT_GT(num_copy_elem, 0,
+                    "Number of element to copy is 0");
```
bertwesarg: Number of element**s** to copy is 0

dash/include/dash/algorithm/Copy.h
```diff
@@ -224,31 +214,88 @@ template <
...
+                    "Number of elements to copy is 0");
+    auto src_ptr  = in_first + num_elem_copied;
+
+    // handle local data locally
+    if (cur_out_first.is_local()) {
+      nonconst_value_type* dest_ptr =
+                        const_cast<nonconst_value_type*>(cur_out_first.local());
+      // if the chunk is less than a cache line or if it is the only transfer
+      // don't bother post-poning it
+      if (num_elem_total == num_copy_elem ||
```
bertwesarg, line 291: keep the order the same as above: `num_copy_elem == num_elem_total`

dash/include/dash/algorithm/Copy.h
```diff
@@ -274,193 +321,21 @@ dash::Future<ValueType *> copy_async(
...
-    DASH_LOG_TRACE("dash::copy_async", "no local subrange");
-    // All elements in input range are remote
-    dash::internal::copy_impl(in_first,
-                              in_last,
-                              dest_first,
-                              *handles);
-    out_last = out_first + total_copy_elem;
-  }
-  DASH_LOG_TRACE("dash::copy_async", "preparing future");
+  auto out_last = internal::copy_impl(in_first, in_last, out_first, *handles);
```
bertwesarg: all other calls are `dash::internal::copy_impl`, please harmonize it

bertwesarg (changes requested):
copy_async local-to-global is missing an empty-range check like global-to-local:

```c++
  DASH_LOG_TRACE("dash::copy_async()", "async, global to local");
  if (in_first == in_last) {
    DASH_LOG_TRACE("dash::copy_async", "input range empty");
    return dash::Future<ValueType *>(out_first);
  }
```

But global-to-local has TRACE_LOGs with `copy_async_impl` in it.

bertwesarg:
This should also fix #657.

dash/include/dash/algorithm/Copy.h
```diff
@@ -3,17 +3,26 @@
...
 #include <memory>
 #include <vector>
 
+
+/*
+ * Control whether dash::copy uses handles to wait for outstanding transfers
+ * to complete. If set to 0, dash::copy will use flush instead to ensure
+ * completion.
+ */
+#if !defined(DASH_COPY_USE_HANDLES)
```
bertwesarg: Where does this came from? How does one control this? Why should someone do not want to use handles?

devreal: Using handles is a trade-off: it allows you to wait for specific transfers instead of all transfers (which may be handy if multiple threads call `dash::copy` or if other transfers are already in flight). As a draw-back, they require memory allocation and might cause noticeable overhead if there many transfers due to short contiguous ranges. I have not done any comparisons, just leaving the option in here to disable the use of handles. You could compile your application using `-DDASH_COPY_USE_HANDLES=0` to disable them.

dash/include/dash/algorithm/Copy.h
```diff
@@ -105,108 +130,83 @@ template <
...
-      num_elem_copied += num_copy_elem;
-      if (handle != DART_HANDLE_NULL) {
-        handles.push_back(handle);
+    } else {
+      auto src_gptr = cur_in.dart_gptr();
+
+      DASH_LOG_TRACE("dash::copy_impl", "dest_ptr", dest_ptr,
+                    "src_gptr", src_gptr, "num_copy_elem", num_copy_elem);
+
+      if (handles != nullptr) {
```
bertwesarg, line 201: Could we move the `#if DASH_COPY_USE_HANDLES` here, instead of this runtime condition. That would make it transparent to the caller, whether handles are used or not.

bertwesarg (changes requested):

dash/include/dash/algorithm/Copy.h
devreal: Nope, `copy_impl` is used by `async_copy` too, which needs to used handles no matter what. The compiler will propagate the `nullptr` constant to optimize the runtime check away. `copy_impl` should not be called by the user directly anyway :)

dash/include/dash/algorithm/Copy.h
bertwesarg: But than, one should be able to make this decision on a call-by-call basis, not all-or-nothing. I'm not convinced, that this is needed right now, so why introduce it with this PR?

dash/include/dash/algorithm/Copy.h
```diff
@@ -224,31 +224,86 @@ template <
...
+
+  DASH_LOG_TRACE("dash::copy_impl",
+                 "total elements:",    num_elem_total,
+                 "expected out_last:", out_last);
+
+  size_type num_elem_copied = 0;
+
+  ContiguousRangeSet<GlobOutputIt> range_set{out_first, out_last};
+
+  std::vector<local_copy_chunk<value_type>> local_chunks;
```
bertwesarg: You are now inconsistent in the use of `ValueType` and `value_type`. The former is used below in `sizeof`

dash/include/dash/algorithm/Copy.h
```diff
@@ -224,31 +224,86 @@ template <
...
+      DASH_LOG_TRACE("dash::copy_impl", "src_ptr", src_ptr,
+                    "dst_gptr", dst_gptr, "num_copy_elem", num_copy_elem);
+      if (handles != nullptr) {
+        dart_handle_t handle;
+        dash::internal::put_handle(dst_gptr, src_ptr, num_copy_elem, &handle);
+        if (handle != DART_HANDLE_NULL) {
+          handles->push_back(handle);
+        }
+      } else {
+        dash::internal::put(dst_gptr, src_ptr, num_copy_elem);
```
bertwesarg, line 309: As we cannot overlap memcpy and comm anymore in this case, I think it does not make sense to postpone the memcpy above at all. Thus that condition should include _|| handles == nullptr_.

dash/include/dash/algorithm/Copy.h
devreal: Why not? Yes, in general controlling that on a call-by-call basis is preferable. Executing policies have been in the ether for so long, I can only guess when they will [vapor|material]ize. Until then, I don't see any harm in having at least a chance to control it globally. If we find that handle-less `std::copy` is more efficient (benchmarks welcome), we can easily change the default. The infrastructure is in place.

bertwesarg (changes requested):

dash/include/dash/algorithm/Copy.h
bertwesarg: > benchmarks welcome

Why should I argue with benchmarks, for something which you want to introduce without showing benchmarks in the first place?

dash/include/dash/algorithm/Copy.h
devreal: Good point, should be consistent now

dash/include/dash/algorithm/Copy.h
devreal: The overlap is still possible, `dash::internal::put` is non-blocking

dash/include/dash/algorithm/Copy.h
devreal: That was mainly meant as a hint that I only have finite resources at hand to work on this issue (it's gotten out of control already, really there is no such thing as a "quick fix"). I will make it a template parameter and add a benchmark example though

devreal:
```
----------------------------------------------------------------------------------
-- Runtime arguments
--   -r              20                                      repetitions per round
--   -n              10                                                     rounds
--   -s            4096      matrix size (number of double elements per dimension)
--   -t             512        tile size (number of double elements per dimension)
----------------------------------------------------------------------------------
units, mpi.impl,                          impl,   total [s],    bandwidth [MB/s]
   16,  craympi,              copy_with_handle,  0.05789930,       2318.12350063
   16,  craympi,           copy_without_handle,  0.03827390,       3506.76905149
   16,  craympi,                    copy_async,  0.05626080,       2385.63490032
   16,  craympi,              copy_with_handle,  0.05594790,       2398.97704829
   16,  craympi,           copy_without_handle,  0.03835725,       3499.14887016
   16,  craympi,                    copy_async,  0.05629005,       2384.39525280
   16,  craympi,              copy_with_handle,  0.05596170,       2398.38546720
   16,  craympi,           copy_without_handle,  0.03835130,       3499.69174448
   16,  craympi,                    copy_async,  0.05625060,       2386.06749084
   16,  craympi,              copy_with_handle,  0.05594790,       2398.97704829
   16,  craympi,           copy_without_handle,  0.03837160,       3497.84027771
   16,  craympi,                    copy_async,  0.05622025,       2387.35558807
   16,  craympi,              copy_with_handle,  0.05598375,       2397.44082881
   16,  craympi,           copy_without_handle,  0.03835040,       3499.77387459
   16,  craympi,                    copy_async,  0.05624890,       2386.13960451
   16,  craympi,              copy_with_handle,  0.05601960,       2395.90657556
   16,  craympi,           copy_without_handle,  0.03834610,       3500.16632721
   16,  craympi,                    copy_async,  0.05623290,       2386.81853506
   16,  craympi,              copy_with_handle,  0.05596895,       2398.07478968
   16,  craympi,           copy_without_handle,  0.03835305,       3499.53205808
   16,  craympi,                    copy_async,  0.05626840,       2385.31267994
   16,  craympi,              copy_with_handle,  0.05596455,       2398.26332920
   16,  craympi,           copy_without_handle,  0.03836975,       3498.00892630
   16,  craympi,                    copy_async,  0.05626480,       2385.46529980
   16,  craympi,              copy_with_handle,  0.05601120,       2396.26588968
   16,  craympi,           copy_without_handle,  0.03837580,       3497.45746017
   16,  craympi,                    copy_async,  0.05623485,       2386.73576972
   16,  craympi,              copy_with_handle,  0.05579010,       2405.76245606
   16,  craympi,           copy_without_handle,  0.03836095,       3498.81136937
   16,  craympi,                    copy_async,  0.05623960,       2386.53418588
```

```
----------------------------------------------------------------------------------
-- Runtime arguments
--   -r              20                                      repetitions per round
--   -n              10                                                     rounds
--   -s            4096      matrix size (number of double elements per dimension)
--   -t              64        tile size (number of double elements per dimension)
----------------------------------------------------------------------------------
units, mpi.impl,                          impl,   total [s],    bandwidth [MB/s]
   16,  craympi,              copy_with_handle,  0.46144025,        290.86697140
   16,  craympi,           copy_without_handle,  0.28127320,        477.17922646
   16,  craympi,                    copy_async,  0.45664310,        293.92260170
   16,  craympi,              copy_with_handle,  0.45475685,        295.14174003
   16,  craympi,           copy_without_handle,  0.28064115,        478.25391251
   16,  craympi,                    copy_async,  0.45654145,        293.98804424
   16,  craympi,              copy_with_handle,  0.45498160,        294.99594709
   16,  craympi,           copy_without_handle,  0.28066755,        478.20892725
   16,  craympi,                    copy_async,  0.45660895,        293.94458431
```

Not using handles clearly has an advantage so I made not using them the default.

devreal:
CI fails: global-to-local copy breaks under MPICH. I don't see any error for Open MPI, not sure why that fails. I suspect MPICH to have a broken `flush_local`, needs investigation...

devreal:
> I suspect MPICH to have a broken flush_local, needs investigation...

My fault, all checks clear now.

devreal:
@bertwesarg Please check my latest changes and approve if there are not other things to change.

bertwesarg (approved):
Fine with me.

PR #660, opened 2019-07-14
Implement global-to-global dash::copy based on #659

bertwesarg:


dash/include/dash/algorithm/Copy.h
```diff
@@ -597,19 +604,143 @@ copy_async(
...
-template <typename ValueType, class GlobInputIt, class GlobOutputIt>
-GlobOutputIt copy(
-    GlobInputIt /*in_first*/,
-    GlobInputIt /*in_last*/,
-    GlobOutputIt /*out_first*/)
+template <
+    class GlobInputIt,
+    class GlobOutputIt,
+    typename ValueType = typename GlobInputIt::value_type>
+GlobOutputIt copy_to(
```
devreal: Why not require the team of units involved in the global-to-global copy to be passed and identify the direction based on that?

bertwesarg: That would result in an API change for `dash::copy`, right? Is this intended? It also will result in a runtime decision. But at least we could stick with `dash::copy` instead of two new names. Not my decision.

Btw, should global-to-global also get an async variant?

bertwesarg: I would like to have a selection based on an template parameter, either true/false or an enum. That would retain the API and we have compile time selection.

devreal: You're right, the copy has to be collective over one of the two teams (either from the input or from the output iterators). There should be no default IMO because that may only be valid in 50% of the cases. A template parameter is fine with me, preferably an `enum` (or `struct` encapsulating the decision on what is local and what is remote).

Another note: global-to-global needs to have a barrier at the end to make sure that all data has been exchanged (including transfers into the unit's local memory). 

> Btw, should global-to-global also get an async variant?

I'd say yes. It would require a non-blocking barrier to facilitate the test/wait in the future though, something we don't have in DART atm (but which is easily added for dart-mpi, not sure about dart-gaspi. @dhinf ?)

dhinf: Let's put it this way. GASPI doesn't provide an IBarrier :)

bertwesarg: > Another note: global-to-global needs to have a barrier at the end to make sure that all data has been exchanged (including transfers into the unit's local memory).

They are already there and also on the function entry. What is the policy here in general?

bertwesarg: > I'd say yes. It would require a non-blocking barrier to facilitate the test/wait in the future though, something we don't have in DART atm (but which is easily added for dart-mpi, not sure about dart-gaspi. @dhinf ?)

I think we should postpone this then.

bertwesarg:
This is green now (though I'm note sure what this MatrixTest tried to achieved). Any more comments? Thanks.

PR #663, opened 2019-07-19
Allow for multiple Team splits and team cloning

### LINKED ISSUE #658: Cannot split a team multiple times
When splitting a team the second time I get the following runtime exception:

```
[    0 ERROR ] [ 12692 ] Team.cc                  :63   | dash::exception::InvalidArgument             | [ Unit 0 ] child of team 0 already set to 1, cannot set to 3
terminate called after throwing an instance of 'dash::exception::InvalidArgument'
  what():  [ Unit 0 ] child of team 0 already set to 1, cannot set to 3
```

Here is an example code:

```c++
#include <iostream>

#include <libdash.h>

int
main(int ac, char *av[])
{
    using TeamSpecT = dash::TeamSpec<2>;

    dash::init(&ac, &av);

    auto& team_all = dash::Team::All();

    dash::Team& team_1= team_all.split(2);

    team_all.barrier();
    dash::Team& team_2= team_all.split(2);

    dash::finalize();

    return 0;
}
```

Is this intentional?

devreal:
Make the team hierarchy a tree and allow for cloning teams, which is useful in multi-threaded environments.

Fixes #658

### DIFF
--- a/dash/include/dash/Team.h
+++ b/dash/include/dash/Team.h
@@ -390,6 +390,32 @@ public:
   }
 
   /**
+   * Create a clone of this team.
+   *
+   * \return A new team instance containing the same set of units as the
+   *         original.
+   */
+  inline Team & clone()
+  {
+    dart_team_t new_dartid;
+    if (is_null()) {
+      return *this;
+    }
+    dart_team_clone(_dartid, &new_dartid);
+    size_t num_siblings;
+    // tell the other teams that they now have a new sibling
+    {
+      std::lock_guard<std::mutex> g(_parent->_mutex);
+      num_siblings = _parent->_children.size();
+      for (Team *team : _parent->_children) {
+        team->_num_siblings++;
+      }
+    }
+    Team  *new_team = new Team(new_dartid, _parent, num_siblings, num_siblings);
+    return *new_team;
+  }
+
+  /**
    * Equality comparison operator.
    *
    * \param    rhs   The Team instance to compare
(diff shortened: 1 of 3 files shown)


bertwesarg:
As you disable all of the traversial API, than why do we have this `_children` member at all? I think most users can live without this member. AFAIK, MPI does not have a _child_ concept at all for its communicators and windows.

devreal:
Good question. It doesn't seem to be used anywhere in the examples or applications and I don't see an immediate use-case to iterate over levels in the `Team` tree. My guess is that since the localities are hierarchical the teams were meant to be as well but I'm not sure that that is useful...

bertwesarg:
There is one reason though, the deconstructor, which also removes all children. I'm fine with this as is.

dash/include/dash/Team.h
```diff
@@ -611,7 +672,8 @@ class Team
   mutable team_unit_t     _myid         = UNDEFINED_TEAM_UNIT_ID;
   mutable size_t          _size         = 0;
   Team                  * _parent       = nullptr;
-  Team                  * _child        = nullptr;
+  std::vector<Team*>      _children;
+  std::mutex              _mutex;
```
bertwesarg, line 676: `mutable`?

bertwesarg (changes requested):

dash/include/dash/Team.h
devreal: It should be mutable once we have `const` Teams :D

bertwesarg: I only referenced the `_mutex` (see the _M&M rule_ [here](https://en.cppreference.com/w/cpp/language/cv)). Also, I think that this **and** the `_children` would be a prerequisite for having `const` Teams in the future.

PR #664, opened 2019-07-22
Leave context empty in single-argument DASH_LOG_* macros calls

devreal:
Otherwise messages without a context won't be printed...

### DIFF
--- a/dash/include/dash/Mutex.h
+++ b/dash/include/dash/Mutex.h
@@ -39,6 +39,7 @@ private:
 
         if (ret != DART_OK) {
           DASH_LOG_ERROR(
+              "dash::Mutex::operator()",
               "Failed to destroy DART lock! "
               "(dart_team_lock_destroy failed)");
         }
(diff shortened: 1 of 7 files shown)


devreal:
Wow, the whole DASH code base is littered with `DASH_LOG_DEBUG("some string")` instances (who dafuq puts them in without needing them?!) so forcing a context explicitly would require fixing a gazillion places (as said before, there is no "quick fixing" here...) I will change the PR so that single messages are printed without context. I cannot currently push to Github, will try again later...

devreal:
Wow, the whole DASH code base is littered with `DASH_LOG_DEBUG("some string")` instances (who dafuq puts them in without needing them?!) so forcing a context explicitly would require fixing a gazillion places (as said before, there is no "quick fixing" here...) I will change the PR so that single messages are printed without context. I cannot currently push to Github, will try again later...

devreal:
Wow, the whole DASH code base is littered with `DASH_LOG_DEBUG("some string")` instances, none of which will ever have produced any output. Forcing a context explicitly would require me to fix a gazillion places (as said before, there is no "quick fixing" here...) I have changed the PR so that single messages are printed without context.

devreal:
Ups, looks like Github had a hickup and posted my earlier message that it said cannot be posted...

dhinf (approved):
works for me.

PR #671, opened 2019-10-30
Bug 659 tests containersize

Rodario:
Fixes the static containersize in the tests: 

- ArrayTest.MoveSemantics
- MatrixTest.UnderfilledBlockedPatternExtents
- MatrixTest.UnderfilledLocalViewSpec
- MatrixTest.LocalMatrixRefs
- MatrixTest.MoveSemantics
- DARTMemAllocTest.LocalAlloc

by relating it to dash::size(). 
The tests ran into a segfault when executed on a larger number of nodes.

Additionally a small bug in the syntax of Matrix-inl.h is fixed.

### DIFF
--- a/dash/test/dart/DARTMemAllocTest.cc
+++ b/dash/test/dart/DARTMemAllocTest.cc
@@ -109,7 +109,7 @@ TEST_F(DARTMemAllocTest, LocalAlloc)
 
 TEST_F(DARTMemAllocTest, SegmentReuseTest)
 {
-  const size_t block_size = 10*dash::size();
+  const size_t block_size = 10;
   dart_gptr_t gptr;
   ASSERT_EQ_U(
     DART_OK,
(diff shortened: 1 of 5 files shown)


dash/test/dart/DARTMemAllocTest.cc
```diff
@@ -109,7 +109,7 @@ TEST_F(DARTMemAllocTest, LocalAlloc)
 
 TEST_F(DARTMemAllocTest, SegmentReuseTest)
 {
-  const size_t block_size = 10;
+  const size_t block_size = 10*dash::size();
```
devreal: In this test the memory is allocated through `dart_team_memalloc_aligned` so it's not necessary to scale the size with the number of units.

devreal (changes requested):
Thanks, good catch! One minor thing inline.

dash/test/dart/DARTMemAllocTest.cc
Rodario: Fixed it.

devreal (approved):

PR #675, opened 2019-12-11
Support DASH builds and tests on MacOS 

anindex:
This PR proposes modifications to support compiling and testing DASH on MacOS.

### Environment Information
- Tested system: High Sierra 10.13 (Darwin 17.7.0) (other MacOS 10.x may also support)
- Compiler: Apple clang LLVM version 10.0.0 

### Dependencies:
All libraries below are built using AppleClang compiler.
- **hwloc** v1.11
- **PAPI** v5.7 (-lrt flag does not work on Mac due to lacking of rt lib => ignore and compile using `make -j8 -k` and then `make install`)
- **openblas** (install using `brew install openblas` and then tell Cmake later)
- **openmpi** v3.1.5 (download [here](https://www.open-mpi.org/software/ompi/v3.1/) and follow install instruction)

### Tested Installation:
- Config with **hwloc** and without **PAPI**: 
```
cmake -DENABLE_PAPI=OFF -DCMAKE_CXX_FLAGS=-I/usr/local/opt/openblas/include .. -L
```
- Config without **hwloc** and with **PAPI**: 
```
cmake -DENABLE_HWLOC=OFF -DENABLE_HWLOC_PCI=OFF -DCMAKE_CXX_FLAGS=-I/usr/local/opt/openblas/include .. -L
```
then compile
```
make -j8
make install
```

### Testing
```
mpirun build/dash/dash-test-mpi
```


dart-impl/base/src/hwinfo.c
```diff
@@ -358,9 +358,9 @@ dart_ret_t dart_hwinfo(
     hw.cpu_id = sched_getcpu();
   }
 #else
-  DART_LOG_ERROR("dart_hwinfo: "
-                 "HWLOC or PAPI required if not running on a Linux platform");
-  return DART_ERR_OTHER;
+//  DART_LOG_ERROR("dart_hwinfo: "
```
devreal: Can we use something like this here? 

https://gerrithub.io/c/geopm/geopm/+/367743/6/src/geopm_sched.c#102

dart-impl/mpi/src/dart_synchronization.c
```diff
@@ -23,7 +23,11 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
-#include <malloc.h>
+#ifdef __APPLE__
+  #include <malloc/malloc.h>
+#else
+  #include <malloc.h>
```
devreal: Do we really need `malloc.h` or can it be removed alltoghether? We already include `stdlib.h` (which provides `malloc`) and I don't see any other functions being used...

dart-if/include/dash/dart/if/dart_types.h
```diff
@@ -328,7 +328,13 @@ typedef enum
 dart_locality_scope_t;
 
 /** Maximum size of a host name string in \ref dart_hwinfo_t */
-#define DART_LOCALITY_HOST_MAX_SIZE        (HOST_NAME_MAX)
+#if defined(HOST_NAME_MAX)
+  #define DART_LOCALITY_HOST_MAX_SIZE      (HOST_NAME_MAX)
```
devreal, line 332: In this manpage it looks like we need to include `unistd.h`: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/gethostname.3.html

Maybe that fixes it?

dash/examples/bench.07.local-copy/main.cpp
```diff
@@ -11,7 +11,11 @@
 #include <cstring>
 #include <limits>
 
-#include <malloc.h>
+#ifdef __APPLE__
+  #include <malloc/malloc.h>
+#else
+  #include <malloc.h>
+#endif
```
devreal: Same as above

dash/examples/bench.07.local-copy/main.cpp
```diff
@@ -22,6 +26,16 @@ using std::endl;
 using std::setw;
 using std::setprecision;
 
+#ifndef __APPLE__
+  #define _aligned_malloc(size, alignment) memalign(alignment, size)
```
devreal: why not use `posix_memalign` instead of `memalign` alltogether? (`memalign` is deprecated anyway)

devreal (changes requested):
Thanks for the work, just a few comments inline

dart-impl/mpi/src/dart_synchronization.c
anindex: I just respect the original include :) if there is no use of `malloc.h`, it is safe to remove

dart-impl/base/src/hwinfo.c
anindex: Let me check it

devreal (approved):
Thanks, LGTM :)

bertwesarg:
This merge brought in some ugly coding style issues:

1. [`osx_sched_getcpu`](https://github.com/dash-project/dash/blob/f84e89ecfd262eefb6b7decf341ebc3fdcef1add/dart-impl/base/src/hwinfo.c#L14) has excessive white spaces at the end of the lines
2. There is an unindented [`if`](https://github.com/dash-project/dash/blob/f84e89ecfd262eefb6b7decf341ebc3fdcef1add/dart-impl/base/src/hwinfo.c#L403)

PR #677, opened 2020-01-02
Add missing include for cstddef before checking __cpp_lib_byte

devreal:
Siehe https://en.cppreference.com/w/cpp/utility/feature_test


devreal:
Ping

dhinf (approved):

PR #684, opened 2020-01-31
Move 'make sure dynamic windows are enabled` define into .c file

bertwesarg:
The `DART_MPI_ENABLE_DYNAMIC_WINDOWS` is only used inside the .c file,
so why adding the fallback into the .h file. Worth, the header is
installed and used by the header-lib part of DASH
`dash/include/dash/iterator/internal/GlobPtrBase.h`. Thus it gives the
impression, that you could change the value of
`DART_MPI_ENABLE_DYNAMIC_WINDOWS` after the installation of DASH.


devreal (approved):

bertwesarg:
Time to merge?

devreal:
Thanks for the reminder!

Issue #682, opened 2020-01-28
hwloc and LIKWID Versions


pauleonix:
It would be very cool to have minimum/maximum Version numbers for these in the readme. For both libraries there can be build problems if the wrong version is loaded. E.g. it doesn't work with hwloc@2.0.2, but with hwloc@1.11.11. If I then add likwid@4.3.2 there are build errors again. I have no other Versions of likwid installed and I would like to know which one will work (newer or older than 4.3.2) before I try to build likwid. 
Also it would be nice to know how DASH benefits exactly from these (and other) libraries, to know if they are worth the effort for a given project.

pauleonix:
To be more precise: with likwid@4.3.2 and hwloc@1.11.11 I get 
```
[  2%] Building C object dart-impl/base/CMakeFiles/dart-base.dir/src/internal/domain_locality.c.o
/home/spielix/dash/dart-impl/base/src/hwinfo.c: In function ‘dart_hwinfo’:
/home/spielix/dash/dart-impl/base/src/hwinfo.c:148:27: error: ‘dart_hwinfo_t’ {aka ‘struct <anonymous>’} has no member named ‘num_sockets’; did you mean ‘num_cores’?
       hw.num_numa    = hw.num_sockets;
                           ^~~~~~~~~~~
                           num_cores
/home/spielix/dash/dart-impl/base/src/hwinfo.c:151:53: error: ‘dart_hwinfo_t’ {aka ‘struct <anonymous>’} has no member named ‘num_sockets’; did you mean ‘num_cores’?
       hw.num_cores   = topo->numCoresPerSocket * hw.num_sockets;
                                                     ^~~~~~~~~~~
                                                     num_cores
make[2]: *** [dart-impl/base/CMakeFiles/dart-base.dir/src/hwinfo.c.o] Error 1
```

pauleonix:
Without likwid but with hwloc@2.0.2 I get 
```
[  3%] Building C object dart-impl/base/CMakeFiles/dart-base.dir/src/internal/domain_locality.c.o
/home/spielix/dash/dart-impl/base/src/hwinfo.c: In function ‘dart_hwinfo’:
/home/spielix/dash/dart-impl/base/src/hwinfo.c:313:33: error: ‘struct hwloc_obj’ has no member named ‘memory’
     hw.system_memory_bytes = obj->memory.total_memory / BYTES_PER_MB;
                                 ^~
[...]

PR #685, opened 2020-01-31
Fix compilation issues #682 hwloc 2.x and likwid

anindex:
This PR proposes portability fixes referent to issue #682. Tested sucessful combinations are:
- hwloc 1.11 + likwid 4.3.2
- hwloc 2.0

I think other versions should also work.

### DIFF
--- a/dart-impl/base/src/hwinfo.c
+++ b/dart-impl/base/src/hwinfo.c
@@ -316,13 +316,21 @@ dart_ret_t dart_hwinfo(
   if(hw.system_memory_bytes < 0) {
     hwloc_obj_t obj;
     obj = hwloc_get_obj_by_type(topology, HWLOC_OBJ_MACHINE, 0);
+#if HWLOC_API_VERSION < 0x00020000
     hw.system_memory_bytes = obj->memory.total_memory / BYTES_PER_MB;
+#else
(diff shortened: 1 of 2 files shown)


dart-impl/base/src/hwinfo.c
```diff
@@ -354,6 +345,36 @@ dart_ret_t dart_hwinfo(
...
+                   "likwid: topology_init failed, returned %d", likwid_ret);
+  } else {
+    CpuInfo_t     info = get_cpuInfo();
+    CpuTopology_t topo = get_cpuTopology();
+    if (hw.min_cpu_mhz < 0 || hw.max_cpu_mhz < 0) {
+      hw.min_cpu_mhz = info->clock;
+      hw.max_cpu_mhz = info->clock;
+    }
+    if (hw.num_numa < 0) {
+      hw.num_numa    = hw.num_sockets;
```
devreal: Is there a way to get `hw.num_sockets` from likwid? Otherwise both will be -1 if hwloc is not available

anindex: I have checked and found how to get numa and socket numbers in likwid :) see my commit below

devreal (approved):

Issue #692, opened 2020-02-21
StoreHDF::write for 1D  compile error


Goon83:
Hi,
I am testing the HDF5 support in DASH. It reports compile error for 1D data at compile time. 
Could you help to look at it ? See below for the code and error output.

Bests,
Bin

Code:
```
#include <libdash.h>
#include <iostream>

using dash::io::hdf5::hdf5_options;
using dash::io::hdf5::StoreHDF;

int main(int argc, char *argv[])
{
	dash::init(&argc, &argv);

	dash::Matrix<int, 1> *h5matrix = new dash::Matrix<int, 1>(dash::SizeSpec<1>(8));
	StoreHDF::write(*h5matrix, "testf.h5", "testg/testd");

	//2D works
	//dash::Matrix<int, 2> *h5matrix = new dash::Matrix<int, 2>(dash::SizeSpec<2>(8, 8));
	//StoreHDF::write(*h5matrix, "testf.h5", "testg/testd");

	dash::finalize();

	return EXIT_SUCCESS;
}
```

Error info:
```
In file included from h5-test.cpp:8:
In file included from /Users/dbin/opt/dash-0.4.0/include/libdash.h:71:
In file included from /Users/dbin/opt/dash-0.4.0/include/dash/io/HDF5.h:4:
/Users/dbin/opt/dash-0.4.0/include/dash/io/hdf5/StorageDriver.h:477:39: error:
      no member named 'local_blockspec' in 'dash::TilePattern<1,
      dash::ROW_MAJOR, long>'
    const auto & lblockspec = pattern.local_blockspec();
                              ~~~~~~~ ^
/Users/dbin/opt/dash-0.4.0/include/dash/io/hdf5/StorageDriver.h:582:12: note:
      in instantiation of function template specialization
      'dash::io::hdf5::StoreHDF::_get_hdf_slab_body<dash::TilePattern<1,
      dash::ROW_MAJOR, long> >' requested here
        1, _get_hdf_slab_body(pattern, {}));
           ^
/Users/dbin/opt/dash-0.4.0/include/dash/io/hdf5/internal/DriverImplZeroCopy.h:32:21: note:
      in instantiation of function template specialization
      'dash::io::hdf5::StoreHDF::_get_hdf_slabs<dash::TilePattern<1,
      dash::ROW_MAJOR, long> >' requested here
  auto hyperslabs = _get_hdf_slabs(container.pattern());
                    ^
[...]

PR #697, opened 2020-03-02
Patch for issue #692

anindex:
This PR proposes bug fix for issue #692 by adding explicitly SizeType and IndexType into template arguments when initializing patttern_t in StoreHDF read and write.


devreal (approved):

devreal:
@anindex Do I understand your commits right that HDF5 is not enabled in CI? It would be nice to have a test case for reported issue somewhere. I believe there is a dedicated set of HDF5 tests.

anindex:
I will write a test case for HDF5 read and write in other PR, for now, I will keep this fix minimal :)

devreal:
Looking at the patch again I realized that it fixes the *symptoms* of the problem (a type-mismatch in the HDF5 driver) but eventually we should address the underlying issues of the DASH types being inflexible with regard to the types they accept. `TilePattern` should, for example, have a c'tor templated on the type of `SizeSpec` and convert it if necessary. Maybe that will come as a byproduct of the pattern test suite :)

devreal:
Also: the codecov CI seems broken. Merging anyway.

PR #703, opened 2020-05-05
fixed dash::copy bug - wrong elements were copied when a view based i…

dhinf:
dash::copy selected wrong elements for copy when a GlobViewIterator was used. The ContiguousRangeIterator only checked against the view extent of the pattern and not the view used by the GlobViewIterator.

### DIFF
--- a/dash/include/dash/iterator/internal/ContiguousRange.h
+++ b/dash/include/dash/iterator/internal/ContiguousRange.h
@@ -78,8 +78,27 @@ public:
 
 private:
 
+
+  template<typename IterT>
+  size_type check_iter(const IterT& t, size_type num_elems, dim_t fast_dim) const {
+
+    return num_elems;
+  }
+
+  template<typename ElementT, typename PatternT, typename GlobMemT, typename PointerT, typename ReferenceT>
+  size_type check_iter(const GlobViewIter<ElementT, PatternT, GlobMemT, PointerT, ReferenceT>& it,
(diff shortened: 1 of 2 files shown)


dash/include/dash/iterator/internal/ContiguousRange.h
```diff
@@ -88,9 +107,9 @@ struct ContiguousRangeIterator {
 
     auto cur_last  = cur_first;
     size_type num_copy_elem = 0;
-    constexpr const int ndim = pattern_type::ndim();
+    constexpr const dim_t ndim = pattern_type::ndim();
     const auto& pattern = m_pos.pattern();
-    const int fast_dim = (pattern.memory_order() == dash::ROW_MAJOR) ? ndim - 1 : 0;
+    const dim_t fast_dim = (pattern.memory_order() == dash::ROW_MAJOR) ? ndim - 1 : 0;
```
devreal, line 112: This can be made `constexpr`, right? Not sure why it wasn't before...

devreal (approved):
Ahh, views are beyond my mental capacities... Thanks for the fix!

dash/include/dash/iterator/internal/ContiguousRange.h
dhinf: the clang compiler complains with constexpr, but it is not really necessary at this point.
gcc is fine with constexpr

PR #707, opened 2020-05-25
dart.h: Remove `extern "C"`

bertwesarg:
All headers include have this already and it is error prone to include
system headers inside such a scope

Reported-by: S.Schueller@itc.rwth-aachen.de


devreal (approved):
Thanks for the patch!

PR #710, opened 2020-06-10
fixes compile and linking errors for gcc 10 and clang 10

dhinf:
- added overloaded assignment operator for nullptr in GlobPtr
- declared two variables extern in dart base to avoid linking errors

### DIFF
--- a/dash/include/dash/GlobPtr.h
+++ b/dash/include/dash/GlobPtr.h
@@ -121,6 +121,15 @@ public:
    */
   self_t &operator=(const self_t &rhs) = default;
 
+  /**
+   * Assignment operator.
+   */
+  self_t &operator=(std::nullptr_t) {
+    m_dart_pointer = DART_GPTR_NULL;
+
+    return *this;
+  }
+
   //clang-format off
   /**
    * Implicit Converting Constructor, only allowed if one of the following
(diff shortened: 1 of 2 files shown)


bertwesarg:
LGTM

dash/include/dash/GlobPtr.h
```diff
@@ -121,6 +121,15 @@ class GlobPtr {
    */
   self_t &operator=(const self_t &rhs) = default;
 
+  /**
+   * Assignment operator.
```
bertwesarg:
```suggestion
   * nullptr assignment operator.
```

devreal (approved):
LGTM, thanks for the fix (I thought along the same way but had to leave for a long weekend before I was able to implement it).

Also, good catch for the DART variables. One of these things where you wonder how that ever actually compiled...

bertwesarg:
> Also, good catch for the DART variables. One of these things where you wonder how that ever actually compiled...

They were `common` symbols. A quite old concept for variables. And they get axed recently in new compilers/linkers:

* [GCC](https://gcc.gnu.org/gcc-10/porting_to.html)
* [LLVM/Clang](https://reviews.llvm.org/D75056), not yet released?
* [ARM Clang](http://www.keil.com/support/man/docs/armclang_ref/armclang_ref_chr1422532346348.htm)

PR #712, opened 2020-06-25
DART-MPI: call into MPI every once in a while for local put/get

devreal:
We need to call into MPI every once in a while as otherwise polling on a local variable may not trigger progress if that is needed.

### DIFF
--- a/dart-impl/mpi/src/dart_communication.c
+++ b/dart-impl/mpi/src/dart_communication.c
@@ -220,16 +220,6 @@ dart__mpi__get_basic(
 {
   if (num_reqs) *num_reqs = 0;
 
-  if (team_data->unitid == team_unit_id.id) {
-    // use direct memcpy if we are on the same unit
-    memcpy(dest, seginfo->selfbaseptr + offset,
-        nelem * dart__mpi__datatype_sizeof(dtype));
-    DART_LOG_DEBUG("dart_get: memcpy nelem:%zu "
-        "source (coll.): offset:%lu -> dest: %p",
-        nelem, offset, dest);
-    return DART_OK;
-  }
-
 #if !defined(DART_MPI_DISABLE_SHARED_WINDOWS)
   DART_LOG_DEBUG("dart_get: shared windows enabled");
   if (seginfo->segid >= 0 && team_data->sharedmem_tab[team_unit_id.id].id >= 0) {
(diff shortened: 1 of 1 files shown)


dhinf:
why 16 ? educated guess?

dart-impl/mpi/src/dart_communication.c
```diff
@@ -30,6 +30,11 @@
 #include <math.h>
 #include <alloca.h>
 
+/* the number of consecutive memcpy for local put/get before calling into MPI */
+#define NUM_CONSECUTIVE_MEMCPY 16
```
dhinf: why 16 ? educated guess?

devreal: Yes, I'm open to other suggestions :D

bertwesarg: at least document, that 16 is just an educated guess

dart-impl/mpi/src/dart_communication.c
```diff
@@ -30,6 +30,11 @@
...
+
+/* number of performed local memcpy between calling into MPI */
+static _Thread_local int num_local_memcpy = 0;
```
bertwesarg: this needs C11, is this already ensured by CMake? I also think its better to use `<threads.h>` and `thread_local` here.

bertwesarg: and no `= 0` needed

dart-impl/mpi/src/dart_communication.c
```diff
@@ -220,13 +225,15 @@ dart__mpi__get_basic(
...
+      num_local_memcpy++ < NUM_CONSECUTIVE_MEMCPY) {
     // use direct memcpy if we are on the same unit
     memcpy(dest, seginfo->selfbaseptr + offset,
         nelem * dart__mpi__datatype_sizeof(dtype));
     DART_LOG_DEBUG("dart_get: memcpy nelem:%zu "
         "source (coll.): offset:%lu -> dest: %p",
         nelem, offset, dest);
+    num_local_memcpy = 0;
     return DART_OK;
   }
```
bertwesarg: `get_shared_mem` and `put_shared_mem` also call `memcpy`, are they also effected by this?

devreal: They do not, as that doesn't depend on progress triggered locally.

dart-impl/mpi/src/dart_communication.c
devreal: You're right, we don't officially require a C11 compiler. I am not going to open the discussion on whether  we should abandon support for >20 year old compilers though...

devreal:
After having given this some more thought I modified the PR to *always* call into MPI for local memory accesses. If fast local access without progress is required the application should just dereference the native pointer provided by DASH. Once we are in DART we have lost the latency race anyway. Everything else just adds complexity.

bertwesarg:
+1

PR #714, opened 2020-07-02
build: Quote properties in `set_target_properties`

bertwesarg:
Avoids errors like:

```
CMake Error at dart-impl/base/CMakeLists.txt:113 (set_target_properties):
  set_target_properties called with incorrect number of arguments.
```


devreal (approved):
LGTM! It looks like CI is tripping, will try again in a few...

